PR #18, opened 2016-07-11
Lb/oomph setup

LorenzoBettini:
This adds/fixes a few features in the Oomph setup:

- Install SWTBot in the IDE (for launch configuration) and in the Targlet for xtext-xtend
- Install in the IDE a few Mylyn connectors
- Queries both bugzilla and github for bugs and issues, respectively, and show them in the Mylyn view

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -243,7 +243,7 @@
   </setupTask>
   <setupTask
       xsi:type="setup.p2:P2Task"
-      label="eclipse">
+      label="Eclipse">
     <requirement
         name="org.eclipse.emf.sdk.feature.group"/>
     <requirement
@@ -265,7 +265,7 @@
   </setupTask>
   <setupTask
       xsi:type="setup.p2:P2Task"
-      label="tools">
+      label="Tools">
     <requirement
         name="org.eclipse.m2e.feature.feature.group"
         versionRange="[1.0.0,1.7.0)"/>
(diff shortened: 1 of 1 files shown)


kthoms:
Similar to the repositories I would move the Mylyn Queries into the subprojects. If you install just "Core", you don't want to have the IDEA issues, for example.

LorenzoBettini:
@kthoms that's what I thought myself at the beginning... I grouped them altogether because you may want to check whether other subprojects, which you haven't imported, have some issues which might be related to what you're working on...

anyway, I can put queries into subprojects if you want.

What do you think?

kthoms:
Just for harmony: The p2 task label "Mylyn" is started with capital, while "eclipse" and "tools" are lower case. I think the labels should be harmonized. I think "eclipse" and "tools" should be capitalized.

kthoms:
@LorenzoBettini I understand this idea. But usually you would have installed all projects. Candidates to exclude might be IDEA, Web, Maven. And if you do not select them, you are really not interested in those issues.

LorenzoBettini:
@kthoms OK, I'll implement your two proposals in a minute

LorenzoBettini:
@kthoms please have a look at the updated PR

kthoms:
Please assign a label to variable "github.user.password": "GitHub user password".

The setup works well. I tested with both all projects and only subset of projects.

LorenzoBettini:
@kthoms Done, but it is "Github user password", not "GitHub user password", to keep consistency with other related github variables

kthoms:
OK, same typo everywhere :-) From my side the setup is OK now.

LorenzoBettini:
@svenefftinge @dhuebner do you think I can merge?

PR #25, opened 2016-07-22
Added generated sources of Eclipse example projects

kthoms:
Executed workflows with on all example projects
Fixed test classes to use the new annotations from xtext.tesing plugin


kthoms:
I have materialized a fresh workspace with Oomph and fixed all the compile issues that occurred in the "Eclipse Examples" working set.

LorenzoBettini:
Since the git repository has been "cleaned up" during the splitting, I was wondering whether it could be the right time NOT to put generated sources at least for examples.  The Oomph setup can be configured to run the mwe2 workflow files (I do that in all my projects), and the Xtend builder in Eclipse would do Xtend compilation automatically (some additional Oomph build steps might be required) and the build could generate the example sources as well (before preparing the zips of the examples).

meysholdt:
@LorenzoBettini let's discuss that here: eclipse/xtext#1066

meysholdt:
Hi Karsten, thank you for the contribution. Can you please split up the commit into smaller ones to make a review possible. I understand that the current single commit mixes at least three logical changes: Adding new files to *.ide/src, refactoring the tests and re-adding xtend-gen and src-gen.

see https://github.com/eclipse/xtext/blob/master/QUALITY_GUIDELINES.md#commits-and-pull-requests

kthoms:
OK, will do when finding time again.

org.eclipse.xtext.xtext.ui.examples/projects/org.eclipse.xtext.example.homeautomation.tests/src/org/eclipse/xtext/example/homeautomation/tests/RuleEngineParsingTest.xtend
```diff
@@ -0,0 +1,29 @@
...
+@InjectWith(RuleEngineInjectorProvider)
+class RuleEngineParsingTest{
+
+	@Inject
+	ParseHelper<Model> parseHelper
+
+	@Test 
+	def void loadModel() {
+		val result = parseHelper.parse('''
+			Hello Xtext!
```
meysholdt: this test would fail, wouldn't it? "Hello Xtext!" does not seem to valid syntax of the rule engine lang.

kthoms: Right. The generator added this class. I will make it more meaningful.

meysholdt:
thank you Karsten, all looks good!

kthoms:
Thanks for the review :)

PR #30, opened 2016-07-28
[commons][types] Retrieve the modifiers for a dirty state type.

### LINKED ISSUE #28: Invalid dirty state modifiers in JdtTypesProposalProvider.searchAndCreateProposals.
Let a grammar that is defining the concept of agent.
Let the definitions of a "final agent" A1 and a "not-final agent" A2:

```
final agent A1 { }
agent A2 { }
```

Let the definition of the agent A3 extending another agent definition. Agent super-types must be obtained with the content assist mechanism (`|` is the current position of the current when the proposal provider is called).

```
agent A3 extends |
```

A `IContentProposalProvider` implementation is provided with a specific filter that is apply at the location of the cursor:

``` java
private class ExtensionFilter extends TypeMatchFilters.AbstractFilter {

    private final ITypesProposalProvider.Filter visibilityFilter;

    private final String modelFullName;

    ExtensionFilter(ContentAssistContext context, int searchFor) {
        super(searchFor);
        this.modelFullName = SARLProposalProvider.this.qualifiedNameProvider.getFullyQualifiedName(
                context.getCurrentModel()).toString();
        this.visibilityFilter = createVisibilityFilter(context, searchFor);
    }

    @Override
    public boolean accept(int modifiers, char[] packageName, char[] simpleTypeName,
                char[][] enclosingTypeNames, String path) {
        // Avoid auto reference of type.
        final String fullName = JavaModelUtil.concatenateName(packageName, simpleTypeName);
        if (Objects.equals(this.modelFullName, fullName)) {
            return false;
        }
        //The following tests are done by the visibility filter.
        //if (TypeMatchFilters.isInternalClass(simpleTypeName, enclosingTypeNames)) {
        //  return false;
        //}
        //if (!TypeMatchFilters.isAcceptableByPreference().accept(modifiers, packageName,
        //      simpleTypeName, enclosingTypeNames, path)) {
        //  return false;
        //}
        // Final modifier test
        if (Flags.isFinal(modifiers)) {
            return false;
        }
        return this.visibilityFilter.accept(modifiers, packageName, simpleTypeName, enclosingTypeNames, path);
    }
}
```

When the definition of A1 is another resource than the one of A3, the proposal provider provides A2 only.

When the definition of A1 is in the same resource pf the one of A3, the proposal provider provides A2 and A1; that is not the expected behavior.

In the case of A1 and A3 defined in the same resource, the `modifiers` argument in the `accept` function for A1 is equal to 1 (Flags.AccPublic). **This value does not reflect the final modifier.**

After deeper debugging, I discover that the JdtTypesProposalProvider.searchAndCreateProposal
considers A1 to be part of the dirty state of the current edited resource. Then the filtering function is invoked with the hard-coded modifiers value Flags.AccPublic. So that is cannot reflect the real modifiers applied to A1.

gallandarakhneorg:
close #28

Signed-off-by: Stéphane Galland galland@arakhne.org

### DIFF
--- a/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
+++ b/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
@@ -38,12 +38,18 @@ import org.eclipse.jdt.internal.ui.viewsupport.JavaElementImageProvider;
 import org.eclipse.jface.text.contentassist.ICompletionProposal;
 import org.eclipse.jface.viewers.StyledString;
 import org.eclipse.swt.graphics.Image;
+import org.eclipse.xtext.common.types.JvmAnnotationType;
+import org.eclipse.xtext.common.types.JvmDeclaredType;
+import org.eclipse.xtext.common.types.JvmEnumerationType;
+import org.eclipse.xtext.common.types.JvmGenericType;
+import org.eclipse.xtext.common.types.JvmMember;
 import org.eclipse.xtext.common.types.JvmType;
 import org.eclipse.xtext.common.types.TypesPackage;
 import org.eclipse.xtext.common.types.access.IJvmTypeProvider;
 import org.eclipse.xtext.common.types.access.jdt.IJavaProjectProvider;
 import org.eclipse.xtext.common.types.access.jdt.IJdtTypeProvider;
 import org.eclipse.xtext.common.types.access.jdt.JdtTypeProviderFactory;
+import org.eclipse.xtext.common.types.util.DeprecationUtil;
 import org.eclipse.xtext.common.types.util.RawSuperTypes;
 import org.eclipse.xtext.conversion.IValueConverter;
 import org.eclipse.xtext.conversion.ValueConverterException;
@@ -78,6 +84,7 @@ import com.google.inject.Provider;
  * @author Sebastian Zarnekow - Initial contribution and API
  * @author Jan Koehnlein - introduced QualifiedName
  * @author Christoph Kulla - added support for hovers
+ * @author Stephane Galland - added function for retreiving dirty state modifiers
  */
 public class JdtTypesProposalProvider extends AbstractTypesProposalProvider {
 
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
```diff
@@ -400,14 +403,66 @@ public boolean isCanceled() {
...
+				if (filter.accept(modifiers, qualifiedName.skipLast(1).toString().toCharArray(), qualifiedName.getLastSegment().toCharArray(), new char[0][0], description.getEObjectURI().toPlatformString(true))) {
 					String fqName = description.getQualifiedName().toString();
-					createTypeProposal(fqName, Flags.AccPublic, fqName.indexOf('$') > 0, proposalFactory, myContext, scopeAware, jvmTypeProvider, valueConverter);
+					createTypeProposal(fqName, modifiers, fqName.indexOf('$') > 0, proposalFactory, myContext, scopeAware, jvmTypeProvider, valueConverter);
 				}
 			}
 		}
 	}
 	
+	/** Compute the JVM modifiers that corresponds to the given description.
```
LorenzoBettini: It should be

/**
 * Compute...

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
```diff
@@ -400,14 +403,66 @@ public boolean isCanceled() {
...
+	 *
+	 * <p>This function fixes the issue related to the missed modifiers given to the content assist.
+	 *
+	 * @param context the current content assist context.
+	 * @param description the description.
```
LorenzoBettini: there should be a

@since 2.11

LorenzoBettini:
It would be good to have a test

gallandarakhneorg:
Ok. I will publish updates and tests soon.

gallandarakhneorg:
I have difficulties for defining the tests in the content assist because they should be based on object-oriented statements and the testing language seems to be not. I have put @Ignore on the tests.

LorenzoBettini:
@svenefftinge should we create a test language for the purpose or should we do without the test?

cdietrich:
or move it to the xtend content assist tests

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
```diff
@@ -400,15 +403,69 @@ public boolean isCanceled() {
...
+	protected int getDirtyStateModifiers(ContentAssistContext context, IEObjectDescription description) {
+		EObject eobject = description.getEObjectOrProxy();
+		if (eobject.eIsProxy()) {
+			eobject = EcoreUtil.resolve(eobject, context.getResource().getResourceSet());
+		}
+		int accessModifiers = Flags.AccPublic;
+		int otherModifiers = 0;
+		if (eobject instanceof JvmMember) {
+			final JvmMember member = (JvmMember) eobject;
+			if (member.isDeprecated()) {
```
svenefftinge: We should use org.eclipse.xtext.common.types.util.DeprecationUtil.isDeprecatedMember(JvmMember) here. It also checks for @Deprecated annotation

svenefftinge:
We could have such tests in Xtend, but then we also need to do the filtering in the proposal provider accordingly. That would be a good feature (excluding final classes) though

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
```diff
@@ -400,15 +403,69 @@ public boolean isCanceled() {
...
+				final JvmDeclaredType type = (JvmDeclaredType) eobject;
+				if (type.isFinal()) {
+					otherModifiers |= Flags.AccFinal;
+				}
+				if (type.isAbstract()) {
+					otherModifiers |= Flags.AccAbstract;
+				}
+				if (type.isStatic()) {
+					otherModifiers |= Flags.AccStatic;
+				}
```
cdietrich: This should retrieve AccInterface, AccEnum, AccAnnotation flags as well see https://bugs.eclipse.org/bugs/show_bug.cgi?id=473453

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
gallandarakhneorg: Ok, I will update the code accordingly.

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
gallandarakhneorg: I will update this PR to keep it independent of Xtend. Then I will provide tests and code that are specific to the Xtend project.

gallandarakhneorg:
Done for the pure Xtext part.

spoenemann:
@LorenzoBettini could you review eclipse/xtext-xtend#23 and merge both PRs if all is fine?

LorenzoBettini:
@spoenemann OK, I'll try

LorenzoBettini:
I found a few problems in the related PR, see https://github.com/eclipse/xtext-xtend/pull/23

cdietrich:
@LorenzoBettini see my new related pullreq https://github.com/eclipse/xtext-xtend/pull/97

LorenzoBettini:
I also verified the Jenkins build, so I'm merging. Then we merge https://github.com/eclipse/xtext-xtend/pull/97 whose tests run fine, once the commit of this PR is used

Issue #4, opened 2016-06-24
Fix Oomph Setup


svenefftinge:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=496683

LorenzoBettini:
for the moment which is the best way to have a working workspace?

spoenemann:
Use the "Gradle Project" import wizard for xtext-lib, xtext-core, xtext-extras, xtext-idea, and xtext-web. The Maven projects probably need to be imported manually.

LorenzoBettini:
And concerning the Eclipse distribution and plug-ins?

spoenemann:
Not sure, but as said before I think currently they have to be imported manually into the workspace, which includes setting the target platform manually.

LorenzoBettini:
I actually meant the plugins you have to install into Eclipse :)

spoenemann:
I am using a standard Eclipse package where I installed Xtext 2.10.

spoenemann:
> Should STS Gradle be replaced by Buildship?

Yes

> Should this be handled in a separate issue?

I'm not sure what there is to be done for that. Currently I'm evaluating checking in the project meta data for projects imported with Buildship.

> Is upgrade to Java-8 OK? (https://bugs.eclipse.org/bugs/show_bug.cgi?id=457927)

Yes, we already switched to Java 8 for all projects.

> Should the setup be splitted into several subprojects, and Git Clone Tasks relative to the Subproject?

Do you mean several Oomph projects? That could be useful for people who want to work on a subset of Xtext.

> Should the Target Platform be set by the Core project?

No, the target platform is needed only by the Eclipse plugins, so it's defined in xtext-eclipse.

> Is it assumed that the Core project is always selected?

Not necessarily. The Gradle and Maven IDE integrations should use snapshot versions for projects that are not in the workspace.

kthoms:
> I'm not sure what there is to be done for that. Currently I'm evaluating checking in the project meta data for projects imported with Buildship.
I'm doing the same right now. The Project Import fails when the buildship settings file is missing, and Oomph Setup refuses to continue then. Not a nice behavior.

I have some progress on the Xtext.setup already. It would be best if you check in the Buildship metadata and delete the .settings/gradle directories.

kthoms:
PR created
https://github.com/eclipse/xtext-eclipse/pull/15

kthoms:
[...]

PR #39, opened 2016-08-03
#4 Removed org.springsource.ide.eclipse.gradle.core preferences since we use Buildship now

kthoms:


cdietrich:
thanks you karsten

PR #48, opened 2016-08-12
[#1070] New subproject "Xtext Issues"

kthoms:
allow separate use of Mylyn and issue database


kthoms:
See https://github.com/eclipse/xtext/issues/1070#issuecomment-239469146 for screenshots of the changed setup. @meysholdt Is this as you would expect?

meysholdt:
yes, it is, thank you!

PR #59, opened 2016-08-19
Add test plugins to SDK feature

kthoms:


meysholdt:
thx. I see you also moved `xbase.junit` to the `xtext.sdk` feature. 

Is it valid to argue that since `xbase.junit` depends on `xtext.junit` and `xtext.junit` always has been in `xtext.sdk` this move in not going to break anybody who has a feature-based something?

kthoms:
If someone would have tests depending on the xbase feature, it would be mandatory to include `xtext.junit` from the SDK feature. Thus, if someone would use feature based descriptions to include `xbase.junit`, they would have to include the SDK feature as well.

meysholdt:
yup, exactly.

PR #63, opened 2016-08-24
Auto-formatting when pasting code.

gallandarakhneorg:
The auto-formatter is listening changes on the document, and is applying
formatting modifications on the changes.

By default, the auto-formatter is used only when doing the paste
operation.

Signed-off-by: Stéphane Galland galland@arakhne.org


spoenemann:
Thanks for the proposal. However I'm not convinced that we should apply this behavior on such a general level. I would prefer to leave it to each application how auto-formatting is handled and to not offer any default behavior.

gallandarakhneorg:
No problem for me. It's work fine in my compiler.

spoenemann:
More opinions on this? Otherwise I would close the PR.

meysholdt:
I think doing auto-formatting on paste by default is confusing and potentially harmful.
But in general this feature reminds me of "paste and match style" just as some word processors provide it.

paste

svenefftinge:
I agree. For eclipse it would be more idiomatic to integrate this as a clean up action on save.

Issue #83, opened 2016-09-15
[oomph] Bogus version 2.10.1 for Xtext & Xtend installed


kthoms:
When setting up an Xtext development environment with Oomph the resulting Eclipse has the version 2.10.1.v201607210527 for Xtext and Xtend installed.

![screenshot 16](https://cloud.githubusercontent.com/assets/265597/18548386/b063e136-7b48-11e6-8c6a-989ca80e4e62.png)

It is expected to install the latest successful CI build.

kthoms:
The current Oomph setup now installs Xtext 2.11.

![screenshot 127](https://cloud.githubusercontent.com/assets/265597/20761378/199ddfb0-b723-11e6-9dba-6831b7947b6c.png)

PR #84, opened 2016-09-15
[#83] Add CI p2 repository for Xtend

kthoms:
Xtext and Xtend must be both updated from the latest CI build. Since the
Xtend p2 repository was missing, Xtext could also not be updated.

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -50,6 +50,10 @@
       value="http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/master/lastStableBuild/artifact/build/p2-repository/"/>
   <setupTask
       xsi:type="setup:VariableTask"
+      name="xtend.nightly.composite"
+      value="http://services.typefox.io/open-source/jenkins/job/xtext-xtend/job/master/lastSuccessfulBuild/artifact/build/p2-repository/"/>
+  <setupTask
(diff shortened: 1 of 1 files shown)


releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -50,6 +50,10 @@
       value="http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/master/lastStableBuild/artifact/build/p2-repository/"/>
   <setupTask
       xsi:type="setup:VariableTask"
+      name="xtend.nightly.composite"
+      value="http://services.typefox.io/open-source/jenkins///job/xtext-xtend/job/master/lastSuccessfulBuild/artifact/build/p2-repository/"/>
```
LorenzoBettini: I would avoid the triple ///
This is probably just cosmetic, but with p2 you never know ;)

kthoms: Yes, it is cosmetic. I copied the URL from Jenkins and used the URL. But you are right, this is better.

kthoms: Change is pushed.

LorenzoBettini:
@kthoms I'm trying to update my workspace setup after this change has been pushed to master; during the p2 directory it uses the new latest repo for Xtend, but oomph keeps saying that there are no updates available and the Xtend SDK is still 2.10.1 in my Eclipse...

does this work for you?

PR #96, opened 2016-09-30
Regenerated Xtend code

kthoms:
Sources produced by current Xtend build 2.11.0.v20160929-1448

Signed-off-by: Karsten Thoms karsten.thoms@itemis.de


LorenzoBettini:
I noted these differences in my workspace as well, but, if I understand correctly, the different generated Java code is due to a new version of jdt (probably you're using Neon as the target platform), in particular, internal parts of jdt which have changed with some new additional parameters.

This makes the generated Java code not compilable with previous versions of the Eclipse target platform (I guess not even with Mars).

If Java code generated by Xtend is not re-generated during the build before releasing then this will cause runtime errors if one install the plugins in, say, a Luna Eclipse.

I might be wrong of course, but that's how I had interpreted the differences in the generated Java code.

kthoms:
This is a very good point! I thought that it was due to a change in the Xtend compiler. So this might be invalid.

PR #105, opened 2016-10-19
made dependency to pde optional. resolves #91

### LINKED ISSUE #91: dependency from 'org.eclipse.xtext.junit4' to PDE should be optional
the dependency was introduced in 2f890f08fd76637ac9b93ccead634f8b065f97fc but it breaks backwards compatibility with projects that don't have PDE in their target platform

cdietrich:
made dependency to pde optional. resolves #91


cdietrich:
please review

spoenemann:
👍

PR #107, opened 2016-10-19
Me kill xtext sdk

meysholdt:


meysholdt:
it's just about restructuring the build

PR #113, opened 2016-11-07
Set workspace encoding to UTF-8

kthoms:
Most projects have explicit encodings set, but we have to make sure that the encoding defaults to UTF-8.


cdietrich:
thx

PR #115, opened 2016-11-07
Fixed working sets for projects from xtext-eclipse

kthoms:
Changed reference project, old reference project org.eclipse.xtext.sdk.feature was moved to other repo
Removed obsolete OrPredicate for natures


kthoms:
Before the change the projects were all in "Other Projects". This is how it now looks like.

screenshot 84

cdietrich:
thx

PR #117, opened 2016-11-09
Use Xtext,Xtend&MWE2 from "latest" compisite at Eclipse

kthoms:


cdietrich:
thx

PR #125, opened 2016-11-22
improve outline refresh performance

Bananeweizen:
When changing the sorting, the complete tree needs to be refreshed.
Especially for big trees disabling all repainting during the update can
cut down the refresh time to half of before.

Signed-off-by: Michael Keppler <michael.keppler@gmx.de>


Bananeweizen:
Thanks. And congratulations to your new committer state within the project. I was one of those assuming you already were...

PR #126, opened 2016-11-23
Kth/issue44

kthoms:
There are some issues with the Domainmodel example


cdietrich:
👍

org.eclipse.xtext.xbase.junit/xtend-gen/org/eclipse/xtext/xbase/compiler/InMemoryJavaCompiler.java
```diff
@@ -238,7 +238,7 @@ public InMemoryJavaCompiler(final ClassLoader parent, final CompilerOptions comp
     InMemoryJavaCompiler.ClassLoaderBasedNameEnvironment _classLoaderBasedNameEnvironment = new InMemoryJavaCompiler.ClassLoaderBasedNameEnvironment(parent);
     this.nameEnv = _classLoaderBasedNameEnvironment;
     this.parentClassLoader = parent;
-    Map _map = compilerOptions.getMap();
+    Map<String, String> _map = compilerOptions.getMap();
```
LorenzoBettini: I think that also this change is due to some internal changed in JDT which are only in the latest versions, so this code would not be runnable in an older Eclipse version (see also the other comment)

org.eclipse.xtext.xbase.ui/xtend-gen/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -50,7 +50,7 @@
   public ITypeRoot getTypeRoot(final IEditorInput it) {
     ITypeRoot _xblockexpression = null;
     {
-      final Object adapter = it.getAdapter(IJavaElement.class);
+      final IJavaElement adapter = it.<IJavaElement>getAdapter(IJavaElement.class);
```
LorenzoBettini: be careful: this will make this Java source file compilable (and runnable) only when using the Neon target platform: since adapters are type paremeterized only starting from Neon.

I think we discussed this problem in another PR or issue

kthoms: Right, I'll revert these classes and update the PR.

kthoms:
Replaced by https://github.com/eclipse/xtext-eclipse/pull/128

Issue #44, opened 2016-08-05
[example] DomainModel-Example is based on an old xtext version


djetzen:
The DomainModel-Example is still based on an old Xtext version. This means, that for example the validator is written in Java and not Xtend, there is still the previous project layout and so on.

meysholdt:
contributions are very welcome :)

The work to be done is to to migrate [GenerateDomainmodel.mwe2](https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.xtext.ui.examples/projects/org.eclipse.xtext.example.domainmodel/src/org/eclipse/xtext/example/domainmodel/GenerateDomainmodel.mwe2) to the new generator infrastructure.

PR #128, opened 2016-11-24
[#44] Update Domainmodel example

kthoms:
Workflow modifications:
- remove version identifier from genertor comment
- avoid generation of generator stub
- Version identfiers removed from generator comment
- Javadoc
- Generated code for Xtend classes was missing
- refactored manifests
- added Xtend nature to .ide project
- ignore some files
- updated plugin.xml files


org.eclipse.xtext.xtext.ui.examples/projects/org.eclipse.xtext.example.domainmodel/META-INF/MANIFEST.MF
```diff
@@ -1,34 +1,32 @@
...
- org.eclipse.xtext.example.domainmodel.services,
- org.eclipse.xtext.example.domainmodel.validation,
- org.eclipse.xtext.example.domainmodel.serializer,
- org.eclipse.xtext.example.domainmodel.scoping
-Bundle-RequiredExecutionEnvironment: JavaSE-1.8
-Import-Package: org.apache.log4j;version="1.2.15"
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Xtext Domainmodel Example
+Bundle-Vendor: Eclipse Xtext
```
dhuebner: As Bundle-Name and Vendor are directly in MANIFEST we can delete the plugin.properties file. Otherwise it is confusing.  
We should also setup the same line-wrap in xtext ws it is hard to review.

org.eclipse.xtext.xtext.ui.examples/projects/org.eclipse.xtext.example.domainmodel.ui/plugin.xml
```diff
@@ -1,215 +1,211 @@
...
-            </reference>
-         </activeWhen>
-      	</handler>
-      	<!-- copy qualified name -->
-        <handler
-            class="org.eclipse.xtext.example.domainmodel.ui.DomainmodelExecutableExtensionFactory:org.eclipse.xtext.ui.editor.copyqualifiedname.EditorCopyQualifiedNameHandler"
-            commandId="org.eclipse.xtext.ui.editor.copyqualifiedname.EditorCopyQualifiedName">
-            <activeWhen>
+	<extension
+		point="org.eclipse.ui.editors">
```
dhuebner: And we need the same TAB vs. 4 Spaces settings :(

dhuebner: I didn't get the change here. Is that just the order of extension tags that changed?

org.eclipse.xtext.xtext.ui.examples/projects/org.eclipse.xtext.example.domainmodel/build.properties
```diff
@@ -2,10 +2,19 @@ source.. = src/,\
...
+                     org.eclipse.xtext.common.types,\
+                     org.eclipse.xtext.xtext.generator,\
+                     org.eclipse.emf.codegen.ecore,\
+                     org.eclipse.emf.mwe.utils,\
+                     org.eclipse.emf.mwe2.launch,\
+                     org.eclipse.emf.mwe2.lib,\
+                     org.objectweb.asm,\
+                     org.apache.commons.logging,\
+                     org.apache.log4j,\
+                     com.ibm.icu
```
dhuebner: ICU as build time dependency? Could you please explain why it is necessary.

kthoms:
@dhuebner Thanks for your remarks. I will consider them carefully and update the PR.

org.eclipse.xtext.xtext.ui.examples/projects/org.eclipse.xtext.example.domainmodel.ui/plugin.xml
kthoms: The UI's plugin.xml is just the generated one, extended by the section "manual additions". The generator will produce tabs only now, so you might see many differences to the previous version. But this is related to the new generator infrastructure. So the generated part can be taken as-is.

This screenshot shows the difference between plugin.xml and plugin.xml_gen. 

screenshot 137

org.eclipse.xtext.xtext.ui.examples/projects/org.eclipse.xtext.example.domainmodel/build.properties
kthoms: Cannot recall how this came in. Reduced the build dependencies so that the workflow is executed without errors.

kthoms:
Don't know where the problems come from. I have rebased my changes.
Will open another PR.

Issue #31, opened 2016-07-29
Fix broken Builder Tests


cdietrich:
Enable and fix the broken builder.tests

TwoProjectsTest is very flaky (at least locally)

cdietrich:
fixed in 2.11

PR #137, opened 2016-12-09
made very flaky tests more stable. solves #31

cdietrich:
made very flaky tests more stable. solves #31


cdietrich:
please review

LorenzoBettini:
Just for curiosity, why is

```
Thread.sleep(10);
```

required? it's just 10 milliseconds delay... I think the important part is to wait for build, or am I missing something?

cdietrich:
wait for build is just calling a incremental build.
i assume there is some eventing missing if the build is started too early

LorenzoBettini:
OK... it's strange that 10 milliseconds are enough to delay the build ;)

PR #141, opened 2016-12-14
Removed duplicate test classes

kthoms:
Classes exist already in org.eclipse.xtext.purexbase.tests
Launch config was configured for wrong project


kthoms:
This PR only deletes those test classes that are runtime tests and have to be removed from o.e.x.purexbase.eclipse.tests. The only remaining test is `BuilderWithTracesTest` which does not use the old XtextRunner.

The corresponding classes in o.e.x.purexbase.tests are already using the new runner, e.g. [CompilerTest.xtend](https://github.com/eclipse/xtext-extras/blob/master/org.eclipse.xtext.purexbase.tests/src/org/eclipse/xtext/purexbase/test/CompilerTest.xtend).

PR #142, opened 2016-12-19
Configure build tasks

kthoms:
This change will add build task configuration per sub project. When using Eclipse for Committers as base setup (recommended) the Mylyn Build feature is already available.

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -75,6 +75,14 @@
     <description>Github user password</description>
   </setupTask>
   <setupTask
+      xsi:type="setup:VariableTask"
+      id="jenkins.url"
+      type="URI"
+      name="jenkins.url"
+      value="http://services.typefox.io/open-source/jenkins">
+    <description>Base URL of Jenkins build server</description>
+  </setupTask>
+  <setupTask
       xsi:type="setup:CompoundTask"
       name="Workspace Preferences">
     <setupTask
@@ -406,7 +414,7 @@
               project="org.eclipse.xtext"/>
           <operand
               xsi:type="workingsets:ExclusionPredicate"
-              excludedWorkingSet="//@setupTasks.14/@workingSets.0"/>
+              excludedWorkingSet="//@setupTasks.15/@workingSets.0"/>
         </predicate>
       </workingSet>
     </setupTask>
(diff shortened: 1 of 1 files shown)


LorenzoBettini:
But if one does not use Eclipse for Committers (and Mylyn Build feature is not installed) what happens?

The feature could be installed from the Oomph setup, couldn't it?

kthoms:
The [contribution guide](https://github.com/eclipse/xtext/blob/master/CONTRIBUTING.md) says 

> On the Product page, select Eclipse IDE for Eclipse Committers.

However, I've checked now with distro _Eclipse IDE for Java Developers_, and there the Mylyn build feature is also avalaible. So I think it should work.

LorenzoBettini:
But why don't we simply install that from the Oomph setup? I mean, if the setup assumes that some features are installed in the IDE then the Oomph setup should install them... @dhuebner what do you think?

kthoms:
The Mylyn build feature is configured to be installed with the "Xtext Issues" subproject. The intention to have it there is to minimize the installed feature set for developers only interested in code.
Originally I thought to put also the build task definitions there, but decided to put it per subproject now.
I could move the Mylyn build feature to the set of always installed features outside of the Xtext issues section.
But it seems that it is not necessary to install it by Oomph explicitly, since it is already provided by the base distro. So I could also remove installation of Mylyn features in the setup also.
In any case, the configured build tasks will populate the Builds view.

LorenzoBettini:
OK.
I mean if nothing bad happens in case that feature is not installed in Eclipse

kthoms:
OK, now I have used Eclipse Platform as base. And again, the Mylyn build feature is available then. Seems that Oomph installs that automatically when a build task is configured.
This again raises the question: Do we want to have build information and Mylyn features installed when any subproject is selected? Or shall I move the build tasks to the _Issues_ subproject?

LorenzoBettini:
For me it would be OK, but probably it's better to discuss that with the whole team :)

dhuebner:
@kthoms Please move it to the "Issues" sub project.
I'm using "Eclipse Platform" because it is the smallest one and Mylyn plugin is probably the most laud eclipse plugin ever.

dhuebner:
You can also consider to rename the "Issues" sub-project to "Issues and Builds"

kthoms:
So you'd prefer to move the build configuration to the separate subproject and not each build config to its subproject?

dhuebner:
I would prefer to have additional "Mylyn builds" tasks under "Xtext Issues" like we have it already for "Mylyn Queries Task"s

kthoms:
Yes, I think this is better now. It is an optional feature that can be installed:

screenshot 3

The build integration feature will only be installed when this subproject is selected. When chosen the _Builds_ view is available and populated:

screenshot 4

cdietrich:
@kthoms what is the state of this pr?

kthoms:
Pending. Have to reactivate it and resolve conflicts.

PR #144, opened 2016-12-21
Target platform: Upgrade org.junit to 4.12

kthoms:
see eclipse/xtext-core#208


sailingKieler:
@kthoms I upgraded to references of `org.junit` in `org.eclipse.xtext.testing` & `org.eclipse.xtext.xbase.testing` to 4.12.0 that doesn't seem to cause any problems in the IDE (I installed mine via oomph). Anyway, I'm fine with upgrading the TP as we rely on 4.12. in the gradle-based build configs.

Issue #119, opened 2016-11-10
[wizard] Improve usability of Generic IDE Support option


kthoms:
![screenshot 96](https://cloud.githubusercontent.com/assets/265597/20172875/19cf2a08-a737-11e6-8b86-8a1cb8ae1d4e.png)

The option "Generic IDE Support" is mandatory for the options "Eclipse Plugin", "IntelliJ IDEA Plugin" and "Web Support". It is not intuitive that this option is the last one. It should be the first option that a user could choose and dependend on the option, the other options should be made selectable.

An option could be to make the different IDE integration options part of a group, as it is with the "Eclipse Plugin" and "Create Feature/Update Site" options. This would also make the error message obsolete, since it is clear that without the Generic IDE option the other options are not valid.

svenefftinge:
There was a longer discussion about this somewhere in bugzilla if I recall correctly.

kthoms:
Could you link that please? If redundant, close this one.

svenefftinge:
I would have done that if I had found it.

kthoms:
I will try to find it tomorrow.

kthoms:
I searched the wizard-related (also closed) bugs, but did not find it either. Thus I would like to discuss this issue here if that's OK.

As said, I suggest that we group specific IDE support options within a generic IDE group. What do you think?

dhuebner:
> The option "Generic IDE Support" is mandatory for the options "Eclipse Plugin", "IntelliJ IDEA Plugin" and "Web Support". It is not intuitive that this option is the last one. It should be the first option that a user could choose and dependend on the option, the other options should be made selectable.

"Generic IDE Support" is more a technical aspect for advance users. Keeping in mind, that wizards are mostly used by people who tries things out or are giving demos (except Xtext consultants), it's better to show the really important thing in top of the wizard page. This things are sorted by importance for a common eclipse user

- I need an eclipse plugin (preselected for eclipse IDE)
- I need an IDEA plugin too (probably not)
- I need web support (probably not)

The two other option like Generic IDE support and tests are considered as 90% needs this and are technical aspects. 

If we compare the default wizard settings with the screen shot above :
![bildschirmfoto 2016-11-15 um 13 41 28](https://cloud.githubusercontent.com/assets/454322/20306217/732fa086-ab39-11e6-8e70-ddf586ba1f4d.png)

[...]

PR #148, opened 2017-01-02
[#119] UX improvements for Generic IDE Support option

kthoms:
The "Generic IDE Support"  option is disabled whenever a UI plugin is selected. If
it was not enabled before, it will be selected automatically and an
information about this will be reported to the status widget.

Initial page: "Generic IDE Support" is disabled, since Eclipse Plugin is preselected.
screenshot 20

The option becomes enabled when Eclipse Plugin is deseleted.
screenshot 21

The option is deselected manually:
screenshot 22

The option "Web Support" is selected. The user is informed about the automatic selection of the "Generic IDE Support" option.
screenshot 23

### DIFF
--- a/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.xtend
+++ b/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.xtend
@@ -40,6 +40,8 @@ class AdvancedNewProjectPage extends WizardPage {
 	Group createUiProjectSubGroup
 
 	StatusWidget statusWidget
+	
+	boolean autoSelectIdeProject
 
 	new(String pageName) {
 		super(pageName)
@@ -75,7 +77,7 @@ class AdvancedNewProjectPage extends WizardPage {
 				]
 				createIdeProject = CheckBox [
 					text = AdvancedNewProjectPage_projIde
-					enabled = true
+					enabled = false
 				]
 				createTestProject = CheckBox [
 					text = Messages.WizardNewXtextProjectCreationPage_TestingSupport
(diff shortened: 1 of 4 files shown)


kthoms:
The _Generic IDE Support_ option is explained in an additional info decoration:

screenshot 24

org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.xtend
```diff
@@ -211,22 +233,11 @@ class AdvancedNewProjectPage extends WizardPage {
...
-			} else {
-				reportIssue(ERROR, '''
-				Projects like '�affectedProjects�' depends on '�createIdeProject.text�' project.
-				Please <a>enable '�createIdeProject.text�'</a> project.''', [
-					createIdeProject.selection = true
-				])
-			}
+		if (autoSelectIdeProject) {
+			autoSelectIdeProject = false
+			reportIssue(INFORMATION, '''
```
holgerschill: This overrides the message about the buildsystem, but the statusWidget can only show one message at once. For me that does not sound like a big issue but maybe other see that different.

org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.xtend
```diff
@@ -211,22 +233,11 @@ class AdvancedNewProjectPage extends WizardPage {
...
+				'�createIdeProject.text�' project was automatically selected as option '�(source as Button).text�' requires it.
```
holgerschill: This overrides the message about the buildsystem, but the statusWidget can only show one message at once. For me that does not sound like a big issue but maybe other see that different.

holgerschill: Beside that the code works like expected and looks good to me.

kthoms: Right, the status widget only stores one message at once. So if the user selects a project/build system combination that raises a message and then selects a UI project while Generic IDE was not selected so far, then the notice will overwrite the build system message. But the user has seen that message before.

PR #151, opened 2017-01-09
[510084] Consider project specific default encoding

kthoms:
Added test case

### DIFF
--- a/org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/encoding/WorkspaceEncodingProviderTest.java
+++ b/org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/encoding/WorkspaceEncodingProviderTest.java
@@ -1,0 +1,89 @@
+/*******************************************************************************
+ * Copyright (c) 2017 itemis AG (http://www.itemis.com) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ui.tests.editor.encoding;
+
+import static org.eclipse.xtext.junit4.ui.util.IResourcesSetupUtil.*;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+
+import org.eclipse.core.resources.IFolder;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.xtext.builder.nature.XtextNature;
+import org.eclipse.xtext.junit4.ui.AbstractWorkbenchTest;
+import org.eclipse.xtext.parser.IEncodingProvider;
+import org.eclipse.xtext.resource.XtextResource;
+import org.eclipse.xtext.resource.XtextResourceSet;
+import org.eclipse.xtext.testlanguages.ui.internal.TestlanguagesActivator;
+import org.eclipse.xtext.ui.editor.WorkspaceEncodingProvider;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.google.inject.Injector;
+
+/**
+ * @author Karsten Thoms - Initial contribution and API
+ */
+public class WorkspaceEncodingProviderTest extends AbstractWorkbenchTest {
+	@Inject
+	private IEncodingProvider encodingProvider;
+	@Inject
+	private IWorkspace workspace;
+	@Inject Provider<XtextResourceSet> resourceSetProvider;
+	
+	private Injector injector;
+	private String defaultEncoding;
+	private IProject project;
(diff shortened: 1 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=510084

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/WorkspaceEncodingProvider.java
```diff
@@ -47,30 +48,27 @@
...
+			while (storages.hasNext()) {
+				Pair<IStorage, IProject> storage = storages.next();
+				if (storage.getFirst() instanceof IEncodedStorage) {
 					try {
-						String result = storage.getSecond().getDefaultCharset(true);
-						return result;
+						return ((IEncodedStorage) storage.getFirst()).getCharset();
 					} catch (CoreException e) {
-						LOG.error("Error getting project's default encoding", e);
+						LOG.warn("Error getting file encoding", e);
```
svenefftinge: why is this now a warning?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/WorkspaceEncodingProvider.java
```diff
@@ -47,30 +48,27 @@
...
 					}
 				}
+				try {
+					return storage.getSecond().getDefaultCharset(true);
+				} catch (CoreException e) {
+					LOG.warn("Error getting project's default encoding", e);
```
svenefftinge: same here? Why warning?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/WorkspaceEncodingProvider.java
```diff
@@ -47,30 +48,27 @@
...
-				while (storages.hasNext()) {
-					Pair<IStorage, IProject> storage = storages.next();
-					if (storage.getFirst() instanceof IEncodedStorage) {
-						try {
-							return ((IEncodedStorage) storage.getFirst()).getCharset();
-						} catch (CoreException e) {
-							LOG.error("Error getting file encoding", e);
-						}
-					}
+		if (workspace != null && uri != null) {
```
svenefftinge: The only semantic difference seems to be, that on uri==null it will no longer return the default charset of the workspace. Why is that desirable?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperImpl.java
```diff
@@ -215,10 +215,14 @@ protected boolean isHandled(IFolder folder) {
 	}
 
 	private Iterable<Pair<IStorage, IProject>> getStorages(/* @NonNull */ URI uri, IFile file) {
+		Iterable<Pair<IStorage, IProject>> storages = null;
 		if (file == null || !file.isAccessible()) {
-			return contribution.getStorages(uri);
+			storages = contribution.getStorages(uri);
+		}
+		if (file!=null && (storages == null || Iterables.isEmpty(storages))) {
+			storages = Collections.singleton(Tuples.<IStorage,IProject>create(file, file.getProject()));
```
svenefftinge: Changing the contract of this method will affect a lot of other locations. Why are you doing this here?

org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/encoding/WorkspaceEncodingProviderTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.eclipse.xtext.ui.editor.WorkspaceEncodingProvider;
+import org.junit.Before;
+import org.junit.Test;
+
+import com.google.inject.Injector;
+
+/**
+ * @author Karsten Thoms - Initial contribution and API
+ */
+public class WorkspaceEncodingProviderTest extends AbstractWorkbenchTest {
```
svenefftinge: I haven't tried but I'd guess that these tests have passed before the changes below, already. No?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/WorkspaceEncodingProvider.java
kthoms: When the encoding could not be retrieved from the resource (which is unlikely to happen anyway) the encoding will fall back to runtime default. So the issue is not critical. I would consider this not as an error and lowered the message level therefore.

org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/encoding/WorkspaceEncodingProviderTest.java
kthoms: No they did not. Only testWorkspaceDefaultEncoding would succeed.

kthoms: Before:
screenshot 44

After:
screenshot 45

kthoms:
If you have such doubt here, please take the test and fix it like you think it is better. The test reproduces the issue, and the change would fix it. No other test is failing.

svenefftinge:
please merge after 2.11

PR #152, opened 2017-01-10
Align the version of jdt.core in maven to the one from gradle

holgerschill:
Signed-off-by: Holger Schill <Holger.Schill@itemis.de>

### DIFF
--- a/releng/org.eclipse.xtext.tycho.parent/pom.xml
+++ b/releng/org.eclipse.xtext.tycho.parent/pom.xml
@@ -25,8 +25,8 @@
 		<log4j.version>1.2.16</log4j.version>
 		<commons-logging.version>1.1.3</commons-logging.version>
 		<commons-cli.version>1.2</commons-cli.version>
-		<equinox-common-version>3.6.0</equinox-common-version>
-		<jdt-core-version>3.10.0</jdt-core-version>
+		<equinox-common-version>3.8.0</equinox-common-version>
+		<jdt-core-version>3.12.2</jdt-core-version>
 		<icu4j.version>52.1</icu4j.version>
 		<antlr.version>[3.2, 3.3)</antlr.version>
 		<antlr-generator-version>[2.1.1, 3.0)</antlr-generator-version>


releng/org.eclipse.xtext.tycho.parent/pom.xml
```diff
@@ -25,8 +25,8 @@
 		<log4j.version>1.2.16</log4j.version>
 		<commons-logging.version>1.1.3</commons-logging.version>
 		<commons-cli.version>1.2</commons-cli.version>
-		<equinox-common-version>3.6.0</equinox-common-version>
-		<jdt-core-version>3.10.0</jdt-core-version>
+		<equinox-common-version>3.8.0</equinox-common-version>
+		<jdt-core-version>3.12.2</jdt-core-version>
```
spoenemann: I don't think these versions are used anywhere. The Tycho build does not have any Maven dependencies. There's a separate parent pom for the Maven plugin build.

holgerschill: Sorry I do not understand what you mean.

spoenemann: I think these properties are leftovers from the old Maven build before the repo splitting. The xtext-eclipse build is purely based on Tycho and a target platform, so all dependencies are resolved via p2 repositories. That means these properties shouldn't be used anywhere in the Maven build (please verify via text search).

holgerschill: That would imply that when we decide to not change the MANIFEST.MF files we cannot do something in maven, right?

holgerschill: Indeed the properties are not used.

spoenemann: There are still Maven builds without Tycho in xtext-maven and xtext-xtend. There the properties are still relevant. But for Tycho builds the dependencies are determined via each bundle's MANIFEST.MF and the common target platform.

holgerschill: I know but this does not answer the question if we can do something if we do not want to change the MANIFEST.MF.

spoenemann: What do you mean by "do something"?

holgerschill: I mean curve the version in stone without changing the MANIFEST.MF files just for the build. Currently it depends on the target-platform.

spoenemann: You mean we shouldn't allow the dependencies to diverge between Maven and the target platform? Maybe we could solve that by re-running the xtext-extras tests in the eclipse build using the target platform, which would verify whether our bundles are binary compatible with Eclipse Luna. That could run as an extra pipeline stage in the Jenkins build.

PR #158, opened 2017-01-13
Don't call InputStream.reset() when it's not supported

spoenemann:
Fixes eclipse/xtext-core#239.

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/EclipseResourceFileSystemAccess2.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/EclipseResourceFileSystemAccess2.java
@@ -240,20 +240,26 @@ public class EclipseResourceFileSystemAccess2 extends AbstractFileSystemAccess2 
 	/**
 	 * @since 2.9
 	 */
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/EclipseResourceFileSystemAccess2.java
```diff
@@ -248,9 +248,11 @@ protected void generateFile(IFile file, InputStream content, IFile traceFile, Ch
 			if (file.exists()) {
 				if (outputConfig.isOverrideExistingResources()) {
 					if (hasContentsChanged(file, content)) {
```
meysholdt: unfortunately `hasContentsChanged()` also reads the stream, so after `hasContentsChanged()` the streams's curser is probably at the stream's end. I think for streams without marker support we'd need to read the stream into a buffer once and then use that buffer's contents for `hasContentsChanged()` and `file.setContents`.

spoenemann: I'll add a test to handle this properly.

spoenemann:
The provided test failed before the fix was applied.

PR #159, opened 2017-01-13
Added Xtend Maven plugin

spoenemann:
For a proper source mapping.


spoenemann:
See eclipse/xtext#1066

PR #160, opened 2017-01-16
[#119] Auto (de) selection of Eclipse projects

kthoms:
The creation of a p2 repository requires the SDK project. If not
selected before, the feature project is automatically selected and an
information message about the selection is presented.
When 'Eclipse plug-in' is deselected, also the feature and p2 projects
are deselected.

Select 'Create Update Site', 'Create Feature' is automatically selected:
screenshot 48

Eclipse plug-in deselected, feature and p2 projects were selected before:
screenshot 47


kthoms:
See also https://github.com/eclipse/xtext-eclipse/pull/148

PR #167, opened 2017-01-20
Improved XtextQuickAssistProcessor.getApplicableAnnotations(..)

spoenemann:
Fixes eclipse/xtext-core#124.

Behavior before this change:

 * If there is at least one annotation left of the cursor position, choose the nearest one with respect to the start offset. Quick fixes are shown for all annotations with exactly same range as the chosen one.
 * If there is at least one annotation right of the cursor position, choose the one with the smallest offset. Quick fixes are shown for all annotations with exactly same range as the chosen one.

Behavior after this change:

 * If there is at least one annotation that includes the cursor position, quick fixes are shown for all annotations that include the cursor position.
 * If there is at least one annotation left of the cursor position, choose the nearest one with respect to the end offset. Quick fixes are shown for all annotations with the same end offset as the chosen one.
 * If there is at least one annotation right of the cursor position, choose the one with the smallest offset. Quick fixes are shown for all annotations with the same offset as the chosen one.

The change also fixes another bug: The selection is supposed to be set to the range of the annotations for which quick fixes are displayed. This was prevented by a wrong condition in the respective method.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/XtextQuickAssistProcessor.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/XtextQuickAssistProcessor.java
@@ -12,6 +12,7 @@ import static java.util.Arrays.*;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import org.eclipse.core.runtime.OperationCanceledException;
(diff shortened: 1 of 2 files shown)


spoenemann:
Please don't merge yet! This should be included in version `2.11.1`.

kthoms:
Now it would be time to merge it

spoenemann:
No, I'd like to further improve the behavior as described in eclipse/xtext-core#124.

spoenemann:
Now it's fine to merge from my side.

PR #182, opened 2017-02-12
Fix typo in refactoring error message

Bananeweizen:
Second try with local editing.


cdietrich:
Thanks for the contribution

PR #184, opened 2017-02-14
GlobalURIEditorOpener: Marked IWorkbench as @Nullable

### LINKED ISSUE #153: IWorkbench Nullable in GlobalURIEditorOpener
`GlobalURIEditorOpener` contains following code

```
	@Inject(optional=true) 
	/* @Nullable */
	private IWorkbench workbench;
``` 

this will not work if `IWorkbench` is provided as null at runtime.

guice will allow null if any `@Nullable` annotation is present at the class. 

thus: would it be possible to annotate our own Nullable annotation there (and potential other places as well - at least those needed to start a eclipse without a workbench / headless)

see https://bugs.eclipse.org/bugs/show_bug.cgi?id=437425 as well

cdietrich:
GlobalURIEditorOpener: Marked IWorkbench as @Nullable so that Xtext Languages can activate in a headless eclipse without workbench

Fixes #153 
Fixes [437425] 


cdietrich:
please review

PR #187, opened 2017-02-17
Adding support of XConstructorCall into Xbase dispatching hover.

gallandarakhneorg:
The XbaseDispatchingEObjectTextHover class does not reply the
XConstructorCall, but the JvmConstructor. This avoids to use Xbase hover
API (the JDT one is used).

This commit adds the support of XConstructorCall to the
XbaseDispatchingEObjectTextHover class.


cdietrich:
looks good

PR #188, opened 2017-02-19
[481981] use file encoding to create stringinputstream to create files

cdietrich:
[481981] use file encoding to create stringinputstream to create files

### DIFF
--- a/org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/ui/util/IResourcesSetupUtil.java
+++ b/org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/ui/util/IResourcesSetupUtil.java
@@ -9,6 +9,7 @@ package org.eclipse.xtext.junit4.ui.util;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.InvocationTargetException;
 
 import org.eclipse.core.resources.ICommand;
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=481981

cdietrich:
please review

org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/ui/util/IResourcesSetupUtil.java
```diff
@@ -220,7 +221,11 @@ protected void execute(IProgressMonitor monitor)
 					InterruptedException {
 				create(file.getParent());
 				file.delete(true, monitor());
-				file.create(new StringInputStream(s), true, monitor());
+				try {
+					file.create(new StringInputStream(s, file.getCharset(true)), true, monitor());
+				} catch (UnsupportedEncodingException exc) {
+					throw new IllegalArgumentException(exc.getMessage(), exc);
```
kthoms: I think it is not correct to throw IAE here, since the exception is not caused by an argument. The UEE is thrown when the determined encoding for the file is not supported. A CoreException is the right alternative here.

cdietrich: A core exception is hard to create, what about a IllegalStateException?

kthoms: How about
```throw new CoreException(new Status(IStatus.ERROR, "org.eclipse.xtext.junit4", exc.getMessage(), exc));```

kthoms: (or even without passing the root cause as 4th arg)

cdietrich:
updated pull req

Issue #194, opened 2017-02-28
Cast to ITextSelection instead of TextSelection


rmitin:
In GoToMatchingBracketAction method run() tests if selection is TextSelection, but depending on editor extensions this text selection could be wrapped into something that only implements ITextSelection. Similar cast could be corrected in StratumBreakpointAdapterFactory.

cdietrich:
fixed in 2.12

PR #195, opened 2017-02-28
[#194] Cast to ITextSelection instead of TextSelection

rmitin:
Change GoToMatchingBracketAction and StratumBreakpointAdapterFactory to
do cast to ITextSelection instead of TextSelection to allow editors wrap
TextSelection into some extended selection while still implementing
ITextSelection.

Signed-off-by: Roman Mitin <roman.mitin@gmail.com>


cdietrich:
Tank you for the contribution

meysholdt:
Thank you, Roman!

Issue #197, opened 2017-03-01
Adapt Popup Colors to behaviour in JDT/Oxygen


cdietrich:
In JDT/Oxygen (at least on Linux) the Color of Popup Windows (e.g. Hover) has changed from yellowish to whitish.

![bug_popup](https://cloud.githubusercontent.com/assets/238736/23461937/2f87326c-fe8c-11e6-9ae4-e9af99d6e02a.png)

https://bugs.eclipse.org/bugs/show_bug.cgi?id=497984
https://bugs.eclipse.org/bugs/show_bug.cgi?id=500196

We should adapt our popups as well.

https://bugs.eclipse.org/bugs/show_bug.cgi?id=512888

the annotation hover / content assist seems to change automatically.

Besides the HTML Printer Stuff

- `org.eclipse.xtext.xbase.ui.hover.XbaseInformationControl.createContent(Composite)`
- RenameRefactoringPopup

PR #198, opened 2017-03-01
Don't make use of COLOR_INFO_* in Oxygen+ #197

cdietrich:


cdietrich:
please review

LorenzoBettini:
There's some duplicated code, and I was wondering if we could come up with some kind of wrapper class including also defaults (which could be specified at constructor time or via lambdas?)... just an idea... otherwise it looks good to me.

PR #201, opened 2017-03-03
Oomph: bugzilla query for Xtend bugs

LorenzoBettini:
Currently there was a mylyn bugzilla query only for Xtext, not for
Xtend.


LorenzoBettini:
@kthoms makes sense! I've updated the PR

Issue #209, opened 2017-03-21
Xtext editor doesn't notice changes of a file outside workspace


sprigogin:
To reproduce:
1. Open a file outside Eclipse workspace using an xtext-based editor.
2. Type something in the editor, but don't save the file
2. Change the file outside Eclipse, for example, by appending a blank line to it
3. Move the focus back to Eclipse

The editor doesn't notice that the file has changed.

Now repeat the same procedure with a file with .txt extension. The editor displays the dialog: "The file 'filename.txt' has been changed on the file system. Do you want to replace the editor contents with these changes?". This is the desired behavior.

This issue is a clone of http://bugs.eclipse.org/490414

sprigogin:
Please review and merge the proposed fix.

cdietrich:
@sprigogin i had to change your code to work with the luna target platform. can you please test if it still fixes your problem

cdietrich:
fixed in 2.12

PR #210, opened 2017-03-21
Fix for https://github.com/eclipse/xtext-eclipse/issues/209

sprigogin:
This change adds timestamp tracking for files outside workspace opened
in Xtext-based editors.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
@@ -22,9 +22,12 @@ import java.nio.charset.UnsupportedCharsetException;
 import java.util.Collections;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/sprigogin_master/4/

svenefftinge:
Thank you!

cdietrich:
this does not work with luna, => build is broken
i can try to do an incanceof adaptable but i dont know if this will fix your actual problem

cdietrich:
https://github.com/eclipse/xtext-eclipse/pull/218/

sprigogin:
Replacing Adapters.adapt with IAdaptable.adapt will not work with objects that can be adapted only by using adapter factories. It makes sense to change it back to Adapters.adapt as soon as Luna support is no longer required.

cdietrich:
that was clear, i was just asking if this solves YOUR usecase ;D

sprigogin:
IAdaptable.adapt is sufficient for my use case because FileStoreEditorInput implements IAdaptable.

PR #220, opened 2017-03-28
don't use deprecated Proposal.apply(Document) anymore. fixes #219

### LINKED ISSUE #219: ContentAssistTest for Template Proposals did not work
The following Content Assist test case on the org.eclipse.xtext.example.domainmodel project fails, but it should succeed.

templateproposaltest

Reproducible with Eclipse Neon.2 and Xtext 2.11.0.

cdietrich:


cdietrich:
please review

LorenzoBettini:
Looks good to me!
The test you added to the domainmodel example is also great, but, as far as I know, the tests of examples are not run by the CI. Do you think we could add a template proposal test to one of the existing testing languages?

cdietrich:
good point, i will have a look

cdietrich:
test is here https://github.com/eclipse/xtext-xtend/pull/189

PR #221, opened 2017-03-30
[builder] fixed lookup of IFiles by source-relative uris when installing trace markers

svenefftinge:


svenefftinge:
http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/se_install_tracemarkers/

Issue #225, opened 2017-04-04
BuilderParticipant.getDerivedResources has a NxM issue


cdietrich:
`org.eclipse.xtext.builder.BuilderParticipant.getDerivedResources(Delta, Map<String, OutputConfiguration>, Map<OutputConfiguration, Iterable<IMarker>>)` has a NxM issue

if you do a full build on a project that contains xtext-based models and there generated files and do not do a clean build before then `BuilderParticipant.getDerivedResources` will have a NxM runtime

(x deltas times x markers/generatedfiles - i assue there is one foreach source file)

cdietrich:
`org.eclipse.core.resources.IMarker.getAttribute(String)`
goes into `org.eclipse.core.internal.resources.MarkerAttributeMap.get(Object)`
which call `String.intern`.
this might become really expensive for 10000(Deltas) times 10000(Resources) = 100.000.000 calls of `org.eclipse.xtext.builder.DerivedResourceMarkers.getSource(IMarker)`

cdietrich:
in a normal clean/build cycle the derived resources are delted by org.eclipse.xtext.builder.BuilderParticipant.cleanOutput(IBuildContext, OutputConfiguration, EclipseResourceFileSystemAccess2, IProgressMonitor) so this problem does not occur.

cdietrich:
trace from yourkit

```
Stacks at 2017-04-04 04:31:48 PM. Uptime is 2m 47s 200ms.

Worker-7 State: RUNNABLE CPU usage on sample: 982ms
java.lang.String.intern() String.java (native)
org.eclipse.core.internal.resources.MarkerAttributeMap.get(Object) MarkerAttributeMap.java:135
org.eclipse.core.internal.resources.MarkerInfo.getAttribute(String) MarkerInfo.java:108
org.eclipse.core.internal.resources.Marker.getAttribute(String) Marker.java:99
org.eclipse.xtext.builder.DerivedResourceMarkers.getSource(IMarker) DerivedResourceMarkers.java:164
org.eclipse.xtext.builder.BuilderParticipant.getDerivedResources(IResourceDescription$Delta, Map, Map) BuilderParticipant.java:390
org.eclipse.xtext.builder.BuilderParticipant.doBuild(List, Map, Map, IXtextBuilderParticipant$IBuildContext, EclipseResourceFileSystemAccess2, IProgressMonitor) BuilderParticipant.java:260
[...]

PR #227, opened 2017-04-05
use reverse lookup map to find source files for generated files. #225

cdietrich:
use reverse lookup map to find source files for generated files. #225


cdietrich:
@svenefftinge could you please have a look (will add a second pull request for xtend builder participant)

Issue #229, opened 2017-04-07
Xtext nature icon can not be superseded


simonmoosbrugger:
Yet there is no possibility to supersede the Xtext nature icon because it is defined in the "org.eclipse.ui.decorators" extension point. If it would be implemented via the "org.eclipse.ui.ide.projectNatureImages" extension point we would gather the chance to influence the order of the decoraters by the order of the project natures in the ".project" file.

see
[https://bugs.eclipse.org/bugs/show_bug.cgi?id=456114](url)

cdietrich:
fixed in 2.12

PR #230, opened 2017-04-07
[456114][#229] Xtext nature overlay with the projectNatureImages extension point

cdietrich:
[456114][#229] Xtext nature overlay with the projectNatureImages extension point

### DIFF
--- a/org.eclipse.xtext.ui.shared/plugin.xml
+++ b/org.eclipse.xtext.ui.shared/plugin.xml
@@ -114,29 +114,12 @@
       </objectContribution>
    </extension>
    <extension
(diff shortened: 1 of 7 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=456114

cdietrich:
please review. do we really want to do this?

cdietrich:
downside is: add xtext nature action adds the nature at the end. so the xtext overlay wont be visible by default

svenefftinge:
+1 
Xtext really is just a framework used to build actual language plugins, so it should not expose itself into the UI too much (if at all).

Issue #237, opened 2017-04-19
[ParallelResourceLoader] Add Factory Method to ResourceLoaderProviders with an increased timeout


cdietrich:
We run eclipse tests with parallel resource loading enabled in a highly virtualized environment. this makes us run into the timeout of `ParallelResourceLoader` from time to time.

It would be nice to have a factory method in ResourceLoaderProviders that allows us to set the timeout.

@svenefftinge would it be ok to add such a method?

```
getParallelLoader(final int nrOfThreads, final int bufferSize, long timeoutTime, TimeUnit timeoutUnit) {
return new AbstractResourceLoaderProvider() {
			@Override
			public IResourceLoader get() {
				ParallelResourceLoader resourceLoader = new ParallelResourceLoader(getResourceSetProvider(), getResourceSorter(), nrOfThreads, bufferSize);
				resourceLoader.setTimeout(timeoutTime, timeoutUnit)
				return resourceLoader;
			}
		};
}
```

cdietrich:
Feature is delivered with Xtext 2.12

PR #239, opened 2017-04-25
[ParallelResourceLoader] Add Factory Method to ResourceLoaderProviders with an increased timeout

cdietrich:
[ParallelResourceLoader] Add Factory Method to ResourceLoaderProviders with an increased timeout
Solves #237

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package org.eclipse.xtext.builder.resourceloader;
 
+import java.util.concurrent.TimeUnit;
+
 import org.eclipse.xtext.builder.resourceloader.IResourceLoader.Sorter;
 import org.eclipse.xtext.ui.resource.IResourceSetProvider;
 
@@ -42,9 +44,22 @@ public final class ResourceLoaderProviders {
 
 	/** Returns a loader with a parallelization degree of 2 to 4 (depending on how many processors are available). */
 	public static Provider<IResourceLoader> getParallelLoader() {
+		int nThreads = getNumberOfThreads();
+		return getParallelLoader(nThreads, 0);
+	}
+
+	/**
+	 * since 2.12
+	 */
+	public static Provider<IResourceLoader> getParallelLoader(long timeoutTime, TimeUnit timeoutUnit) {
+		int nThreads = getNumberOfThreads();
+		return getParallelLoader(nThreads, 0, timeoutTime, timeoutUnit);
+	}
+
+	private static int getNumberOfThreads() {
 		int nProcessors = Runtime.getRuntime().availableProcessors();
 		int nThreads = Math.max(2, Math.min(4, nProcessors));
-		return getParallelLoader(nThreads, 0);
+		return nThreads;
 	}
 
 	public static Provider<IResourceLoader> getParallelLoader(final int nrOfThreads) {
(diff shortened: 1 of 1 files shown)


cdietrich:
please review

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
```diff
@@ -42,15 +44,39 @@ public Sorter getResourceSorter() {
...
+	public static Provider<IResourceLoader> getParallelLoader(long timeoutTime, TimeUnit timeoutUnit) {
+		int nThreads = getNumberOfThreads();
+		return getParallelLoader(nThreads, 0, timeoutTime, timeoutUnit);
+	}
+
+	/**
+	 * calculates the number of threads to be used for parallel resource loading
+	 * since 2.12
+	 */
+	public static int getNumberOfThreads() {
```
kthoms: why is this public? It can't be overloaded nor clients might be interested.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
```diff
@@ -42,15 +44,39 @@ public Sorter getResourceSorter() {
...
 	}
 
 	public static Provider<IResourceLoader> getParallelLoader(final int nrOfThreads) {
 		return getParallelLoader(nrOfThreads, 0);
 	}
 
+	/**
+	 * @since 2.12
+	 */
+	public static Provider<IResourceLoader> getParallelLoader(final int nrOfThreads, long timeoutTime, TimeUnit timeoutUnit) {
```
kthoms: Do we need this convenience method? I think the last one is sufficient.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
cdietrich: if client want to create their own factory for some reasons they can reuse that logic

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
cdietrich: we dont need but can.

cdietrich: (to be symmetric with existing methods)

kthoms: That's not a strong reason. I think we should keep the codebase small here. It is unlikely that users need the customization (although you do of course) and then they would be happy without this method.

cdietrich: so you would get want to get rid of both convenciene methods?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
kthoms: For this unlikely case they could duplicate the method, or we introduce a provider. It is not a method that I would see as a method of ResourceLoaderProviders.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
kthoms: Only this one.

cdietrich: so you would vote for not doing that at all?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
cdietrich: so you would make the method private?

kthoms: yes

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ResourceLoaderProviders.java
kthoms: no, just this one method is superfluous. You can use `getParallelLoader(final int nrOfThreads, final int bufferSize, long timeoutTime, TimeUnit timeoutUnit)`

cdietrich: So you would copy and paste the 0 then?

cdietrich:
@kthoms i updated the pull request

PR #241, opened 2017-04-27
Remove a few warnings

franzbecker:
I had too many warnings in my workspace so I removed a few.

The pattern `(MyClass) getAdapter(MyClass.class);` was repeated so often I was wondering if we need this for backwards compatibility?!


cdietrich:
this will not work with luna i think.

franzbecker:
Okay I removed the commit that remove the casts.

cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/fb_remove_warnings/5/

cdietrich:
Thank you @franzbecker

PR #244, opened 2017-04-28
Fixed Problem with Models that dont have any ast created. https://github.com/eclipse/xtext-core/issues/145

cdietrich:
Fixed Problem with Models that dont have any ast created. https://github.com/eclipse/xtext/issues/2586


cdietrich:
@meysholdt what do you think? is it ok to add this guard.
the formatter2 itself seems heavily involved regarding having a root eobject

PR #246, opened 2017-05-01
[515846] getProjectConfig return null when service provider unavailable

kthoms:
null return value will be considered by
AbstractTraceForURIProvider#getTraceToTarget

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
@@ -12,6 +12,7 @@ import java.io.InputStream;
 import java.util.Collections;
 import java.util.List;
 
+import org.apache.log4j.Logger;
 import org.eclipse.core.resources.IFile;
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=515846

kthoms:
consider xtext-core#329 first

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
```diff
@@ -139,8 +140,13 @@ protected AbsoluteURI getAbsoluteLocation(IFile file) {
 	@Override
 	protected IProjectConfig getProjectConfig(IFile sourceFile) {
 		AbsoluteURI location = getAbsoluteLocation(sourceFile);
-		EclipseProjectConfigProvider projectConfigProvider = getServiceProvider(location).get(EclipseProjectConfigProvider.class);
```
cdietrich: i wonder how this can ever happen. this method should be called for methods only that xtext knows by means of a iresourceserviceprovider. did you gain any clues about this?

PR #263, opened 2017-05-11
Do Better Test what we are actually doing and thus get more stable tests

cdietrich:


org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/Bug486584Test.java
```diff
@@ -119,11 +173,30 @@ public void testFullBuildWhenClasspathChanged_3() throws CoreException, IOExcept
 		waitForBuild();
 		assertFalse(getEvents().isEmpty());
 		getEvents().clear();
+		
+		final CountDownLatch latch = new CountDownLatch(1);
+		ResourcesPlugin.getWorkspace().addResourceChangeListener(new IResourceChangeListener() {
```
LorenzoBettini: Nothing crucial, but these anonymous classes all look the same, so you can add a single implementation of the listener and pass the latch as a constructor argument, just to avoid code duplication

cdietrich:
@LorenzoBettini better this way?

LorenzoBettini:
yes indeed :)

PR #267, opened 2017-05-18
Prevent NPE with old formatter

cdietrich:
Prevent NPE with old formatter
Fixes https://github.com/eclipse/xtext-core/issues/339


cdietrich:
please review. would be nice to see this in 2.12

spoenemann:
Shouldn't do any harm...

PR #282, opened 2017-06-09
Add getOutputConfiguration function in BuilderConfigurationBlock.

gallandarakhneorg:
This function may be overridden by subtypes of `BuilderConfigurationBlock` in order to filter the output configurations that will be displayed in the preference/property pages.

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderConfigurationBlock.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderConfigurationBlock.java
@@ -47,6 +47,7 @@ import com.google.inject.Inject;
 
 /**
  * @author Michael Clay - Initial contribution and API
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderConfigurationBlock.java
```diff
@@ -91,6 +92,15 @@ protected Control doCreateContents(Composite parent) {
 		return mainComp;
 	}
 
+	/** Replies the output configurations from the given project that should be displayed by the configuration block.
+	 *
+	 * @param project the project, usually the result of {@link #getProject()}.
+	 * @return the output configurations.
+	 */
```
cdietrich: can you please add a `@since 2.13`

gallandarakhneorg:
`@since 2.13` was added as you recommended me.

cdietrich:
Thank you @gallandarakhneorg

Issue #278, opened 2017-06-07
OutlineView flickers on every Keystroke


HBuender:
The OutlineView flickers whenever it is updated by the OutlineRefreshJob. The behavior can be reproduced even with a small instance of the DomainModel sample. The flickering can be cirumvented by setting 
`getTreeViewer().getTree().setRedraw(false);` before the refresh in `protected void refreshViewer(final IOutlineNode rootNode, final Collection<IOutlineNode> nodesToBeExpanded,	final Collection<IOutlineNode> selectedNodes)`is executed. Afterwards redraw should be set back to true.

szarnekow:
Could you please create a pull request?

HBuender:
Yes, just have to update my environment ;)

cdietrich:
fixed in 2.13

PR #283, opened 2017-06-14
[#278] OutlineView flickers on every Keystroke

HBuender:
Set redraw(false) before refreshing the treeviewer to avoid flickering.
Reset to redraw(true) after refresh.

Signed-off-by: HBuender <hendrik.buender@gmail.com>


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/outline/impl/OutlinePage.java
```diff
@@ -318,6 +319,8 @@ public void run() {
 					}
 				} catch (Throwable t) {
 					LOG.error("Error refreshing outline", t);
+				}finally{
```
kthoms: just format: insert spaces around finally

Issue #292, opened 2017-07-02
EmbeddedEditorFactory.Builder API extension to set the embedded editor SWT style.


miklossy:
Currently, the EmbeddedEditorFactory creates an embedded editor using the `SWT.BORDER | SWT.V_SCROLL | SWT.H_SCROLL` SWT style. This style is hardly customizable for use cases when creating a single-line embedded editor is desired (e.g.  https://www.eclipse.org/forums/index.php/t/1076254/  https://www.eclipse.org/forums/index.php/m/1746440/ ).

Therefore, the EmbeddedEditorFactory.Builder API should be extended to be able to configure the SWT style of the embedded editor.

I will create a PR to this issue soon.

cdietrich:
fixed in 2.13

PR #293, opened 2017-07-02
[#292] Add style(int) method to the EmbeddedEditorFactory.Builder API.

miklossy:
- Extend the EmbeddedEditorFactory.Builder API with the style(int)
method to be able to configure the SWT style of the embedded editor.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/embedded/EmbeddedEditorFactory.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/embedded/EmbeddedEditorFactory.java
@@ -140,6 +140,11 @@ public class EmbeddedEditorFactory {
 		protected boolean editorBuild;
 		protected IValidationIssueProcessor issueProcessor;
 		
+		/**
+		 * @since 2.13
+		 */
+		protected int style = SWT.BORDER | SWT.V_SCROLL | SWT.H_SCROLL;
+		
 		public Builder showErrorAndWarningAnnotations() {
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/embedded/EmbeddedEditorFactory.java
```diff
@@ -139,6 +139,16 @@
 		protected Boolean readonly;
 		protected boolean editorBuild;
 		protected IValidationIssueProcessor issueProcessor;
+		protected int style = SWT.BORDER | SWT.V_SCROLL | SWT.H_SCROLL;
+		
+		/**
+		 * Configures the SWT style of the embedded editor.
+		 * @since 2.13
+		 */
+		public Builder style(int style){
```
kthoms: I think `withStyle` would be a better name. Would you rename it?
Please move the method after `withParent`, since it is the last builder method that was added.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/embedded/EmbeddedEditorFactory.java
```diff
@@ -139,6 +139,16 @@
 		protected Boolean readonly;
 		protected boolean editorBuild;
 		protected IValidationIssueProcessor issueProcessor;
+		protected int style = SWT.BORDER | SWT.V_SCROLL | SWT.H_SCROLL;
```
kthoms: Please add `@since 2.13` also to this field

kthoms:
Thanks for the contribution!

PR #302, opened 2017-07-10
Bug472993

kthoms:
This will make the "Restore Defaults" and "Apply" button on the language root page invisible, since there is nothing to reset or apply here. The ResetPromptDialogSettingsField on the page must be initialized correctly by calling init().

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/LanguageRootPreferencePage.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/LanguageRootPreferencePage.java
@@ -19,6 +19,10 @@ import com.google.inject.Inject;
  */
 public class LanguageRootPreferencePage extends AbstractPreferencePage {
 	private @Inject ResetPromptDialogSettingsField resetPromptDialogSettings;
+	
+	public LanguageRootPreferencePage () {
+		noDefaultAndApplyButton();
+	}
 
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472993

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/LanguageRootPreferencePage.java
```diff
@@ -21,6 +21,13 @@
 	private @Inject ResetPromptDialogSettingsField resetPromptDialogSettings;
 
 	@Override
+	public void createControl(Composite parent) {
+		super.createControl(parent);
+		getApplyButton().setVisible(false);
```
cdietrich: noDefaultAndApplyButton ?

kthoms: which sense do they make on the root page? They would make sense if we could reset all sub-pages, but I did not find a way to get the sub-pages of LanguageRootPreferencePage.

screenshot 48

cdietrich: Call this method instead of you code

PR #303, opened 2017-07-11
[475431] Add version constraint on org.eclipse.xtext

kthoms:
UI plugin is likely to fail with a newer runtime bundle

### DIFF
--- a/org.eclipse.xtext.ui.shared/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.ui.shared/META-INF/MANIFEST.MF
@@ -6,7 +6,7 @@ Bundle-Version: 2.13.0.qualifier
 Bundle-Localization: plugin
 Bundle-Activator: org.eclipse.xtext.ui.shared.internal.Activator
 Bundle-Vendor: Eclipse Xtext
-Require-Bundle: org.eclipse.xtext.ui;resolution:=optional;x-installation:=greedy,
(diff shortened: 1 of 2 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=475431

org.eclipse.xtext.ui.shared/META-INF/MANIFEST.MF
```diff
@@ -6,7 +6,7 @@ Bundle-Version: 2.13.0.qualifier
 Bundle-Localization: plugin
 Bundle-Activator: org.eclipse.xtext.ui.shared.internal.Activator
 Bundle-Vendor: Eclipse Xtext
-Require-Bundle: org.eclipse.xtext.ui;resolution:=optional;x-installation:=greedy,
+Require-Bundle: org.eclipse.xtext.ui;bundle-version="[0.0.0,2.14.0)";resolution:=optional;x-installation:=greedy,
```
cdietrich: if i have a look at the linked bugzilla it looks like we should require a minimal version as well (if we go this path) @szarnekow what do you think?

szarnekow: My gut feeling is, that we should always have exact matches between the xtext bundles. It's unlikely that xtext.ui will ever reliably work with another version of xtext or xtext.ui.shared.

Issue #312, opened 2017-07-25
DirtyStateResourceDescription.Manager.computeTypes(Resource) does not copy userdata


cdietrich:

cdietrich:
this might be a source of getting "red" stale editors as well

cdietrich:
is not solvable that easy: https://github.com/eclipse/xtext-eclipse/pull/313

maybe we can enhance the resourcedescriptionstrategy with a method that returns the copyable keys

PR #313, opened 2017-07-26
Copy UserData in DirtyStateResourceDescription.Manager.computeTypes(Resource) #312

cdietrich:
Copy UserData in DirtyStateResourceDescription.Manager.computeTypes(Resource) eclipse/xtext#2452

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateResourceDescription.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateResourceDescription.java
@@ -14,7 +14,9 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.log4j.Logger;
 import org.eclipse.emf.common.util.URI;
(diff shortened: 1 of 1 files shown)


cdietrich:
Tests in https://github.com/eclipse/xtext-xtend/pull/233
please review

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateResourceDescription.java
```diff
@@ -73,9 +75,13 @@ protected IResourceDescription createResourceDescription(Resource resource) {
 			IResourceDescription delegateDescription = delegate.getResourceDescription(resource);
 			List<IEObjectDescription> result = newArrayList();
 			String hash = getTextHash(resource);
-			ImmutableMap<String, String> userData = ImmutableMap.of(TEXT_HASH, hash);
 			for (IEObjectDescription eObjectDescription : delegateDescription.getExportedObjects()) {
-				result.add(new EObjectDescription(eObjectDescription.getQualifiedName(), eObjectDescription.getEObjectOrProxy(), userData));
+				Map<String,String> userData = new HashMap<String,String>(eObjectDescription.getUserDataKeys().length + 1);
+				for (String userDataKey : eObjectDescription.getUserDataKeys()) {
+					userData.put(userDataKey, eObjectDescription.getUserData(userDataKey));
```
szarnekow: Userdata may be expensive to compute. Maybe it's worthwhile to simply export a timestamp instead of the hash?

szarnekow: Is there a way to write tests for this fix?

cdietrich: see the pull request in xtext-xtend

cdietrich: i dont know the intent of the hash itself.

cdietrich: the other question is: if this affects dirty resources only, is the performance impact really that bad?

szarnekow: I'm not sure from the top of my head, but it could be, that this operation is blocking the entire resource set for read / write access. In that case, I'd rather like to avoid the overhead - iff type resolution doesn't happen anyway during within this lock.

PR #314, opened 2017-07-26
setToDefault - use Objects.equal() to compare values

kthoms:


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/FixedScopedPreferenceStore.java
```diff
@@ -703,7 +705,7 @@ public void setToDefault(String name) {
 			// removing a non-existing preference is a no-op so call the Core
 			// API directly
 			getStorePreferences().remove(name);
-			if (oldValue != defaultValue){
+			if (!Objects.equal(oldValue, defaultValue)) {
```
cdietrich: i wonder if this should be fixed in platform as well

kthoms: where in platform?

cdietrich: org.eclipse.ui.preferences.ScopedPreferenceStore.setToDefault(String)

kthoms: I do think so.

PR #322, opened 2017-08-02
Configuration of Maven tooling

kthoms:
- removed version constraint on m2e feature
- defined variables for m2e connector p2 repositories
- upgraded m2e tycho connector to 0.9.0-LATEST

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -83,6 +83,14 @@
       value="http://download.eclipse.org/buildship/updates/e46/releases/2.x/"/>
   <setupTask
       xsi:type="setup:VariableTask"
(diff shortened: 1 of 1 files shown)


releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -281,8 +289,7 @@
       xsi:type="setup.p2:P2Task"
       label="Tools">
     <requirement
-        name="org.eclipse.m2e.feature.feature.group"
-        versionRange="[1.0.0,1.7.0)"/>
```
cdietrich: having no range at all is intended?

kthoms: I do not see the sense in it. Current version is 1.8.0 and works fine.

PR #323, opened 2017-08-03
Bug#438625 [EclipseResourceFileSystemAccess2] fileBasedTraceInformation should be of type ITraceForStorageProvider

kthoms:
EclipseResourceFileSystemAccess was injecting the implementation class
TraceForStorageProvider and could not inject the interface, since 2
methods on the implementation class were used that are not part of the
interface. For these methods ITraceForStorageProviderExtension was
created.

This allows disabling writing trace files by binding
ITraceForStorageProvider.Null (usually performance reason when no trace
is required).

Also cleaned up unused field `workspace`.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
@@ -180,12 +180,14 @@ public class TraceForStorageProvider extends AbstractTraceForURIProvider<IFile, 
 	/**
 	 * @since 2.3
 	 */
+	@Override
 	public IFile getTraceFile(IFile generatedFile) {
 		String originLastSegment = generatedFile.getFullPath().lastSegment();
 		IFile traceFile = generatedFile.getParent().getFile(new Path(getTraceFileNameProvider().getTraceFromJava(originLastSegment)));
 		return traceFile;
 	}
 
+	@Override
 	public boolean isTraceFile(IStorage storage) {
 		if (!(storage instanceof IFile)) {
 			return false;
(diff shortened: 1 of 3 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=438625

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/ITraceForStorageProviderExtension.java
```diff
@@ -0,0 +1,36 @@
...
+	/**
+	 * Returns the trace file for the given derived resource. May return
+	 * <code>null</code> if no such file is available for the given storage.
+	 * 
+	 * @param generatedFile
+	 *            the resource whose trace is requested. May not be <code>null</code>.
+	 * @return the trace file for <code>generatedFile</code> or <code>null</code>.
+	 */
+	/*@Nullable*/
+	IFile getTraceFile(/*@NonNull*/ IFile generatedFile);
```
szarnekow: Could these become default methods on the ITraceForStorageProvider? This would greatly simplify the implementation.

kthoms: Good idea. My first use case for default methods :)

kthoms:
Updated. Please review.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/generator/trace/TraceForStorageProvider.java
```diff
@@ -36,7 +36,7 @@
  * @noextend This class is not intended to be subclassed by clients.
  * @noinstantiate This class is not intended to be instantiated by clients.
  */
-public class TraceForStorageProvider extends AbstractTraceForURIProvider<IFile, StorageAwareTrace> implements ITraceForStorageProvider {
+public class TraceForStorageProvider extends AbstractTraceForURIProvider<IFile, StorageAwareTrace> implements ITraceForStorageProvider, ITraceForStorageProviderExtension {
```
szarnekow: Is this change correct?

kthoms:
Done

Issue #329, opened 2017-08-17
"Runtime bundle is NEWER than UI bundles" may be false positive


meysholdt:
Just saw this during some Tycho build.
```
The installed Xtext bundle versions are:
- org.eclipse.xtext 2.13.0.v20170816-0516 [ACTIVE]
- org.eclipse.xtext.ide 2.13.0.v20170816-0516 [ACTIVE]
- org.eclipse.xtext.ui 2.13.0.v20170808-0854 [ACTIVE]
- org.eclipse.xtext.ui.shared 2.13.0.v20170808-0854 [STARTING]
Runtime bundle is NEWER than UI bundles! Please make sure that both bundles are installed with the same version!
```

The conclusion `Runtime bundle is NEWER than UI bundles!` is a false positive since our Xtext umbrella build repackages the artifacts from the `xtext-core` and `xtext-eclipse` builds. Therefore the artifacts do most likely not have the exactly same version qualifier.

cdietrich:
cc @kthoms

cdietrich:
no i think we should digg why guice fails.

cdietrich:
do you have a complete log for the build @meysholdt

meysholdt:
>  i think we should digg why guice fails as well
A user bound a new class in the shared state module which threw an exception in the constructor :)

cdietrich:
ah ok so this is really about the error message. we did not screw up the module loading in xtext itself

meysholdt:
So this is the cause: eclipse/xtext#2453. But message and cause remain independent issues.

cdietrich:
Fixed in Xtext 2.13

PR #331, opened 2017-08-17
"Runtime bundle is NEWER than UI bundles" may be false positive #329

cdietrich:
"Runtime bundle is NEWER than UI bundles" may be false positive #329


cdietrich:
please review

PR #335, opened 2017-08-24
[472992] Add null guard when position could be retrieved for annotation

kthoms:
IAnnotationModel.getPosition() might return null. This would later lead
to invalid proposals with null position and to NPE when accessing them.
Also guarded arguments to QuickAssistCompletionProposal.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=472992

PR #336, opened 2017-08-29
incr. version range for m2e

cdietrich:

### DIFF
--- a/org.eclipse.xtext.m2e/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.m2e/META-INF/MANIFEST.MF
@@ -12,7 +12,7 @@ Require-Bundle: org.eclipse.m2e.core;bundle-version="1.0.0";resolution:=optional
  org.eclipse.xtend.lib,
  org.eclipse.xtext.builder,
  org.eclipse.xtext.ui,
- org.eclipse.m2e.maven.runtime;bundle-version="[1.0.0,1.6.0)";resolution:=optional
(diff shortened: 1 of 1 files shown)


cdietrich:
please review

org.eclipse.xtext.m2e/META-INF/MANIFEST.MF
```diff
@@ -12,7 +12,7 @@ Require-Bundle: org.eclipse.m2e.core;bundle-version="1.0.0";resolution:=optional
  org.eclipse.xtend.lib,
  org.eclipse.xtext.builder,
  org.eclipse.xtext.ui,
- org.eclipse.m2e.maven.runtime;bundle-version="[1.0.0,1.6.0)";resolution:=optional
+ org.eclipse.m2e.maven.runtime;bundle-version="[1.0.0,1.8.0)";resolution:=optional
```
kthoms: Do we need the constraint at all?

cdietrich: no idea

kthoms: I don't think that we need it. It worked with 1.6 and also 1.8. Likely 1.8+ also. From my point of view you could remove it or make an upper bound to 2.0.0.

cdietrich: updated

PR #340, opened 2017-08-30
[473623] Xbase creates Imports when refactoring even if the language has no XImport Section Support

cdietrich:
[473623] Xbase creates Imports when refactoring even if the language has no XImport Section Support


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=473623

PR #349, opened 2017-09-05
[#348] Avoid recreation of markers when no change happened

### LINKED ISSUE #348: Build loop by TaskMarkerContributor
Today I observed a build loop in a customer project. After saving a DSL file the XtextBuilder was called recurrent with a resource delta for the changed resource.

After debugging a while it turned out that the resource change was occurring while recreation of task markers during a refresh.

To solve this issue TaskMarkerContributor should only delete and create markers when there is actual a change to the existing markers.

kthoms:
Fixes #348

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.xtend
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.xtend
@@ -11,6 +11,7 @@ import com.google.inject.Inject
 import java.util.List
 import org.apache.log4j.Logger
 import org.eclipse.core.resources.IFile
+import org.eclipse.core.resources.IMarker
 import org.eclipse.core.resources.IResource
 import org.eclipse.core.runtime.CoreException
 import org.eclipse.core.runtime.IProgressMonitor
@@ -19,14 +20,14 @@ import org.eclipse.emf.ecore.resource.Resource
 import org.eclipse.xtext.tasks.ITaskFinder
 import org.eclipse.xtext.tasks.Task
 import org.eclipse.xtext.ui.markers.IMarkerContributor
+import org.eclipse.xtext.tasks.Priority
 
 /**
  * @author Stefan Oehme - Initial contribution and API
  * @since 2.6
  */
 class TaskMarkerContributor implements IMarkerContributor {
-
-	static final val log = Logger.getLogger(TaskMarkerContributor);
+	static val LOG = Logger.getLogger(TaskMarkerContributor);
 
 	@Inject
 	TaskMarkerCreator markerCreator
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.xtend
```diff
@@ -43,10 +44,39 @@ class TaskMarkerContributor implements IMarkerContributor {
...
+				createTaskMarkers(file, tasks, monitor);
+			}
 		} catch (CoreException e) {
-			log.error(e.getMessage(), e);
+			LOG.error(e.getMessage(), e);
+		}
+	}
+	
+	private def boolean match (Task task, IMarker marker) {
+		marker.getAttribute(IMarker.LOCATION) == "line " + task.lineNumber
```
kittaakos: Just in case you have to change something else too; you could move down this to the very end right after `CHAR_END` and pull `USER_EDITABLE` up. The first check concatenates a string with an integer which might be unnecessary... I know this is extremely minor performance tweak.

kittaakos: Ahh, ignore my comment and sorry for the noise.

kthoms:
@szarnekow good point, updated.

cdietrich:
@kthoms what do you think of https://github.com/eclipse/xtext-eclipse/compare/cd_issue348

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.xtend
```diff
@@ -43,10 +44,41 @@ class TaskMarkerContributor implements IMarkerContributor {
...
+				createTaskMarkers(file, tasks, monitor);
+			}
 		} catch (CoreException e) {
-			log.error(e.getMessage(), e);
+			LOG.error(e.getMessage(), e);
+		}
+	}
+	
+	private def boolean match (Task task, IMarker marker) {
+		val attributes = marker.getAttributes(#[IMarker.LOCATION, IMarker.PRIORITY, IMarker.MESSAGE, IMarker.LINE_NUMBER, IMarker.CHAR_START, IMarker.CHAR_END, IMarker.USER_EDITABLE]);
```
szarnekow: Please extract constant array

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.xtend
```diff
@@ -43,10 +44,41 @@ class TaskMarkerContributor implements IMarkerContributor {
...
+		return 
+		   attributes.get(0) == "line " + task.lineNumber
+		&& attributes.get(1) == getPriority(task.tag.priority)
+		&& attributes.get(2) == task.fullText
+		&& attributes.get(3) == task.lineNumber
+		&& attributes.get(4)  == task.offset
```
szarnekow: Whitespace

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerContributor.xtend
```diff
@@ -43,10 +44,41 @@ class TaskMarkerContributor implements IMarkerContributor {
...
+				deleteMarkers(file, monitor);
+				createTaskMarkers(file, tasks, monitor);
+			}
 		} catch (CoreException e) {
-			log.error(e.getMessage(), e);
+			LOG.error(e.getMessage(), e);
+		}
+	}
+	
+	private def boolean match (Task task, IMarker marker) {
```
szarnekow: Whitespace after method name

kthoms:
@cdietrich If you really don't want the default impl be extended... I still think that this extension of the default impl would not harm. Why is it bad to merge by default? Admitted, we still don't know how the deletion/recreation causes resource changes. I still prefer to change TaskMarkerContributor.xtend.
@szarnekow What do you think?

cdietrich:
i still prefer to have two classes and the developer to opt in

kthoms:
But it is hard to figure out when to use the other implementation. Users might recognize a build loop, but do not have any clue what to do. And may even have no influence. DSL users might run into this issue only in certain environments.

cdietrich:
yes but i still dont think that the change fixes anything with the buildlop but just covers the real bug under the carpet

szarnekow:
I fear the same. Marker creation will usually not cancel / restart the build. There must be something else going on in the errorneous case.

kthoms:
Sure, but how much would we invest into finding the root cause? I can only report that this fixed the problem on customer's side. It may cover the cause, but in effect the user does not care as long as he don't experience a problem. And I fear that many DSLs won't make use of the extended class since they usually won't have an issue. And if they have it then it might be too late to change.

szarnekow:
Is this PR still relevant?

cdietrich:
@kthoms is it ok if we close this PR?

kthoms:
We can close the PR but leave the branch. It is likely that the issue still exists.

PR #351, opened 2017-09-06
[465642] Handle IAE thrown by ASTNode#setSourceRange

kthoms:
The IAE might be thrown when the source is missing. By catching the
exception more details of the actual context is available by the thrown
ISE.

This does not resolve the cause, but will give a better message to the user and might help to handle the situation better in the future.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=465642

PR #353, opened 2017-09-07
removed deprecated popup menu extension point

cdietrich:


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/nature/ToggleXtextNatureCommand.java
```diff
@@ -0,0 +1,86 @@
...
+	private static final Logger log = Logger.getLogger(ToggleXtextNatureCommand.class);
+
+	@Override
+	public Object execute(ExecutionEvent event) throws ExecutionException {
+		ISelection selection = HandlerUtil.getCurrentSelection(event);
+		if (selection instanceof IStructuredSelection) {
+			for (Iterator<?> it = ((IStructuredSelection) selection).iterator(); it.hasNext();) {
+				Object element = it.next();
+				IProject project = null;
+				if (element instanceof IProject) {
```
kthoms: IProject extends IAdaptable and should be fine to use the second clause only

cdietrich: did just stupid copy&paste

org.eclipse.xtext.ui.shared/plugin.properties
```diff
@@ -4,6 +4,8 @@ XtextBuilder.name=Xtext Project Builder
 XtextNature.name=Xtext Project Nature
 XtextNature.description=Adds an icon decoration to projects with Xtext enabled.
 XtextNatureAddAction.label=Convert to Xtext Project
+XtextNatureAddCommand.label=Convert to Xtext Project
```
kthoms: I know there is a difference between the deprecated action and the new command, but shall we just reuse the message? It's just a "wrong" key name. What do you think?

cdietrich: i think i even can remove the old key

kthoms:
Besides the 2 comments OK.

org.eclipse.xtext.ui.shared/plugin.properties
cdietrich: updated

Issue #355, opened 2017-09-07
migrate CodetemplateLanguage to new generator


cdietrich:

cdietrich:
fixed in Xtext 2.13

PR #356, opened 2017-09-07
migrated wf to new generator #355

cdietrich:
migrated wf to new generator #355


cdietrich:
https://github.com/eclipse/xtext-umbrella/pull/37

PR #358, opened 2017-09-08
[refactoring] change for read-only files

JanKoehnlein:


cdietrich:
@JanKoehnleincould this/a similar fix be applied for https://bugs.eclipse.org/bugs/show_bug.cgi?id=427566 ?

JanKoehnlein:
This issue is about the new refactoring infrastructure. But a similar check could be done in the old one as well. ```org.eclipse.xtext.ui.refactoring.impl.DefaultRefactoringDocumentProvider.get(URI, StatusWrapper)``` could be an appropriate location.

Or do you think I should also check for derived files?

cdietrich:
i think it would be a good point at least to warn. dont know if we can check if the derived file is ok to refactor or not

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/participant/ChangeConverter.xtend
```diff
@@ -63,19 +64,24 @@ class ChangeConverter implements IAcceptor<IEmfResourceChange> {
 	protected def dispatch void handleReplacements(IEmfResourceChange change) {
 		val outputStream = new ByteArrayOutputStream
 		tryWith(outputStream) [
+			val file = change.resource.URI.toFile
+			if (file.isReadOnly) 
+				issues.add(Severity.ERROR, '''Affected file '�file.fullPath�' is read-only''')
```
szarnekow: I wonder if this should be customizable, .e.g in Eclipse you would want to use `org.eclipse.core.resources.IWorkspace.validateEdit(IFile[], Object)` instead of a plain readonly check

JanKoehnlein: Hmm, I would have expected that to check for _derived_ as well, but it doesn't. Anyway good hint!

JanKoehnlein:
I've added checking for derived files as well. 

Unfortunately, the same approach for the rename refactoring infrastructure doesn't work and results in "No suitable storage found for resource XXX" error messages instead of my custom ones.

Issue #360, opened 2017-09-11
Add an Option to disable support for secondary types in JdtTypeProvider


cdietrich:
Add an Option to disable support for secondary types in JdtTypeProvider
(similar to org.eclipse.xtext.naming.QualifiedName.USE_INTERNING)

Secondary type handling is quite a bottleneck in xbase/xtend inside eclipse.
we should consider to add a flag to disable the support for it (findSecondaryType returning null)

cdietrich:
Fixed in Xtext 2.13

PR #361, opened 2017-09-11
introduced property to make the support of secondary types skippable #360

cdietrich:
introduced property to make the support of secondary types skippable #360


cdietrich:
please review

cdietrich:
anyone?

cdietrich:
which call site? the JdtTypeProvider is the call site. there are tons of methods calling each other. if you know you dont have secondary types i simply dont want to try to search for them

svenefftinge:
Is it only called here? 
https://github.com/eclipse/xtext-eclipse/pull/361/files#diff-0cb9deaaf914735c46acc5512b84bf07L324
Then I would add the check to the if guard.

cdietrich:
of will move the guard

cdietrich:
@svenefftinge i changed the code

Issue #365, opened 2017-09-15
Impossible to reuse XbaseProposalProvider.getProposalFactory


OLibutzki:
I have an Xbase based language and would like to customize the ProposalFactory. I could override org.eclipse.xtext.xbase.ui.contentassist.XbaseProposalProvider.getProposalFactory(String, ContentAssistContext), but it's impossible to extend the ProposalCreator returned by XbaseProposalProvider.getProposalFactory as it's an anonymous class.

It would be better to have a XbaseProposalCreator which extends DefaultProposalCreator. This solution provides the possibility to extend XbaseProposalCreator and fallback to its implementation.

OLibutzki:
Ok, I could use delegation instead:
```
override protected getProposalFactory(String ruleName, ContentAssistContext contentAssistContext) {
	val superProposalFactory = super.getProposalFactory(ruleName, contentAssistContext)
	return new DefaultProposalCreator(contentAssistContext, ruleName, getQualifiedNameConverter()) {
		
		override apply(IEObjectDescription candidate) {
			// might create a proposal
			// [...]
			// fallback
			return superProposalFactory.apply(candidate)
		}
	}
}
```
Anyway, implementing an anonymous class with much more than 100 lines of code might not be the best solution.

cdietrich:
DefaultProposalCreator is a protected inner class so it wont be visible => XbaseProposalCreator would need to be a inner class as well

OLibutzki:
@cdietrich exactly

cdietrich:
fixed in Xtext 2.13

PR #369, opened 2017-09-20
introduced XbaseProposalCreator

cdietrich:
introduced XbaseProposalCreator instead of anonymous inner subclass of DefaultProposalCreator to allow better extensibility #365

### DIFF
--- a/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseProposalProvider.java
+++ b/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseProposalProvider.java
@@ -128,6 +128,155 @@ public class XbaseProposalProvider extends AbstractXbaseProposalProvider impleme
 	@Inject
 	private SyntaxFilteredScopes syntaxFilteredScopes;
 	
+	protected class XbaseProposalCreator extends DefaultProposalCreator {
+		
+		private ContentAssistContext contentAssistContext;
+		private String ruleName;
+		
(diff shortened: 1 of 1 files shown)


cdietrich:
@OLibutzki is this what you thought of?

OLibutzki:
@cdietrich, yes, it's possible to extend XbaseProposalCreator and override `apply(IEObjectDescription)` in order to add some special behaviour and call `super.apply(candidate)` as fallback. It's just the same pattern XbaseProposalCreator and DefaultProposalCreator use.

Thanks for "fixing" this that quickly!

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseProposalProvider.java
```diff
@@ -128,6 +128,155 @@ public boolean apply(IEObjectDescription input) {
 	@Inject
 	private SyntaxFilteredScopes syntaxFilteredScopes;
 	
+	protected class XbaseProposalCreator extends DefaultProposalCreator {
```
kthoms: maybe static inner class or not?

cdietrich: parent is not static as well and it and the sub use stuff from outer context

kthoms: just wanted to make sure that this was a decision that way

PR #383, opened 2017-09-27
Regenerated Xbase & PureXbase

kthoms:
see also eclipse/xtext-extras#182

### DIFF
--- a/org.eclipse.xtext.purexbase.eclipse.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.purexbase.eclipse.tests/META-INF/MANIFEST.MF
@@ -17,7 +17,8 @@ Require-Bundle: org.eclipse.xtext.purexbase,
  org.eclipse.xtext.xbase.lib,
  org.eclipse.xtext.testing,
  org.eclipse.xtext.xbase.testing,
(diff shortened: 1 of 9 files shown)


org.eclipse.xtext.purexbase.eclipse.tests/META-INF/MANIFEST.MF
```diff
@@ -17,7 +17,8 @@ Require-Bundle: org.eclipse.xtext.purexbase,
  org.eclipse.xtext.xbase.lib,
  org.eclipse.xtext.testing,
  org.eclipse.xtext.xbase.testing,
- org.eclipse.xtext.ui.testing
+ org.eclipse.xtext.ui.testing,
+ org.eclipse.xtext.junit4
```
cdietrich: maybe we need a change to the generator to skip this.

kthoms: yes, but that is what the generator currently produces. doesn't harm, but can be improved.

cdietrich: yes this is why i propose to make this choosable in the generator.

Issue #384, opened 2017-09-28
Add Photon Section to Xtext.setup


cdietrich:
- consider consuming Emf 2.14

PR #389, opened 2017-09-29
[#384] Add Photon to Xtext targlet's repository lists

kthoms:
Also made MPC optional since to available on Photon M2 and it is not a
mandatory feature for the development environment.


kthoms:
Why should we? I think it is better to use releases of depending plugins unless we have to use a feature.

cdietrich:
well it would at least allow us to tests the stuff ed willink wants to change

Issue #393, opened 2017-10-02
Photon: JavaProject.computePackageFragmentRoots api change


cdietrich:
JavaProject.computePackageFragmentRoots signature changs with photon.
https://github.com/eclipse/eclipse.jdt.core/commit/0201874afdd1b68acce876105bbdf3cc69d05ccc

another issue is at removed org.eclipse.xtext.xbase.compiler.OnTheFlyJavaCompiler.PatchedFileSystem.findType(char[][], boolean)
 plus tons of new methods in filesystem

cdietrich:
breaks org.eclipse.xtext.common.types.access.jdt.JdtTypeProvider.collectSourcePackageFragmentRoots(JavaProject, HashSet<String>, IClasspathEntry, ObjectVector)

cdietrich:
@szarnekow what is the intention of that code

cdietrich:
another issue is at org.eclipse.xtext.xbase.compiler.OnTheFlyJavaCompiler.PatchedFileSystem.findType(char[][], boolean)

cdietrich:
Options for TypeProviderIssue

(a) reflection

```
case IClasspathEntry.CPE_SOURCE:
					try {
						Method declaredMethod = JavaProject.class.getDeclaredMethod("computePackageFragmentRoots", IClasspathEntry.class, ObjectVector.class, HashSet.class,IClasspathEntry.class, boolean.class, Map.class);
						declaredMethod.invoke(javaProject, entry,
							result,
							rootIDs,
							referringEntry,
							true,
							null);
					} catch (NoSuchMethodException e) {
						try {
						Method declaredMethod = JavaProject.class.getDeclaredMethod("computePackageFragmentRoots", IClasspathEntry.class, ObjectVector.class, HashSet.class,IClasspathEntry.class, boolean.class, boolean.class, Map.class);
						declaredMethod.invoke(javaProject, entry,
							result,
							rootIDs,
							referringEntry,
							true,
							true,
							null);
						} catch (NoSuchMethodException e2) {
							LOG.error("did not find any 'computePackageFragmentRoots' method", e2);
						} catch (IllegalAccessException e2) {
							LOG.error("could not invoke 'computePackageFragmentRoots' method", e2);
						} catch (IllegalArgumentException e2) {
							LOG.error("could not invoke 'computePackageFragmentRoots' method", e2);
						} catch (InvocationTargetException e2) {
[...]

PR #395, opened 2017-10-03
fixed compatibility issues with eclipse oxygen 1a / photon #393

cdietrich:
fixed compatibility issues with eclipse oxygen 1a / photon #393

### DIFF
--- a/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/access/jdt/JdtTypeProvider.java
+++ b/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/access/jdt/JdtTypeProvider.java
@@ -13,6 +13,7 @@ import java.util.HashSet;
 import java.util.List;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IWorkspaceRoot;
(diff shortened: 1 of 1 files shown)


cdietrich:
please review. if you prefer a reflective solution i can create a pr for that as well

cdietrich:
am not sure how to test that. how to have secondary types in eclipse

cdietrich:
@svenefftinge no i think we can throw away the code completely and do

cdietrich:
https://github.com/eclipse/xtext-eclipse/compare/cd_issue393c

cdietrich:
see https://github.com/eclipse/xtext-eclipse/pull/396

svenefftinge:
But the comment somehow seems alarming. If the `resolveClasspath` is done too eagerly and early it might harm. Did you check with @szarnekow why he did it?

cdietrich:
one reason was the jdk version thingythat is fixed in Luna SR2. for the others i dont know

svenefftinge:
Ok, so the current code avoids opening jars at all, while the change in #396 would do. The comment mentions the jdk version, so if that was fixed we might indeed go with removing all that code. I assume that the classpath gets resolved anyhow and since that information is cached, it might not be harmful to do it. OTOH it could be that we call this stuff before JDT does and maybe even worse in the UI thread.
I think this one is less-risky, but the other proposal better in the long run. I'm going with #396. If it is too bad we can restore this code.

cdietrich:
i am still doing some measurements regarding performance.

cdietrich:
i see a performance drop 1:55->2:05 with the getAllPackageFragmenteRoots solution.

cdietrich:
bildschirmfoto 2017-10-04 um 09 58 51
bildschirmfoto 2017-10-04 um 09 58 27

cdietrich:
so i propose to stick with this PR

cdietrich:
a reflective solution for PatchedFileSystem seems no to be possible or we would even use reflection for the method signature types. but for return types? ....

cdietrich:
replaced by https://github.com/eclipse/xtext-eclipse/pull/397

PR #396, opened 2017-10-04
made jdt type provider oxygen compatible #393

cdietrich:


cdietrich:
no it does not. it removes the workaround that required copy&pasting jdt code

cdietrich:
closed cause of the performance implications

PR #404, opened 2017-10-09
[quickfix] Add a possibility to implement multi quick-fixes

dhuebner:
New adapter is based on WorkbenchMarkerResolution and allows the resolutions be grouped
 with others that are similar to allow multi selection.

Implemented Xtext's EMPTY_ENUM_LITERAL quick-fix using the new API.

bildschirmfoto 2017-10-09 um 14 10 24


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/MarkerResolutionGenerator.java
```diff
@@ -88,19 +93,28 @@ public boolean hasResolutions(IMarker marker) {
...
-			return emptyResult;
-		
 		Issue issue = getIssueUtil().createIssue(marker);
-		if(issue == null)
+		if (issue == null)
 			return emptyResult;
-		final Iterable<IssueResolution> resolutions = getResolutions(issue, editor.getDocument());
-		return getAdaptedResolutions(Lists.newArrayList(resolutions));
+		Iterable<IssueResolution> resolutions = getResolutionProvider().getResolutions(issue);
+		boolean isMutiFix = StreamSupport.stream(resolutions.spliterator(), false)
```
JanKoehnlein: typo isMultiFix

kthoms:
Could you squash the commits, please?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/MarkerResolutionGenerator.java
dhuebner: Thanks. Fixed

meysholdt:
Some thoughts from offline discussions:
a) an existing semantic quick-fix may crash as a multi-fix, since it may do weird things. Therefore multi-fix should be opt-in, not default.
b) when issue data contains FragmentURIs or FQNs, those need to be resolved before the first multi-fix is applied.
c) a single quick-fix impl should be executable in the editor, on multiple markers, or both. Making it available in the editor and in the Problems View should not require redundant declarations/implementations.

We can fix (b) by having a two-stage hook: stage one: prepare context. stage two: do the actual modification. stage one of all quick-fixes must be executed before the first stage two is executed. With this mechanism it can be guaranteed that FragmentURIs and Names do still resolve.
```
@Fix(EMPTY_ENUM_LITERAL)
  public void fixEmptyEnumLiteral(final Issue issue, IssueResolutionAcceptor acceptor) {
    acceptor.accept(issue, "Fix empty enum literal", "Fix empty enum literal", NULL_QUICKFIX_IMAGE, (FixContext ctx) -> {
      // 1. get EObject from context
      EnumLiteralDeclaration enumLiteralDeclaration = (EnumLiteralDeclaration) ctx.getElement();
      // 2. resolve EObject from issue Data
      EObject obj2 = enumLiteralDeclaration.eResource().getResourceSet().getEObject(issue.getData("foo"));
      // 3. do the actual modification. 
      ctx.addModification(enumLiteralDeclaration, (EnumLiteralDeclaration decl) -> {
      	Keyword keyword = XtextFactory.eINSTANCE.createKeyword();
        keyword.setValue(enumLiteralDeclaration.getEnumLiteral().getName().toLowerCase());
        enumLiteralDeclaration.setLiteral(keyword);      
      });
    });
  }
```

For (a), we don't need the `@MultiFix` annotation because we already have a new `acceptor.accept(...FixContext)` method. Only quick-fixes that make use of the `FixContext` should be available as a multi-fix. Having two annotations `@MultiFix` and `@Fix` makes it very inconvenient to support the three cases: quickfix-in-editor-only, multifix-only, both.

For (c) we could introduce `IssueResolutionAcceptor#isEditor()`, `IssueResolutionAcceptor#isMultipleMarkers()`. This would allow toolsmiths to programmatically distinguish when a fix should be shown.

dhuebner:
@meysholdt 
The new commit allows users to write the following quickfix:

```
@Fix(EMPTY_ENUM_LITERAL)
public void fixEmptyEnumLiteral(final Issue issue, IssueResolutionAcceptor acceptor) {
	acceptor.accept(issue, "Fix empty enum literal", "Fix empty enum literal", NULL_QUICKFIX_IMAGE, (FixContext ctx) -> {
		// 1. get EObject from context
		EnumLiteralDeclaration enumLiteralDeclaration = (EnumLiteralDeclaration) ctx.getElement();
		// 2. resolve EObject from issue Data
		EObject obj2 = enumLiteralDeclaration.eResource().getResourceSet().getEObject(issue.getUriToProblem(), false);
		// 3. do the actual modification. 
		ctx.setModification(obj2, (EObject decl) -> {
			EnumLiteralDeclaration enumToModify = (EnumLiteralDeclaration) decl;
			Keyword keyword = XtextFactory.eINSTANCE.createKeyword();
			keyword.setValue(enumToModify.getEnumLiteral().getName().toLowerCase());
			enumToModify.setLiteral(keyword);
		});
	});
}
```

szarnekow:
Does the API allow to write `ctx.setModification(obj2, (EnumLiteralDeclaration decl) -> {..}` instead of `ctx.setModification(obj2, (EnumLiteralDeclaration decl) -> {..}`, e.g. to avoid the casting?

dhuebner:
No, I doesn't use type parameter

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/WorkbenchMarkerResolutionAdapter.xtend
```diff
@@ -0,0 +1,130 @@
...
+	override void run(IMarker marker) {
+		if (!marker.exists) {
+			return
+		}
+		val resolutionData = marker.resolution
+		run(resolutionData.key, resolutionData.value, new NullProgressMonitor)
+	}
+
+	def collectResolutions(IMarker... markers) {
+		markers.map[marker|marker.resolution].filterNull.toList
```
szarnekow: Can we make this cancellable in case of plenty of markers / expensive resolution computaiton?

dhuebner: Similar to the CancelableResolutionAcceptor or using the progress monitor?

szarnekow: I think using the progress monitor that is available in the super type on the run(..) method would be good enough - it should be checked after each marker.

dhuebner: @szarnekow I was looking for a good test class to test the different kinds of quick-fix implementations, especially the Pre-Initialazed one, it should work with markers and custom QuickFixProvider implementation. Do you know a good one in xtext.ui.tests or xtext.xtext.ui.tests?

szarnekow: Hey @dhuebner , unfortunately I'm not aware of a great utility to test quickfixes.

dhuebner:
A simple and a flexible/advanced API is now available in acceptor:

```
@Fix(QuickfixCrossrefTestLanguageValidator.MULTIFIXABLE_ISSUE)
public void addDocumentation(final Issue issue, IssueResolutionAcceptor acceptor) {
	acceptor.acceptMulti(issue, "Multi fix", "Multi fix", null, (IMultiModificationContext ctx) -> {
		EObject toModify = ctx.getElement();
		// do other things here
		ctx.setModification(toModify, (obj) -> {
			((Element) obj).setDoc("Better documentation");
		});
	});
}

@Fix(QuickfixCrossrefTestLanguageValidator.MULTIFIXABLE_ISSUE_2)
public void addDocumentation2(final Issue issue, IssueResolutionAcceptor acceptor) {
	acceptor.acceptMulti(issue, "Multi fix 2", "Multi fix 2", null, (EObject obj) -> {
		((Element) obj).setDoc("Even Better documentation");
	});
} 
```

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/edit/IMultiModificationContext.java
```diff
@@ -0,0 +1,22 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ui.editor.model.edit;
+
+import org.eclipse.emf.ecore.EObject;
+
+/**
+ * @author dhuebner - Initial contribution and API
+ */
+public interface IMultiModificationContext {
```
cdietrich: missing `@since 2.13`

dhuebner: Thanks

dhuebner:
@meysholdt I have a feeling the the last xtext-core changes breaks the xtext-eclipse tests. On my machine all the failing tests are passing

meysholdt:
yes, the test failures were caused by code from xtext-core

PR #409, opened 2017-10-11
Fixed NPE in ShowQuickOutlineActionHandler

tivervac:
Can be triggered by opening "Quick Outline" in an Xtend file through the quick access.


cdietrich:
could reproduce problem and verify the fix. thx for the contribution

PR #413, opened 2017-10-12
Try to make falky test more robust

meysholdt:
Since the test asserts the contents of the "Index" (ClusteringBuilderState) and I don't see explicit invocation of the Eclipse Builder - maybe the test is flaky because autobuild *sometimes* is late.

It adds to the situation that multiple tests create test files with the same name and one test asserts that this file isn't in the index.

Anyway, this is just a guess, since I can't reproduce the failure locally.


meysholdt:
build is green :) let's hope this wasn't flaky coincidence.

PR #414, opened 2017-10-12
Don't use autobuild in tests to avoid flakyness

meysholdt:

### DIFF
--- a/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/refactoring/AbstractResourceRelocationTest.xtend
+++ b/org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/refactoring/AbstractResourceRelocationTest.xtend
@@ -37,6 +37,7 @@ abstract class AbstractResourceRelocationTest {
 	
 	@Before
 	def void setup() {
+		autobuild = false
 		project = createProject('test')
 		addNature(project, XtextProjectHelper.NATURE_ID)
 		addBuilder(project, XtextProjectHelper.BUILDER_ID)
@@ -45,6 +46,7 @@ abstract class AbstractResourceRelocationTest {
 	@After
 	def void teardown() {
 		project.delete(true, true, new NullProgressMonitor)
+		autobuild = true
 	}
 	
 	protected def file(String name, CharSequence content) {
@@ -80,6 +82,7 @@ abstract class AbstractResourceRelocationTest {
 		val change = refactoring.createChange(new NullProgressMonitor)
 		change.perform(new NullProgressMonitor)
 		project.refreshLocal(IResource.DEPTH_INFINITE, null)
+		project.build(IncrementalProjectBuilder.FULL_BUILD, null)
 	}
 	
 }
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.ui.tests/src-longrunning/org/eclipse/xtext/ui/tests/refactoring/AbstractResourceRelocationTest.xtend
```diff
@@ -80,6 +82,7 @@ abstract class AbstractResourceRelocationTest {
 		val change = refactoring.createChange(new NullProgressMonitor)
 		change.perform(new NullProgressMonitor)
 		project.refreshLocal(IResource.DEPTH_INFINITE, null)
+		project.build(IncrementalProjectBuilder.FULL_BUILD, null)
```
szarnekow: Isn't this was org.eclipse.xtext.ui.testing.util.IResourcesSetupUtil.waitForBuild() is supposed to do?

meysholdt: true... that *could* be changed, but I actually like `project.build()` better since it expresses more clearly that we're synchronously running the build in the same thread. `waitForBuild()` sounds more like a method that's blocking until a job *running asynchronously in another thread* has finished. But that's what `waitForAutoBuild()` does. 

For the record, `waitForBuild()` wraps `ResourcesPlugin.getWorkspace().build(IncrementalProjectBuilder.INCREMENTAL_BUILD`

szarnekow: I suggest to not use yet another abstraction to handle builds in tests. It is just confusing and prone to introduce technical dept. Either the old abstraction doesn't cut it anymore or the existing one is good enough.
Why would you want to do a full build when an incremental build is sufficient?

szarnekow: But of course it is only a test ...

meysholdt: yes, it's just a test. I wasn't passionate about it when I changed it. My choice for `project.build()` actually came from the fact that a few lines above it was already used and I didn't want to mix styles in the same method. 

To pick up the discussions about abstractions here... I'm happy to see it as Eclipse Resource API plus Xtext Utilities. The Eclipse Resource API is something developers need to be familiar with anyway so I don't see any value in hiding it for the sake of hiding it. But utilities can definitely be useful to enforce best practices: 
* progress monitor: null or NullImpl?
* refresh before build?
* incremental build or full build?

Here I don't mind if we do a full build or a incremental build. The test project is so small that I don't see that it makes a difference in performance. Potentially the full build is more robust due to lower implementation complexity.

so.. merging to proceed with the RC.

PR #428, opened 2017-10-19
[526246] handle cancellation in ReferenceQuery

cdietrich:
[526246] handle cancellation in ReferenceQuery

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/ReferenceQuery.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/ReferenceQuery.java
@@ -12,11 +12,13 @@ import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.common.util.WrappedException;
 import org.eclipse.search.ui.ISearchQuery;
 import org.eclipse.search.ui.ISearchResult;
 import org.eclipse.xtext.resource.IReferenceDescription;
 
 import com.google.common.base.Predicate;
+import com.google.common.base.Throwables;
 import com.google.inject.Inject;
 
 /**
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=526246

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/ReferenceQuery.java
```diff
@@ -75,7 +76,14 @@ public ISearchResult getSearchResult() {
 	@Override
 	public IStatus run(IProgressMonitor monitor) throws OperationCanceledException {
 		searchResult.reset();
-		finder.findAllReferences(targetURIs, localContextProvider, searchResult, monitor);
+		try {
+			finder.findAllReferences(targetURIs, localContextProvider, searchResult, monitor);
+		} catch (WrappedException e) {
+			if (e.getCause() instanceof OperationCanceledException) {
```
szarnekow: Maybe add `|| monitor.isCancelled()` to the condition and simply throw a new OCE?

cdietrich: thought of that but that would cover other exceptions if cancelling happens after exception and before catch

szarnekow: I'm not sure that the OCE will be wrapped into a single WrappedException all the time. Do we need the call to finish in case of errors, too?

cdietrich: youn mean catch and log the rest? obviously the caller catches all bad exceptions and log them so it looks good to me to rethrow. maybe we can use the cause finder from guava?

szarnekow: :+1:

PR #432, opened 2017-10-22
Enable PluginImageHelper getting image from another plugin.

gallandarakhneorg:

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/PluginImageHelper.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/PluginImageHelper.java
@@ -23,11 +23,30 @@ import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import com.google.inject.name.Named;
 
-/**
+/** Helper for loading images from plugins.
+ *
+ * <p>
+ * This helper enables to find the images within the current plugin and other
+ * plugins. In case the image is located within another plugin than the current
+ * one, the image's name should be qualifed with the plugin's identifier:
+ * {@code platform:/plugin/<plugin_identifier>/<image_name>.<image_filename_extension>}.
+ * The image name must not contains a dot character.
+ * For example, the following name corresponds to the image with the name
+ * {@code "mypicture.png"} within the plugin with the identifier
+ * {@code "org.myorganization.myplugin"}.
+ * <pre><code>platform:/plugin/org.myorganization.myplugin/mypicture.png</code></pre>
+ * </p>
+ *
  * @author Sebastian Zarnekow
+ * @author Stephane Galland - Add function for searching the images within plugins.
  */
 @Singleton
 public class PluginImageHelper implements IImageHelper, IImageDescriptorHelper, SynchronousBundleListener {
+
+	private static final String PLATFORM_URL_PREFIX = "platform:/plugin/";
+
+	private static final String PLATFORM_URL_SEPARATOR = "/";
+
 	private Map<ImageDescriptor, Image> registry = Maps.newHashMapWithExpectedSize(10);
 
 	@Inject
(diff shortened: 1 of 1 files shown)


cdietrich:
I think it’s a bad idea to use jdt code
Maybe looking at this bug will give a better solution 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=404741

kthoms:
Second that. We had recently trouble with JDT internals, we should avoid it. Could you find a way with public API only?

gallandarakhneorg:
Make sense to me. I will update my PR by removing the JDT reference, and taking into consideration the proposal from Christian.

gallandarakhneorg:
Reference to JDT can be removed safely. I used in a previous version it but not any more.
The code within the bug report provided by Christian seems to be the same as the one already inside `PluginImageHelper`, am I right? It does not consider references to images within another plugin as the current plugin. I have updated the code for enabling cross-plugin references.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/PluginImageHelper.java
```diff
@@ -184,4 +193,26 @@ public void bundleChanged(BundleEvent event) {
 		}
 	}
 
+	/** Find the descriptor of the image with the given id.
+	 *
+	 * @param name the identifier of the image. It may be qualified with the plugin's id.
+	 * @return the descriptor.
+	 * @since 2.14
+	 */
+	protected ImageDescriptor findImage(String name) {
```
cdietrich: what is the name you are looking for? how do you encode the image name

cdietrich: pluginname.imagename.extension?

gallandarakhneorg: The supported format is of name is "<pluginid>.<imagename>".

cdietrich: Can you please make this a bit more explicit in the javadoc with an example

szarnekow: I suggest to use this syntax for cross-bundle lookup:
`/bundlename/path/image.name.with.dot.ext`
That is, a leading slash and a bundleName, and slash-delimited path to the resource of that bundle.

gallandarakhneorg: We could also use the standard Eclipse RCP notation: `platform:/plugin/<bundlename>/<path>/<image.name.with.dot.ext>`

What is your preferred one?

kthoms: I‘d prefer platform plugin URIs. 
@szarnekow @cdietrich WDYT?

szarnekow: I don't have a strong preference. Most importantly it has to be unambiguous. Testing for a '/' as the first char and putting `platform:/plugin` in front of it are requiring the complete URI is both ok to me.

gallandarakhneorg: I have use the platform-uri syntax.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/PluginImageHelper.java
```diff
@@ -23,11 +23,30 @@
...
 
-/**
+/** Helper for loading images from plugins.
+ *
+ * <p>
+ * This helper enables to find the images within the current plugin and other
+ * plugins. In case the image is located within another plugin than the current
+ * one, the image's name should be qualifed with the plugin's identifier:
+ * {@code platform:/plugin/<plugin_identifier>/<image_name>.<image_filename_extension>}.
+ * The image name must not contains a dot character.
```
kthoms: typo: must not contain...

gallandarakhneorg: Done.

gallandarakhneorg: done.

kthoms:
Thank you, Stéphane!

Issue #439, opened 2017-11-03
TargetPlatformUtil no longer works with Oxygen/Photon


cdietrich:
TargetPlatformUtil does 
add by adding the (parent) folder of a test project e.g. xtend.ide.tests to the target platform
and having a change behaviour
no longer work
it e.g. add xtend.ide.tests.data in following way
<img width="548" alt="bildschirmfoto 2017-11-02 um 19 58 38" src="https://user-images.githubusercontent.com/238736/32364603-063e2dbe-c076-11e7-9369-c1d6191dd5e6.png">
which makes the tests break

cdietrich:
asked on tycho user for advice: https://dev.eclipse.org/mhonarc/lists/tycho-user/msg07815.html

cdietrich:
Possible solution:

- Differentiate between tycho and non tycho in `TargetPlatformUtil` (both variants)
- Make Tycho and opt in and in the code a non tycho opt out
```
public class TargetPlatformUtil {
	
	public static final Boolean FROM_TYCHO = Boolean.getBoolean("fromTycho");
	
	public static void setTargetPlatform() throws Exception {
		if (!FROM_TYCHO) {
			return;
		}
```

```
<argLine>${memory-settings} ${os-jvm-flags} -DfromTycho=true</argLine>
```

This might work until a source layout plugin (the test plugin itself) is needed

cdietrich:
other ideas welcome

LorenzoBettini:
I think the original code of that class was based on code that I've always added to my projects (in turn, my code was based on a solution provided by Mickael Istria); if you have nested projects (e.g., Maven layout), you need to exclude the current bundle, see e.g. https://github.com/LorenzoBettini/edelta/blob/master/edelta.parent/edelta.ui.tests/src/edelta/ui/tests/utils/PDETargetPlatformUtils.java#L63

My solution is based on the very current test project so it should be adapted to make it general.

The problem is not there if you don't have projects as subdirectories of the parent project.

In any case, I always trigger the manual setting of the target platform ONLY when building with Tycho (see the system property at the beginning of the class), which is passed from the pom.

Hope this helps.

cdietrich:
so 

```
Bundle currentBundle = FrameworkUtil.getBundle(PDETargetPlatformUtils.class);
```

will give the test plugin? or is in your case test plugin and util the same?

LorenzoBettini:
[...]

PR #440, opened 2017-11-03
fixed target platform util to work better on oxygen+ #439

cdietrich:
fixed target platform util to work better on oxygen+ #439


cdietrich:
please review

PR #452, opened 2017-11-13
Xtext Setup Updates

kthoms:
- Exclude Xtend IDEA projects from import
- Enable install of Xtend subproject alone
  => extend target platform
- Add LSP4J to target platform


releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -1342,6 +1270,233 @@
...
+          <repository
+              url="${p2.swtbot}"/>
+          <repository
+              url="${p2.xtext.orbit}"/>
+          <repository
+              url="${p2.emf}"/>
+          <repository
+              url="${p2.draw2d}"/>
+          <repository
+              url="http://download.eclipse.org/staging/oxygen/"/>
```
cdietrich: sicher dass das oxygen 2 enbthält. für mich sieht das nach nur oygen aus (ist auch in meiner target leider noch so drinne)

kthoms: You would have to choose "Oxygen" instead of "Oxygen Staging". The "Staging" platforms only make sense when you really want to use staging instead of release.

cdietrich: No is there a oxygen staging for .2??

szarnekow:
This will always lead to compiler errors when we change code, that is used from idea. I don't like the idea to remove the idea only from the setup. Either we keep it here or we also remove it from the code base.

kthoms:
Nobody is maintaining IDEA code at the moment, and Xtend .idea projects show always compile errors when freshly set up.
IMHO the workspace should be clean after fresh provisioning.

szarnekow:
Hmm so far I don't have compile errors in my workspace with all idea projects imported. Which errors do you see?

kthoms:
A fresh oomphed workspace does not configure the xtend.idea projects properly. 

The IDEA working set is completely broken, dependencies are not properly initialized.

For the moment I tend to remove these and handle these in a separate issue to fix it?

screenshot 170

kthoms:
screenshot 171
screenshot 172

cdietrich:
thats for idea in xtend. these projects are not even part of the build
for xtext-idea itself there should be no problem (at least i did not have any the last time i tried)
the question is what are the kinds of errors you get?

kthoms:
As you see in the first of the 3 screenshots the xtend.idea projects are not properly initialized:
- source folders not properly configured
- no libraries configured

I'm not sure what causes this, but I think we should not solve this right now, raise an issue for that and exclude these projects from importing until it is fixed.

cdietrich:
in xtext idea the src folders are correct.
but i dont see whats wrong with the containers.
is there any errors logged by buildship?

kthoms:
Maybe in your workspace. This is about what a user gets when executing the Oomph setup. No errors in the error log or setup log.
It is just not working and I want the users have a working setup, and if problems are related to IDEA we should drop that for the moment! It is not worth dealing with idea related code until someone works on it.

cdietrich:
@oehme do you have any idea why buildship does not execute downloadIdea after importing the projects or on a gradle fresh project?

kthoms:
Problem is that after project import the IDEA SDK is not downloaded to ~/.gradle/ideaSDK/145.972.3

This should happen automatically. Problem could be solved by triggering manually a build and then do a Gradle Refresh in Eclipse afterwards.

kthoms:
For the IDEA issue I have opened #454

oehme:
@cdietrich Buildship does not support running tasks on import yet. You can follow https://github.com/eclipse/buildship/issues/265

cdietrich:
that was clear so far. but i thought at least a classpath resolution happens

cdietrich:
@kthoms anything left?

PR #456, opened 2017-11-15
Update even more plugin.xmls

brodo:
See #542 in xtext-core


brodo:
Updated now!

kthoms:
Thanks, Julian!

PR #459, opened 2017-11-16
[#457] Restructure target platforms for Tycho build

### LINKED ISSUE #457: [tycho] Put all build target definitions below org.eclipse.xtext.target
At the moment targets are distributed into separate subprojects:
- org.eclipse.xtext.target
- org.eclipse.xtext.oxygen.target
- org.eclipse.xtext.photon.target

This does not scale for more targets. Instead all targets should be put into org.eclipse.xtext.target and the build should pick them all. The used target for the build should be selectable through a profile.

This article is a good reference for this approach: http://blog.vogella.com/2013/01/03/tycho-advanced/

kthoms:
Fixes #457


releng/org.eclipse.xtext.tycho.parent/pom.xml
```diff
@@ -83,18 +78,14 @@
...
-				<groupId>org.eclipse.tycho</groupId>
 				<artifactId>target-platform-configuration</artifactId>
 				<version>${tycho-version}</version>
 				<configuration>
 					<target>
 						<artifact>
 							<groupId>org.eclipse.xtext</groupId>
-							<artifactId>${target-platform-artifactId}</artifactId>
+							<artifactId>org.eclipse.xtext.target</artifactId>
+							<classifier>org.eclipse.xtext.target${target-platform-classifier}</classifier>
```
cdietrich: no separator?

releng/org.eclipse.xtext.target/pom.xml
```diff
@@ -12,4 +12,36 @@
 
 	<packaging>eclipse-target-definition</packaging>
 	<artifactId>org.eclipse.xtext.target</artifactId>
+
+	<build>
```
cdietrich: does this verfy the target files on build?

cdietrich: or does tycho not verfiy at all?

releng/org.eclipse.xtext.tycho.parent/pom.xml
kthoms: Right, otherwise the default case does not work. The separator is part of the classifier.

releng/org.eclipse.xtext.target/pom.xml
kthoms: There is no verification. These are just files. Verification happens only when used as target.

releng/org.eclipse.xtext.target/pom.xml
```diff
@@ -12,4 +12,36 @@
 
 	<packaging>eclipse-target-definition</packaging>
```
LorenzoBettini: @kthoms just in case, you can use "pom" as packaging and have builder-helper-maven plugin only attach the target file based only on the classifier, see, e.g., https://github.com/LorenzoBettini/swtbot-example/blob/master/swtbot.example.targetplatform/pom.xml

kthoms: Good idea. I think we could change this in that direction further.

PR #480, opened 2017-12-07
Saving a resource with auto build disabled keeps the dirty state available until next build. Issue#1151.

baemapo:
Saving a dirty resource won't discard the dirty state information until
a build event is received for the affected resource. This allows to keep
the index up to date regardless of the auto build state.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateManager.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateManager.java
@@ -20,6 +20,7 @@ import org.eclipse.xtext.naming.QualifiedName;
 import org.eclipse.xtext.resource.IEObjectDescription;
 import org.eclipse.xtext.resource.IExternalContentSupport.IExternalContentProvider;
 import org.eclipse.xtext.resource.IResourceDescription;
+import org.eclipse.xtext.resource.IResourceDescription.Delta;
 import org.eclipse.xtext.resource.impl.AbstractResourceDescriptionChangeEventSource;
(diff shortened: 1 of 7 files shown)


szarnekow:
Hi @baemapo,

thank you very much for your PR. There are a few things that we should take into consideration.
Please do not change existing interfaces lightheartedly since that effectively would be API breakage. What we could do though is adding default methods to interfaces.

Another thing: closing an editor without saving should behave differently from closing a saved editor. If you want to adjust your PR, let us know so we can provide a little more guidance.

baemapo:
Hi and thanks for your feedback!

I could move the indexUpdate method to another interface and let the DirtyStateManager implement it to avoid changing IDirtyStateManagerExtension.
I will also check the case of closing the editor without saving...

szarnekow:
Did you find the time to look at it?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateManager.java
```diff
@@ -35,24 +36,30 @@
 public class DirtyStateManager extends AbstractResourceDescriptionChangeEventSource implements IDirtyStateManager, IDirtyStateManagerExtension {
 
 	private ConcurrentMap<URI, IDirtyResource> managedResources;
+	private ConcurrentMap<URI, IDirtyResource> discardCandidates;
```
szarnekow: Introducing an additional field will require us to work with synchronization to keep both states consistent.

cdietrich:
@baemapo any news on this

cdietrich:
ping @baemapo

cdietrich:
will close due missing feedback

Issue #472, opened 2017-12-01
Target Platforms are not properly installed


cdietrich:
please make sure target platforms are propertly installed so that i can call `mvn clean install` on a subproject as well 

proposal: dont use eclipse target platform as packaging but pom and attach the luna platform explicitely

```
<?xml version="1.0"?>
<project
    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
    xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <groupId>org.eclipse.xtext</groupId>
        <artifactId>org.eclipse.xtext.tycho.parent</artifactId>
        <version>2.14.0-SNAPSHOT</version>
        <relativePath>../org.eclipse.xtext.tycho.parent</relativePath>
    </parent>
    <packaging>pom</packaging>
    <artifactId>org.eclipse.xtext.target</artifactId>
    <build>
        <plugins>
            <plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>build-helper-maven-plugin</artifactId>
                <executions>
                    <execution>
                        <id>attach-artifacts</id>
                        <phase>package</phase>
                        <goals>
                            <goal>attach-artifact</goal>
                        </goals>
                        <configuration>
                            <artifacts>
                                <artifact>
                                    <file>${project.artifactId}.target</file>
                                    <type>target</type>
                                    <classifier>org.eclipse.xtext.target</classifier>
                                </artifact>
                                <artifact>
                                    <file>${project.artifactId}-oxygen.target</file>
                                    <type>target</type>
                                    <classifier>oxygen</classifier>
                                </artifact>
                                <artifact>
                                    <file>${project.artifactId}-photon.target</file>
                                    <type>target</type>
                                    <classifier>photon</classifier>
                                </artifact>
                            </artifacts>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
</project>
```

same issue applys to github.com/eclipse/xtext-xtend

cdietrich:
cc @kthoms

cdietrich:
> All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config.

cdietrich:
TODO update to oxygen sr2

[...]

PR #481, opened 2017-12-07
Fixed problem with Target Platforms not being properly installed #472

cdietrich:
Fixed problem with Target Platforms not being properly installed #472


releng/org.eclipse.xtext.target/pom.xml
```diff
@@ -28,6 +28,11 @@
 						<configuration>
 							<artifacts>
 								<artifact>
+                                    <file>${project.artifactId}.target</file>
```
szarnekow: Formatting?

PR #487, opened 2017-12-11
Generalized DontAskAgain

tivervac:
DontAskAgainDialogs allowed you to manage "Do not ask again" settings, but it always did this via a dialog. There are other ways to show this option to the user, thus I made the class more generic.

This also allows for more granular resets of such settings. Before, when you reset the DontAskAgain setting, all of them got reset. Now you can create two classes (e.g. XtendDontAskAgainPreferences and JavaDontAskAgainPrefences.) that both extend DontAskAgainPreferences. These would have two different prefixes, which allows for their settings to be cleared separately for each of them.

Signed-off-by: Titouan <titouan.vervack@sigasi.com>

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/DontAskAgainDialogs.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/DontAskAgainDialogs.java
@@ -13,7 +13,7 @@ import org.eclipse.jface.dialogs.MessageDialogWithToggle;
 import org.eclipse.swt.widgets.Shell;
 
 /**
- * A utility class to ask a user questions, trough a dialog, and remember his decision.<br>
+ * A utility class to ask a user questions, through a dialog, and remember his decision.<br>
  * The settings are stored in {@link IDialogSettings} of your language UIPlugin.
  * 
  * @author Dennis Huebner - Initial contribution and API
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/DontAskAgainDialogs.java
```diff
@@ -12,16 +12,14 @@
 import org.eclipse.jface.dialogs.MessageDialogWithToggle;
 import org.eclipse.swt.widgets.Shell;
 
-import com.google.inject.Inject;
-
 /**
- * A utility class to ask a user questions and remember his decision.<br>
+ * A utility class to ask a user questions, trough a dialog, and remember his decision.<br>
```
cdietrich: through

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/DontAskAgainPreferences.java
```diff
@@ -0,0 +1,104 @@
...
+import org.eclipse.jface.dialogs.MessageDialogWithToggle;
+
+import com.google.inject.Inject;
+
+/**
+ * A utility class to remember a user's decision concerning "Don't ask again" preferences.<br>
+ * The settings are stored in {@link IDialogSettings} of your language UIPlugin.
+ *
+ * @author Titouan Vervack - Initial contribution and API
+ * @since 2.14
```
cdietrich: inconsisten use of @since on class and methods. if i understood correct @since on class is enhough. correct @szarnekow

szarnekow: Yes, that would be sufficient as far as I can tell. It doesn't do any harm, though

tivervac: Fixed it on methods, but left it on the constant because that's how it was in DontAskAgainDialogs.java

tivervac:
@cdietrich ok to go?

cdietrich:
@szarnekow @kthoms can you please have a look

Issue #489, opened 2017-12-14
[oomph] Use Xtext Latest as default for installed Xtext version


kthoms:
The variables page should offer a dropdown box to choose the Xtext version to install:
- Latest Nightly
- Milestone
- Release

kthoms:
New choice on variables page:

![screenshot 29](https://user-images.githubusercontent.com/265597/34108394-20364f96-e400-11e7-81c5-a8edd6a0f435.png)

PR #495, opened 2017-12-18
[#489] Added choice to select Xtext version

kthoms:
This change offers a drop down box to select "Xtext Version" to use for
p2 install and target platform. The choice will influence the p2 repos
to use for Xtext,MWE,Xpand.
Added Xtext SDK to the target platform for the case that only a subset
of subprojects are installed and required bundles are added from the
target platform then.

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -80,24 +80,25 @@
   </setupTask>
   <setupTask
       xsi:type="setup:VariableTask"
-      name="version.mwe2"
-      value="2.9.0"/>
-  <setupTask
-      xsi:type="setup:VariableTask"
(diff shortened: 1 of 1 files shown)


kthoms:
The new choice:

screenshot 29

releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -80,32 +80,33 @@
   </setupTask>
   <setupTask
       xsi:type="setup:VariableTask"
-      name="version.mwe2"
-      value="2.9.0"/>
```
cdietrich: is this no longer needed?

kthoms: No, since we use the composite sites. This includes MWE and Xpand.

Issue #462, opened 2017-11-23
Xtext validation not removing markers in Eclipse?


szschaler:
Hi,

There seems to be some inconsistency in how Xtext handles Eclipse markers (or possibly its own annotations). I had originally asked a question about this in the [forum](https://www.eclipse.org/forums/index.php/m/1776927/). After some discussion, @cdietrich suggested this may actually be a bug, so I am reporting this here.

Specifically, I have an expensive validation that only needs to run if a certain flag is set in my model. I set up the validation as follows (DSML elements faked for the purposes of this post):

```
class MyDSLValidator ....

  @Check(EXPENSIVE)
  def void myExpensiveCheck (Model m) {
    if (m.flag) {
      // perform expensive validation
      error ("Expensive check went wrong", ...)
    }
```

In order to flag up to the user that there may be an error, but we haven't checked this yet, I also add a warning marker to the flag through a FAST check:

```
  @Check
  def void warnAboutMissingCheck (Model m) {
    if (m.flag && !checkMode.shouldCheck(CheckType.EXPENSIVE)) {
      warning("Please run validate for complete checking.", ...)
    }
  }
}
```

As expected, when I create a model in the Xtext editor and enable the flag (by providing a suitable keyword), the warning marker is created and shown in the editor. When I run Validate, I also get the error marker (assuming the model had the problem I'm checking for), but the warning marker is also still there. I have debugged through my code; `warnAboutMissingCheck` is run when I run Validate, but the check works and the warning is not added in that case.

It seems there is no clean Xtext way to remove annotations / markers once they've been created.

@cdietrich had the following thoughts on the [forum](https://www.eclipse.org/forums/index.php/m/1776927/):

> for me this could be a bug .
>org.eclipse.xtext.ui.editor.validation.MarkerIssueProcessor.processIssues(List<Issue>, IProgressMonitor)
>seems to add stuff only but not remove existing ones.
>
>org.eclipse.xtext.ui.editor.validation.AddMarkersOperation.execute(IProgressMonitor) deletes the markers.
>org.eclipse.ui.texteditor.ResourceMarkerAnnotationModel.batchedUpdate(IMarkerDelta[])
>treats thst in the editor.
>
>but thats marker annotations only.
>this leaves the xtext annotations unattended
>
>these are handle by org.eclipse.xtext.ui.editor.validation.AnnotationIssueProcessor.
>
[...]

PR #498, opened 2017-12-18
[#462] Remove annotations not based on resource markers.

ArneDeutsch:
Xtext adds annotations when validating from the editor. These
annotations are set in addition to the ones based on resource markers.
If an expensive validation is triggered it is necessary to remove these
annotations as well to ensure that validations that are NOT triggered by
expensive validation are not shown (of course they will be re-added in
case the validation is triggered e.g. by an edit again).

Change-Id: I0b2708e8a24a54e382f8a6a809dc8883b99c8637

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/AddMarkersOperation.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/AddMarkersOperation.java
@@ -9,7 +9,9 @@
 package org.eclipse.xtext.ui.editor.validation;
 
 import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
 
@@ -18,6 +20,8 @@ import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jface.text.source.Annotation;
+import org.eclipse.jface.text.source.IAnnotationModel;
 import org.eclipse.ui.actions.WorkspaceModifyOperation;
 import org.eclipse.xtext.ui.validation.MarkerTypeProvider;
 import org.eclipse.xtext.validation.Issue;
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/AddMarkersOperation.java
```diff
@@ -38,34 +52,50 @@
...
+	@Deprecated
+	public AddMarkersOperation(IResource resource, List<Issue> issues, Set<String> markerIds, boolean deleteMarkers,
+			MarkerCreator markerCreator, MarkerTypeProvider markerTypeProvider) {
+		this(resource, issues, markerIds, deleteMarkers, null, markerCreator, markerTypeProvider);
+	}
+
+	/**
+	 * @since 2.14
+	 */
+	public AddMarkersOperation(IResource resource, List<Issue> issues, Set<String> markerIds, boolean deleteMarkers,
```
szarnekow: Time to introduce a parameter object?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/AddMarkersOperation.java
```diff
@@ -108,4 +138,20 @@ protected void execute(final IProgressMonitor monitor) throws CoreException, Inv
 		}
 	}
 
+	private void deleteAnnotations() {
+		if (annotationModel != null) {
+			@SuppressWarnings("unchecked")
+			Iterator<Annotation> iterator = annotationModel.getAnnotationIterator();
+			List<Annotation> toRemove = new ArrayList<>();
+			while (iterator.hasNext()) {
+				Annotation next = iterator.next();
```
szarnekow: We should only remove annotations that are not yet marked as deleted

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/ValidatingEditorCallback.java
```diff
@@ -54,9 +54,11 @@ public void afterSave(XtextEditor editor) {
 	private ValidationJob newValidationJob(XtextEditor editor) {
 		IValidationIssueProcessor issueProcessor;
 		if (editor.getResource() == null) {
-			issueProcessor = new AnnotationIssueProcessor(editor.getDocument(), editor.getInternalSourceViewer().getAnnotationModel(), issueResolutionProvider);
+			issueProcessor = new AnnotationIssueProcessor(editor.getDocument(), editor.getInternalSourceViewer().getAnnotationModel(),
```
szarnekow: Should we expose getAnnotationModel on the editor?

ArneDeutsch:
I can not imagine a case where this change influences grey annotation markers. Before this change all the resource markers where already removed by the validation job. In result all annotations created by these markers where removed as well. But the annotations created by the editor only, in the case of the bug report a warning created only when the expensive validations where NOT executed, have remained in the editor. They where painted in yellew/red because the editor handles them as "created as annotation but with no backing marker". Removing these annotations seems to me just consequent because the validation job will recreate all required markers and annotations anyway.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/ValidatingEditorCallback.java
ArneDeutsch: Well, this exposure is not new but is already used in AnnotationIssueProcessor. Probably it would be more nice to not expose the model but that would be quite a lot of refactoring work.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/AddMarkersOperation.java
ArneDeutsch: Good point, I will add " && !next.isMarkedDeleted()"

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/AddMarkersOperation.java
ArneDeutsch: I just experimented a bit. Of course it is possible to introduce some parameter object or a builder. But because of the backwards compatibility (the old constructors) the telescoping constructor remains and in result the code gets just more complicated. If it is ok with you I would prefer to leave it as it is.

cdietrich:
@szarnekow @kthoms please look if ok and if yes merge

kthoms:
Thanks, Arne!

Issue #51, opened 2016-08-15
[quickfix] Add quickfix for create action


kthoms:
A rule 
   `Foo: 'id' (name=ID)?;Foo: 'id' (name=ID)?;`
produces the warning 
   `The rule 'Foo' may be consumed without object instantiation.
   Add an action to ensure object creation, e.g. '{Foo}'.`

Adding this action could be provided by a quickfix.

meysholdt:
What would be the proposed strategy to decide where and what actions to insert? Cann all cases be handled? 

Example:

  Rule:
      name=ID? | {Other} 'other';

kthoms:
Your example is a good example for that not all cases can be handled correctly. But is this the intention of a quickfix? It is just a proposal for what you could do to correct the situation. It is not said that is the perfect solution, this is true also for other quick fixes. And as the warning already proposes is that you could add a create action for the rule that you just created.

Many rules that I have seen place the create action at the beginning, so that is at least a valid fix. If you have to decide more decent where the create action is to be placed and what kind of object should be created, you would not want to use the quickfix anyway. But newbies might appreciate such a fix that does what the warning message proposes.

I think the quickfix could be labeled "Insert create action `'{Foo}'` on rule entry.".

meysholdt:
I think a quickfix should only be available for scenarios that it can fix reliably. A quickfix that turns a warning into an error makes the tool look broken.

And yes, there are scenarios where a simple quickfix can work:

`R1: 'X'?;`  --> `R1: {R1} 'X'?;`
`R2: name=ID?;`  --> `R2: {R2} 'name=ID'?;`

For others, "place the create action at the beginning", leads to errors:

`R3: ("true" | {False} "false")` --> `R3: {R3} ("true" | {False} "false")`

The latter is erroneous because current EObject may only be instantiated once.

So a prerequisite for this quickfix implementation would be that it does not offer itself for scenarios it can't solve.

kthoms:
Agreed.

ArneDeutsch:
Added implementation that can handle all cases together with test cases.

kthoms:
Please describe the feature also in the release notes:
https://github.com/eclipse/xtext/blob/website-master/xtext-website/_posts/releasenotes/version-2-14-0.md

szarnekow:
Im asking to reopen this bug and address the issues from the review. Also Moritz mentioned a few things in this discussion.

ArneDeutsch:
[...]

PR #503, opened 2017-12-20
[#51] Add quick fix for create warning.

ArneDeutsch:
Quick fix reuses logic for detection of the warning and hence is able to
handle all cases.
Add test cases for many cases.

Change-Id: I82667455d23cbeceddef7725fa83f1892c6d9d32

### DIFF
--- a/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
+++ b/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
@@ -42,9 +42,13 @@ import org.eclipse.emf.ecore.xmi.impl.URIHandlerImpl;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.text.edits.InsertEdit;
+import org.eclipse.text.edits.MultiTextEdit;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.actions.WorkspaceModifyOperation;
+import org.eclipse.xtext.AbstractElement;
 import org.eclipse.xtext.AbstractRule;
+import org.eclipse.xtext.Alternatives;
 import org.eclipse.xtext.EcoreUtil2;
 import org.eclipse.xtext.EnumLiteralDeclaration;
 import org.eclipse.xtext.Keyword;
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
```diff
@@ -474,4 +479,40 @@ public Integer exec(XtextResource state) throws Exception {
...
+
+	@Fix(RuleWithoutInstantiationInspector.ISSUE_CODE)
+	public void addAction(final Issue issue, IssueResolutionAcceptor acceptor) {
+		acceptor.accept(issue, "Add actions to ensure object creation", "Inserts the missing actions to ensure object creation.", null,
+				new ISemanticModification() {
+					@Override
+					public void apply(EObject element, IModificationContext context) throws Exception {
+						ParserRule rule = (ParserRule) element;
+						MultiTextEdit textEdit = new MultiTextEdit();
+						applyToNode(rule.getAlternatives(), "{" + rule.getName() + "} ", textEdit);
```
szarnekow: The action must produce an instance of the type that is returned by the rule. It may need to be qualified with the metamodels alias. The code as is is likely to produce invalid grammars. Another thing is, that the return type may be an abstract type.

ArneDeutsch: I have introduceed code to handle "returns" even for static models. Writing tests for the cases where the model is not generated will be quite a lot of effort. There are no tests yet at all for static models which means to setup (and probably refactor) testing infrastructure.

Would you be able to guide me to examples about cases where the rule needs to be qualified or the returned type is abstract? I was not able to come to such examples by my own. Maybe I miss something important here.

ArneDeutsch: My latest comit: https://github.com/ArneDeutsch/xtext-eclipse/commit/a25f740de01b2aea7416b77c9e2cdb14389495ab

szarnekow: A good starting point would be to use a `generate` clause with an alias, e.g. `generate myDsl "http://www.xtext.org/example/mydsl/MyDsl" as alias` and `Model returns alias::Model: ...` respectively.

ArneDeutsch: Added code and test to handle alias as well.

Issue #501, opened 2017-12-20
[examples] Add build metadata


kthoms:
The example projects do not contain build files. The examples should be extended by that metadata. It would be better to restructure the projects so that each of them is moved into a sub directory. The packaging scripts have to be adjusted.

cdietrich:
ae we sure everybody has maven or gradle?

kthoms:
Does this really matter? The examples all have Eclipse UI, thus the projects have to be build with Maven anyway. If they have not Maven, the examples have just appropriate examples with them and do not harm. Otherwise you would be glad that you have a reference already in the examples.

cdietrich:
i wonder hat happens if projects have maven nature and stuff but no maven installed

kthoms:
They won't have a Maven nature, that's not necessary. They will just have pom.xml files and additional projects for parent and target.

cdietrich:
@kthoms what is the state of this issue?

kthoms:
Status: Done for Domainmodel, open for the others.

cdietrich:
since we have support for build in domainmodel.
do we really want to clutter other samples with that as well?

cdietrich:
cc @kthoms @szarnekow

kthoms:
I'm fine with having just Domainmodel so. I'll close this one. If someone objects please reopen.

PR #507, opened 2017-12-21
[#501] Add build for Domainmodel example

kthoms:
Adds build definition for the Domainmodel example.

Test:
1. run `mvn package` on project org.eclipse.xtext.xtext.ui.examples. This will bundle the domainmodel projects
2. In runtime Eclipse, create the Domainmodel example with the wizard
3. Execute `buildDomainmodel.launch` in the new releng project.

screenshot 37
 

- add project 'org.eclipse.xtext.example.domainmodel.releng'
  - contains parent pom.xml
  - contains submodule 'targetplatform' defining the build target
    (avoiding yet another project)
- register new project to build.xml and plugin.xml
- added pom.xml files for all domainmodel projects
- added aggregator pom.xml in folder 'projects'


org.eclipse.xtext.xtext.ui.examples/build.xml
```diff
@@ -60,6 +60,7 @@
 		<zip_example subdir="domainmodel"    name="org.eclipse.xtext.example.domainmodel.tests" />
 		<zip_example subdir="domainmodel"    name="org.eclipse.xtext.example.domainmodel.ui" />
 		<zip_example subdir="domainmodel"    name="org.eclipse.xtext.example.domainmodel.ui.tests" />
+		<zip_example subdir="domainmodel"    name="org.eclipse.xtext.example.domainmodel.releng" />
```
cdietrich: usual naming convention is .parent

kthoms: it is more than just the parent, it contains also the target platform. I don't like to have too much top-level projects which contain only releng related stuff. Also for repository and feature project.

kthoms: I know it differs from the structure we usually produce with the wizard. But the EMF example wizard can only install separate projects. Better would be to Zip the whole example as one archive and place the parent POM in the root folder.
I had to create a separate project for the releng stuff, thus decided to call it "releng".

cdietrich: dont get you. id like to have the name a.b.c.parent as we do it at different places and which is  a common pattern

kthoms: The parent project created with the wizard is physically the root folder of the other projects, which is common Maven style.
We can't do this with the wizard, we have to produce a sibling project that contains the parent pom.xml. Of course, we could name that "parent".

That's the project structure created by the wizard:
screenshot 40

That's the proposed Domainmodel example project structure:
screenshot 41

If we would want to use the typical structure, we can't go with the EMF example project anymore. Or we just install the "root" project and have to import the nested projects afterwards.

cdietrich: ok go for releng

PR #509, opened 2017-12-28
529269 add icon to xtext build console

Bananeweizen:
Signed-off-by: Michael Keppler <Michael.Keppler@gmx.de>


szarnekow:
I'm not an expert on this, but do we need to provide a HDPI icon?

Bananeweizen:
Good catch. I must admit I don't really know myself, but I can add an @2 icon, to be on the safe side.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/debug/XtextBuildConsole.xtend
```diff
@@ -24,7 +28,10 @@ class XtextBuildConsole extends IOConsole {
 	val PrintStream out
 
 	new() {
-		super("Xtext Build", "xtextBuildConsole", null, true)
+		super("Xtext Build", "xtextBuildConsole",
+			ImageDescriptor.createFromURL(
+				FileLocator.find(Platform.getBundle("org.eclipse.xtext.builder"), new Path("icons/console.png"), null)),
```
kthoms: `Activator.^default.bundle.getEntry("icons/console.png")` should do the job also

Bananeweizen:
Thanks for your comment. I will try to remember avoiding that lookup for other situations. That's what happens if you just look at the FAQ for getting a bundle based image. :)

kthoms:
Thank you, Michael!

PR #510, opened 2018-01-03
[#51] Handle case where rule contains a "returns"

ArneDeutsch:
Also for static model case.

Add test case for rule with "returns" expression.

Change-Id: I3bdc260e9e95e117277b55ee781014faf5aec579


org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/quickfix/XtextGrammarQuickfixProvider.java
```diff
@@ -487,8 +487,11 @@ public void addAction(final Issue issue, IssueResolutionAcceptor acceptor) {
 					@Override
 					public void apply(EObject element, IModificationContext context) throws Exception {
 						ParserRule rule = (ParserRule) element;
+						String actionName = rule.getName();
+						if (rule.getType().getClassifier() != null)
+							actionName = rule.getType().getClassifier().getName();
```
szarnekow: Please also include the metamodel alias if it is not empty.

Issue #347, opened 2017-09-04
Highlighting is done twice


cdietrich:
Having a XtextEditor opened, Semantic highlighting is done twice.
once via `org.eclipse.xtext.ui.editor.syntaxcoloring.HighlightingReconciler.refresh()`  called via `org.eclipse.xtext.ui.editor.XtextEditor.installHighlightingHelper()` and once via `org.eclipse.xtext.ui.editor.XtextEditor.forceReconcile()` 

@spoenemann do you remember why you introduced the second one being called from `org.eclipse.xtext.ui.editor.XtextEditor.createPartControl(Composite)`

spoenemann:
See https://github.com/eclipse/xtext-eclipse/commit/d4fe755d6dd211ce0ca0d498d3f9e5913d4659e6:

> added forceReconcile() to XtextEditor.createPartControl(..) to avoid conflicts during initialization of the outline view

cdietrich:
yes but this tells me and sebastian zero. )=> what kind are these conflicts of

spoenemann:
Don't remember...

cdietrich:
cc @svenefftinge any thoughts?

svenefftinge:
Sorry, no idea.

ArneDeutsch:
Experimented with highlighting while open/close outline view, setting breakpoints etc. Can not find any side effect when removing the "forceReconcile()" call. I guess the original bug 454227 was fixed by the other changes inside https://github.com/eclipse/xtext-eclipse/commit/d4fe755d6dd211ce0ca0d498d3f9e5913d4659e6 and the added "forceReconcile()" was a mistake. After removing the call no duplicate highlighting happens any more. This should improve performance for large files or files with expensive semantic highlighting quite a lot.

dhuebner:
@ArneDeutsch  After switching to 2.14 I'm facing an issue with missing Semantic Highlighting in a custom DSL Editor and synchronized custom View. What is the strategy for users to easily fix the misbehavior on the synchronized View side (similar to outline view)

dhuebner:
I will answer my question. The first thing to do is, to scan the code for `document.priorityReadOnly` calls. 😑

dhuebner:
@ArneDeutsch The problem was that the initial semantic highlighting stopped working after updating to 2.14. The reason was a `priorityReadOnly` call that prevented the semantic highlighter to finish the job. 
[...]

PR #513, opened 2018-01-08
[#347] Avoid double semantic highlighting by removing bad call.

ArneDeutsch:
Experimented with highlighting while open/close outline view, setting
breakpoints etc. Can not find any side effect when removing the
"forceReconcile()" call. I guess the original bug 454227 was fixed by
the other changes inside
https://github.com/eclipse/xtext-eclipse/commit/d4fe755d6dd211ce0ca0d498d3f9e5913d4659e6
and the added "forceReconcile()" was a mistake. After removing the call
no duplicate highlighting happens any more. This should improve
performance for large files or files with expensive semantic
highlighting quite a lot.

Change-Id: I3a543dfa1f0e7b1e690bf40fbf1db766c73217b5


cdietrich:
as discussed "offline" please have a look at the failing quickfix tests

PR #533, opened 2018-01-24
Recompiled Xtend sources

kthoms:
Sources changed after fresh Oomph setup against Luna target on Oxygen.

I have the same changes in my Photon environment.

### DIFF
--- a/org.eclipse.xtext.xbase.ui/xtend-gen/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
+++ b/org.eclipse.xtext.xbase.ui/xtend-gen/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
@@ -47,7 +47,7 @@ public class XbaseEditorInputRedirector {
   public ITypeRoot getTypeRoot(final IEditorInput it) {
     ITypeRoot _xblockexpression = null;
     {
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.xbase.ui/xtend-gen/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -47,7 +47,7 @@
   public ITypeRoot getTypeRoot(final IEditorInput it) {
     ITypeRoot _xblockexpression = null;
     {
-      final Object adapter = it.getAdapter(IJavaElement.class);
```
cdietrich: are you sure that works with luna tp?

kthoms: hmm, will double check

kthoms: I thought I had a Luna TP enabled, but now set up a new devenv with Luna TP. Actually the generics are not present on Luna, makes this one invalid. I'll close it.

Issue #40, opened 2016-08-04
Performance bottleneck in QueuedBuildData when building huge number of resources


kittaakos:
I [ran into a performance issue] when I was trying to build more than 500k very simple and small resources in my workspace. My workspace contains 256 projects, each project contains 256 folders and the resources are balanced into these folders. Although I managed to perform a full clean build, I experienced some performance loss in the <tt>QueuedBuildData#queueURI(URI)</tt> method.

Although the collections in the <tt>QueuedBuildData</tt> should not contain duplicates it is currently implemented with <tt>LinkedList</tt> to comply the <tt>Queue</tt> API. To avoid duplicate resource URIs, <tt>contains</tt> is being called on these <tt>LinkedList</tt>s which is pretty expensive in my case. Actually, that is the 20% of my overall build process.

I would recommend to implement some sort of factory and inject it into the <tt>QueuedBuildData</tt>. This factory will be responsible for creating the collection instances for <tt>uris</tt> and the values of the <tt>projectNameToChangedResource</tt> field as well. By default, a <tt>LinkedList</tt> instances will be created. But, via injection, one could inject a factory that creates <tt>LinkedHashSet</tt> instances. Since the iteration order of the <tt>LinkedHashSet</tt> equals to the insertion order, it can be wrapped into an <tt>AbstractQueue</tt> and can act as a queue.

I believe this neither a bug nor a critical feature request, but it could be useful for others as well.

By the way, I am happy to contribute a patch.

[ran into a performance issue]: https://www.eclipse.org/forums/index.php/t/1079817/

kittaakos:
I have the feeling what I have "spammed" above previously is false, or at least just partially true.

[...]

PR #534, opened 2018-01-25
[#40] use set based queue instead of linked list

cdietrich:
[#40] use set based queue instead of linked list

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
@@ -1,0 +1,85 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.builder.impl;
+
+import java.util.AbstractQueue;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Queue;
+
+/**
+ * {@link Queue} with unique valuess
+ * 
+ * @noextend This class is not intended to be subclassed by clients.
+ * @noinstantiate This class is not intended to be instantiated by clients.
+ * @author Christian Dietrich - Initial contribution and API
+ */
+public class LinkedHashSetQueue<T> extends AbstractQueue<T> {
+	private final LinkedHashSet<T> data;
+
+	public LinkedHashSetQueue(LinkedHashSetQueue<T> base) {
+		this.data = new LinkedHashSet<>(base.data);
+	}
+
+	public LinkedHashSetQueue() {
+		this.data = new LinkedHashSet<>();
+	}
+
+	@Override
+	public boolean offer(T e) {
+		data.add(e);
+		return true;
+	}
+
+	@Override
+	public T poll() {
+		Iterator<T> iterator = data.iterator();
+		if (iterator.hasNext()) {
+			T next = iterator.next();
+			iterator.remove();
+			return next;
+		}
+		return null;
+	}
+
+	@Override
+	public T peek() {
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Queue;
+
+/**
+ * {@link Queue} with unique valuess
+ * 
+ * @author Christian Dietrich - Initial contribution and API
+ */
+public class LinkedHashSetQueue<T> extends AbstractQueue<T> {
```
szarnekow: Since this impl is not general purpose, I'd like to make this package private.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
```diff
@@ -0,0 +1,87 @@
...
+	private final LinkedHashSet<T> data;
+
+	public LinkedHashSetQueue(LinkedHashSet<T> newData) {
+		this.data = newData;
```
szarnekow: A general purpose impl would probably create a defensive copy of newData

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	public LinkedHashSetQueue() {
+		this.data = new LinkedHashSet<>();
+	}
+
+	@Override
+	public boolean offer(T e) {
+		data.add(e);
+		return true;
```
szarnekow: ... and return false here

cdietrich: this wont work. will lead to queue full

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
cdietrich: agreeded

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
cdietrich: agreed

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
cdietrich: but our test structure disallows this

szarnekow: Understood.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
szarnekow: My bad :)

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/LinkedHashSetQueue.java
cdietrich: Will add our nietend no Instantiate javadoc

kittaakos:
I have just briefly glanced into this PR, so I might be wrong, but I think you have just missed the whole point, because `contains` is not overridden, so it will delegate into the `AbstractCollection`'s contains, which uses the iterator approach. Did I miss something?

Ignore me, if everything is good as is.

cdietrich:
@kittaakos you are right but contains should no longer be called but ill change this

cdietrich:
added contains and containsAll

kittaakos:
You are right, I have missed [it](https://github.com/eclipse/xtext-eclipse/pull/534/files#diff-ba0cb32b61f530593b2c1999fb7750f4L291). Thanks for the clarification!

PR #537, opened 2018-01-25
Prevent NPE

tivervac:


tivervac:
Note that this might also be a problem in other places.

cdietrich:
thx @tivervac

PR #544, opened 2018-02-01
Fixed NPE when opening an uri with empty fragment

tivervac:
Happens when you try to open an URI with an empty fragment i.e. a file.


cdietrich:
i wonder what the guard if `(uri.fragment() != null) {` was meant for ....

tivervac:
The problem is that that guard just cancels the open request. We could however move the null guard to the new guard

cdietrich:
Will have a look @szarnekow what do you think?

szarnekow:
The change looks good to me. The empty fragment points to the root instance of the resource and that could be at a different offset than zero when the file starts with a comment.

PR #545, opened 2018-02-01
Fixed NPE in SyncUtil

tivervac:
I Have no way of reproducing this but there are plenty of nullchecks on ValidationJob in other places.


cdietrich:
thx @tivervac

PR #547, opened 2018-02-02
Kth issue546

kthoms:


org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/targetplatform/org.eclipse.xtext.example.domainmodel.targetplatform.target
```diff
@@ -1,25 +0,0 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
cdietrich: why this rename?

kthoms: To make it shorter

Issue #552, opened 2018-02-06
Reduce usage of internal API


kthoms:
We have to avoid internal API usage where possible. This is an umbrella issue for refactorings in xtext-eclipse for Xtext 2.14.

cdietrich:
see https://github.com/eclipse/xtext/issues/2393

kthoms:
Closing this one, DUP of eclipse/xtext#2393

PR #553, opened 2018-02-06
[#552] Resolve usage of internal class Region

kthoms:
This will remove the dependency on `org.eclipse.jdt.internal.core.Region`.


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/smap/DebugSourceInstallingCompilationParticipant.java
```diff
@@ -185,8 +185,26 @@ public void buildStarting(BuildContext[] files, boolean isBatch) {
 	}
 
 	protected List<IFile> findGeneratedJavaClassFiles(IJavaElement element) {
-		Region region = new Region();
-		region.add(element);
+		IRegion region = new IRegion() {
```
cdietrich: why not javacore.newRegion?

kthoms: just was not aware of that. will refactor it.

kthoms: updated

Issue #551, opened 2018-02-05
ContentFormatter performance very slow for lange documents


bdznkr:
`org.eclipse.xtext.ui.editor.formatting2.ContentFormatter.format()` needs 15-16 Minutes for a document with 10.000 lines. The initial formatting creates 50.000 `TextEdit` Objects.
The most time gets lost in the `DocumentPartitioner`.

With this workaround:
```
	public void format(IDocument document, IRegion region) {
		IXtextDocument doc = (IXtextDocument) document;

		TextEdit r = doc.priorityReadOnly(new FormattingUnitOfWork(doc, region));
		if (r != null) {

			Map<String, IDocumentPartitioner> partitioners =  TextUtilities.removeDocumentPartitioners(document);

			try {
				r.apply(document);

			} catch (BadLocationException e) {
				throw new RuntimeException(e);
			}  

			TextUtilities.addDocumentPartitioners(document, partitioners);
 		}
	}
```
execution time was reduced to 45-60 seconds.

now most time is lost in:

`XtextDocument`:

```
	protected void fireDocumentChanged(DocumentEvent event) {
		....
		tokenSource.updateStructure(event);
		....
	}
```

cdietrich:
could you please provide a complete testcase/example?

cdietrich:
jdt handles this here:

```
				if (partitioner instanceof IDocumentPartitionerExtension3) {
					IDocumentPartitionerExtension3 extension= (IDocumentPartitionerExtension3) partitioner;
					if (extension.getActiveRewriteSession() != null)
						continue;
				}
```

=> we should have a look at the concept on an Active Rewrite Session

bdznkr:
If you format the attached .xtend file you will have the similar performace problem,
but be careful it might hang your eclipse for several minutes.
Maybe, try 2000 or 4000 lines first.

[FormatterPerformanceTest.zip](https://github.com/eclipse/xtext-eclipse/files/1695663/FormatterPerformanceTest.zip)

cdietrich:
```
DocumentPartitioner.stopRewriteSession(DocumentRewriteSession) line: 900	
XtextDocument(AbstractDocument).checkStateOfPartitioner(IDocumentPartitioner, String) line: 1574	
XtextDocument(AbstractDocument).computePartitioning(String, int, int, boolean) line: 1320	
[...]

PR #554, opened 2018-02-06
[#551] Notify on Real Changes only

cdietrich:
[#551] Notify on Real Changes only

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocument.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocument.java
@@ -528,11 +528,13 @@ public class XtextDocument extends Document implements IXtextDocument {
 					throw e.getWrapped();
 				} finally {
 					try {
-						if(potentialUpdaterCount.decrementAndGet() == 0 && (hadUpdates || isCancelReaders)) { 
+						if(potentialUpdaterCount.decrementAndGet() == 0 && (hadUpdates || isCancelReaders)) {
+							boolean wasHadUpdates = hadUpdates;
 							hadUpdates = false;
 							if (getCancelIndicator().isCanceled() && isCancelable)
 								throw new OperationCanceledException();
-							notifyModelListeners(state);
+							if (wasHadUpdates)
+								notifyModelListeners(state);
 						}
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocument.java
```diff
@@ -528,11 +528,13 @@ protected XtextResource getState() {
...
 					try {
-						if(potentialUpdaterCount.decrementAndGet() == 0 && (hadUpdates || isCancelReaders)) { 
+						if(potentialUpdaterCount.decrementAndGet() == 0 && (hadUpdates || isCancelReaders)) {
+							boolean wasHadUpdates = hadUpdates;
 							hadUpdates = false;
 							if (getCancelIndicator().isCanceled() && isCancelable)
 								throw new OperationCanceledException();
-							notifyModelListeners(state);
+							if (wasHadUpdates)
+								notifyModelListeners(state);
```
kthoms: code style: shall we encapsulate single-line statements in blocks?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocument.java
```diff
@@ -528,11 +528,13 @@ protected XtextResource getState() {
 					throw e.getWrapped();
 				} finally {
 					try {
-						if(potentialUpdaterCount.decrementAndGet() == 0 && (hadUpdates || isCancelReaders)) { 
+						if(potentialUpdaterCount.decrementAndGet() == 0 && (hadUpdates || isCancelReaders)) {
+							boolean wasHadUpdates = hadUpdates;
```
kthoms: strange variable name.
how about setting hadUpdates in a finally-block to false after notifying the listeners? I did not see that it is mandatory to set it to false before.

```
try {
   ...
   if (hasUpdates) { notifyModelListeners(state); }
} finally {
   hadUpdates = false;
}
```

cdietrich: i just kept the existing styles ;)

cdietrich:
ping @szarnekow

PR #574, opened 2018-02-18
get rid of reflective code.

cdietrich:
https://github.com/eclipse/xtext-core/issues/506

### DIFF
--- a/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
+++ b/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
@@ -7,7 +7,6 @@
  *******************************************************************************/
 package org.eclipse.xtext.common.types.xtext.ui;
 
-import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
@@ -24,7 +23,6 @@ import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IType;
 import org.eclipse.jdt.core.JavaModelException;
-import org.eclipse.jdt.core.WorkingCopyOwner;
 import org.eclipse.jdt.core.compiler.IProblem;
 import org.eclipse.jdt.core.search.IJavaSearchConstants;
 import org.eclipse.jdt.core.search.IJavaSearchScope;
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtTypesProposalProvider.java
```diff
@@ -243,13 +241,9 @@ protected IJavaSearchScope createSearchScope(IJavaProject project, JvmType super
...
 		try {
-			// Method is available in JDT 3.6 or better
-			// more than twice as fast as the older alternative
-			Method method = SearchEngine.class.getMethod("createStrictHierarchyScope", IJavaProject.class, IType.class, Boolean.TYPE, Boolean.TYPE, WorkingCopyOwner.class);
-			method.setAccessible(true);
-			IJavaSearchScope result = (IJavaSearchScope) method.invoke(null, project, type, Boolean.TRUE, Boolean.TRUE, null);
+			IJavaSearchScope result = SearchEngine.createStrictHierarchyScope(project, type, Boolean.TRUE, Boolean.TRUE, null);
 			return result;
-		} catch (Exception e) {
+		} catch (JavaModelException e) {
```
szarnekow: I think most of the stuff in the catch block was for older jdt versions. Shouldn't this be removed?

cdietrich: i dont see how this is connected to reflection.

szarnekow: The caught exception was for the missing method. Now we catch JavaModelException, but that's also declared on the method signature and likely to be thrown again by the code in the catch block.

cdietrich: i am not sure. the old could fail with java model exception or reflective exception

szarnekow: True. Point taken.

cdietrich: and JavaModelException extends Exception

cdietrich: so you would simply remove the catch?

szarnekow: Yes, that's what I would have done. But it could have been the wrong thing to do in this case. So  you convinced me that the fallback could still be valuable.

cdietrich: ;+1:

PR #593, opened 2018-03-14
[Refactoring] Save editors when persisted files are affected

meysholdt:
Editors should remain unsaved after a refactoring only when no persited
files are affected. Otherwise, the builder will find an inconsistent
state since it can't see the dirty state.


meysholdt:
thank you for the review. The test "works on my machine"... I mean, Jenkins: http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/me%252Frefactoring/1/testReport/org.eclipse.xtext.ui.tests.editor.model/DocumentLockerTest/testPriorityReadOnlyCancelsReaders/

PR #621, opened 2018-03-29
Warnings

tivervac:


org.eclipse.xtext.xbase.ui/.classpath
```diff
@@ -4,6 +4,13 @@
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins">
 		<accessrules>
 			<accessrule kind="accessible" pattern="org/eclipse/jdt/internal/ui/actions/ActionMessages**"/>
+			<accessrule kind="accessible" pattern="org/eclipse/jdt/internal/core/DefaultWorkingCopyOwner"/>
```
cdietrich: am not sure if i want this. this hides internal usage warning. correct?

org.eclipse.xtext.common.types.eclipse.tests/tests/org/eclipse/xtext/common/types/access/xtext/ui/JdtBasedConstructorScopeTest.java
```diff
@@ -73,7 +73,7 @@ public boolean apply(IEObjectDescription input) {
 		assertFalse(Iterables.any(contents, new Predicate<IEObjectDescription>() {
```
cdietrich: should be assertTrue i thing. @szarnekow what do you think?

org.eclipse.xtext.common.types.eclipse.tests/tests/org/eclipse/xtext/common/types/access/xtext/ui/JdtBasedConstructorScopeTest.java
```diff
@@ -85,7 +85,7 @@ public boolean apply(IEObjectDescription input) {
...
```
cdietrich: should be assertTrue i thing. @szarnekow what do you think?

org.eclipse.xtext.common.types.eclipse.tests/tests/org/eclipse/xtext/common/types/access/xtext/ui/JdtBasedConstructorScopeTest.java
cdietrich: but i am not sure here, cause default constructor is protected

org.eclipse.xtext.xbase.ui/.classpath
tivervac: Correct, right now only warnings for ActionMessages** were ignored. I can understand not wanting to hide those. However, the warnings are only useful if people actually look at them, not sure if that is the case.

cdietrich: we plan
(1) to reduce internal api
(2) become xfriends with the eclipse internal api we use so that we get notified on changes.
this is why i want to keep the warnings for now

tivervac: Sounds like a good idea to me, I'll revert this change.

org.eclipse.xtext.common.types.eclipse.tests/tests/org/eclipse/xtext/common/types/access/xtext/ui/JdtBasedConstructorScopeTest.java
tivervac: Having debugged the test I can say that AbstractCollection and AbstractList are present, I think this should be assertTrue as well.

org.eclipse.xtext.common.types.eclipse.tests/tests/org/eclipse/xtext/common/types/access/xtext/ui/JdtBasedConstructorScopeTest.java
tivervac: Having debugged the test I can say that AbstractList is present, I think this should be assertTrue as well.

szarnekow:
I think the tests were broken before the fix since a QualifiedName was never equality a String

cdietrich:
But what is the expected behaviour for the constructor scope?

cdietrich:
@tivervac can you please change both tests so that they pass?

szarnekow:
> But what is the expected behaviour for the constructor scope?

The scope has to contain the constructors of the abstract types, too. The link resolution should be successful, but languages would have the chance to validate the resolved references as broken because the instantiated types are abstract.

PR #622, opened 2018-03-29
Added I[Read|Write]Access and Priority to XtextDocumentLocker

tivervac:
This is based on feedback of #617
It also needs https://github.com/eclipse/xtext-core/pull/682


cdietrich:
thx @tivervac

PR #633, opened 2018-04-09
[eclipse/xtext-core#686] Update new project wizard.

ArneDeutsch:
Fix case where unsupported JDK is added to the selection combo in case
it is selected as a default. Instead, in the unlikely case that the
default JDK is not supported, the last JDK from the list is selected.

Remove IDEA from the new project wizard. IDE plugins are not well
maintained any more, no need to bother new users with not working stuff.

Adding "Automatic-Module-Name" to manifest if selected java version is
at least 1.8.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/PluginProjectFactory.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/PluginProjectFactory.java
@@ -48,7 +48,7 @@ public class PluginProjectFactory extends JavaProjectFactory {
 		this.getRequiredBundles().addAll(requiredBundles);
 		return this;
 	}
-	
+
 	/**
 	 * @since 2.14
 	 */
@@ -62,7 +62,7 @@ public class PluginProjectFactory extends JavaProjectFactory {
 		this.getExportedPackages().addAll(exportedPackages);
 		return this;
 	}
-	
+
 	/**
 	 * @since 2.14
 	 */
@@ -76,7 +76,7 @@ public class PluginProjectFactory extends JavaProjectFactory {
 		this.getImportedPackages().addAll(importedPackages);
 		return this;
 	}
-	
+
 	/**
 	 * @since 2.14
 	 */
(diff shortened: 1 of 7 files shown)


cdietrich:
Automatic-Module-Name is a java 9 thingy?!?

kthoms:
It is recommended to add this header to all manifests. It will be ignored for older platforms, but is relevant to Java modules. It is not specifically Java 9, but J9 and higher.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/PluginProjectFactory.java
```diff
@@ -130,6 +131,9 @@ public String apply(String input) {
 
 	protected void createManifest(IProject project, IProgressMonitor progressMonitor) throws CoreException {
 		final StringBuilder content = new StringBuilder("Manifest-Version: 1.0\n");
+		if (JavaVersion.fromBree(getBreeToUse()).isAtLeast(JavaVersion.JAVA8)) {
```
kthoms: Why not add always?

cdietrich:
id like to have it for java9 and higher only

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/PluginProjectFactory.java
cdietrich: cause it might confuse users

kthoms:
What if plugins are created for BREE 1.8, but runs on J9+ ? Then this header would be required.

cdietrich:
@szarnekow what do you think?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/PluginProjectFactory.java
szarnekow: If we want to have a guard, it should be at least Java9 I suppose. But I’m with @kthoms here - guess it’s safe to have it for older versions, too. As long as it does not produce a warning.

cdietrich:
ok.

to be left: how to deal with java 5,6,7
(a) remove
(b) add warning

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/PluginProjectFactory.java
ArneDeutsch: I removed the condition. It started with java 9 but the fact that it is ignored for java 8 have let me to the wrong conclusion that it was a java 8 thing.

ArneDeutsch:
Removed all JDKs < 8 (the same way it was done up to now < 6).

ArneDeutsch:
Now: allow all from java 6 but warn if it is not at least java 8.

PR #634, opened 2018-04-10
[Bug 407831] Handle open declaration in external file editor.

ArneDeutsch:
Detect that a file is external to the workspace a create a specialized
IStorage. Then detect this specialized IStorage and create the same
IEditorInput that is created by eclipse when opening external files.
This way open declaration works in external editors and editor instances
are reused.


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/utils/EditorUtils.java
```diff
@@ -82,7 +84,7 @@ public static XtextEditor getActiveXtextEditor(ExecutionEvent event) {
 		IEditorPart activeEditor = HandlerUtil.getActiveEditor(event);
 		if (activeEditor == null)
 			return null;
-		XtextEditor xtextEditor = (XtextEditor) activeEditor.getAdapter(XtextEditor.class);
+		XtextEditor xtextEditor = activeEditor.getAdapter(XtextEditor.class);
```
szarnekow: Is the missing cast Luna compatible?

cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/arne-bug-407831/

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/FileStoreStorage.java
```diff
@@ -0,0 +1,80 @@
...
+		return fileInfo.getAttribute(EFS.ATTRIBUTE_READ_ONLY);
+	}
+
+	public IFileStore getFileStore() {
+		return fileStore;
+	}
+
+	@Override
+	public String getCharset() throws CoreException {
+		return null; // encoding unknown, use default encoding
```
szarnekow: Why do we implement IEncodedStorage if we do nothing about the encoding?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/FileStoreStorage.java
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	@Override
+	public boolean equals(Object object) {
+		return getFullPath().equals(object);
```
szarnekow: Equals is not symmetric

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/utils/EditorUtils.java
ArneDeutsch: Will fix for Luna.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/FileStoreStorage.java
ArneDeutsch: The code using the instance makes a "instanceof" check. If the class does not implement "IEncodedStorage" it falls back to ask the workspace for the encoding. Leading to an exception in the end. Implementing the interface leading to a null. In turn the "default" encoing from eclipse is used what is the bast we can do at this place (as the encoding of external files is unknown).

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/FileStoreStorage.java
ArneDeutsch: Will fix.

ArneDeutsch:
The failing tests seem to be unrelated to these commits. Both tests succeed for me locally + I can see no logical connection to my copy changes.

cdietrich:
tests are green here: http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/arne-bug-407831/

cdietrich:
@szarnekow @ArneDeutsch how is the review state of this pr

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/FileStoreStorage.java
szarnekow: Ok

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperImpl.java
```diff
@@ -216,9 +219,22 @@ protected boolean isHandled(IFolder folder) {
...
+			Iterable<Pair<IStorage, IProject>> result = contribution.getStorages(uri);
+			if (result == null || !result.iterator().hasNext()) {
+				 // Handle files external to the workspace. But check contribution first to be backwards compatible.
+				if (uri.isFile()) {
+					Path filePath = new Path(uri.toFileString());
+					IFileStore fileStore = EFS.getLocalFileSystem().getStore(filePath);
+					IFileInfo fileInfo = fileStore.fetchInfo();
+					if (fileInfo.exists() && !fileInfo.isDirectory()) {
+						return Collections.singletonList(
+								Tuples.<IStorage, IProject> create(new FileStoreStorage(fileStore, fileInfo, filePath), (IProject) null));
```
szarnekow: Should we also override internalGetURI for FileStoreStorages?

szarnekow: There is also https://github.com/eclipse/n4js/blob/master/plugins/org.eclipse.n4js.ts.ui/src/org/eclipse/n4js/ts/ui/navigation/URIBasedStorage.java

Would it be possible to use a similar impl? It appears to be slightly simpler, e.g. no EFS and stuff

szarnekow: Ping

ArneDeutsch: I adapted "internalGetUri"

ArneDeutsch:
I have adapted to all the requested changes. Guess this can be merged to fix the bug.

szarnekow:
Some tests for this feature would be great. Apart from that, it looks good to me.

cdietrich:
@ArneDeutsch @szarnekow what is the state of this pr?

ArneDeutsch:
I have finished the implementation but have not yet invested the time to investigate how I can add UI tests for this.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/IResourcesSetupUtil.java
```diff
@@ -246,6 +248,15 @@ protected void execute(IProgressMonitor monitor)
 		return file;
 	}
 
+	public static File createTempFile(String fileName, String suffix, String content)
+			throws Exception {
+		File file = File.createTempFile(fileName, suffix);
+		FileWriter writer = new FileWriter(file);
```
szarnekow: It's always surpising if a closable is not closed in a finally block

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractEditorTest.java
```diff
@@ -8,53 +8,60 @@
...
 public abstract class AbstractEditorTest extends AbstractWorkbenchTest {
 
 	static final long STEP_DELAY = 0;
-	
+
 	protected abstract String getEditorId();
 
 	protected XtextEditor openEditor(IFile file) throws Exception {
-		IEditorPart openEditor = openEditor(file, getEditorId());
+		return getXtextEditor(
```
szarnekow: Nice. Can we add an openEditor with a URI, too? That would allow us to test with files in archives.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/IResourcesSetupUtil.java
ArneDeutsch: Fixed.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractEditorTest.java
ArneDeutsch: Probably possible, althought I'm not sure if it would work in the end with more or less every URI or just for the archives case. But I would say in scope of this ticket this is not needed.

szarnekow: In general we can open a stream for all kinds of URIs so a URI based input should do the trick. In any case you are right - this is not in the scope of this task.

PR #641, opened 2018-04-18
use asm 6.1.1 in oomph

cdietrich:

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -99,7 +99,7 @@
   <setupTask
       xsi:type="setup:VariableTask"
       name="p2.orbit"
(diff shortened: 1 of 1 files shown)


releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -767,10 +767,10 @@
             name="org.eclipse.xtext.sdk.feature.group"/>
         <requirement
             name="org.objectweb.asm"
-            versionRange="[6.0.0,7.0.0)"/>
+            versionRange="[6.1.1,7.0.0)"/>
```
holgerschill: I would prefer to work with a fixed version.

cdietrich: i dont want to do this now. can you create an issue at eclipse/xtext so we can discuss this in general and fix it between the beta and release

PR #650, opened 2018-04-25
Improved filtering of annotations for hover

### LINKED ISSUE #649: Hover does not show appropriate validation annotation
I found two general problems with hover over errors, warnings etc.

 * When multiple annotations overlap each other, the hover box shows an arbitrary one, even if it has lower priority than other annotations at the same position in the document. For example, when a text region is marked with both an error and a warning, the hover sometimes shows only the warning.
 * No hover is shown for annotations that span multiple lines unless you hover over the first line.

spoenemann:
Fixes #649.

 * `AbstractProblemHover.getAnnotations(int, int)` now returns all annotations that include the given offset (if it is ≥ 0). Those that match the given line number are put first in the list so consumers of that method will favor them over annotations that start on preceding lines.
 * `ProblemAnnotationHover` and `AnnotationWithQuickFixesHover` further sort the list of annotations such that they first consider errors, then warnings, then the rest.

### DIFF
--- a/org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/hover/AbstractProblemHoverTest.java
+++ b/org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/hover/AbstractProblemHoverTest.java
@@ -13,8 +13,10 @@ package org.eclipse.xtext.ui.tests.editor.hover;
 import java.util.List;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IMarker;
 import org.eclipse.jface.text.IRegion;
 import org.eclipse.jface.text.ITextViewer;
+import org.eclipse.jface.text.source.Annotation;
 import org.eclipse.ui.IWorkbenchPart;
 import org.eclipse.xtext.resource.XtextResource;
 import org.eclipse.xtext.ui.MarkerTypes;
@@ -34,6 +36,8 @@ import org.junit.Test;
  * @author Christoph Kulla - Initial contribution and API
  */
 public class AbstractProblemHoverTest extends AbstractEditorTest {
+	
+	protected IFile file;
 
 	protected XtextEditor editor;
 	
@@ -47,7 +51,7 @@ public class AbstractProblemHoverTest extends AbstractEditorTest {
 	public void setUp() throws Exception {
 		super.setUp();
 		modelAsText = "stuff mystuff\nstuff yourstuff refs _mystuff// Comment";
-		IFile file = IResourcesSetupUtil.createFile("test/test.testlanguage", modelAsText);
+		file = IResourcesSetupUtil.createFile("test/test.testlanguage", modelAsText);
 		editor = openEditor(file);
 		document = editor.getDocument();
 		hover = TestsActivator.getInstance().getInjector(getEditorId()).getInstance(MockHover.class);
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
```diff
@@ -67,38 +68,81 @@ protected IAnnotationModel getAnnotationModel() {
...
+				boolean isError2 = markerAnnotationAccess.isSubtype(a2.getType(), "org.eclipse.ui.workbench.texteditor.error");
+				if (isError1 && isError2)
+					return 0;
+				if (isError1)
+					return -1;
+				if (isError2)
+					return 1;
+				boolean isWarning1 = markerAnnotationAccess.isSubtype(a1.getType(), "org.eclipse.ui.workbench.texteditor.warning");
+				boolean isWarning2 = markerAnnotationAccess.isSubtype(a2.getType(), "org.eclipse.ui.workbench.texteditor.warning");
+				if (isWarning1 && isWarning2)
```
szarnekow: We should probably sort the annotations by message as the last argument to get a stable order.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
```diff
@@ -67,38 +68,81 @@ protected IAnnotationModel getAnnotationModel() {
...
+	 * Find all handled annotations that include the given offset. If there are multiple annotations,
+	 * those that start on the given line number are put first in the list, then those that start one
+	 * line before, etc.
+	 */
 	public List<Annotation> getAnnotations(final int lineNumber, final int offset) {
 		if (getAnnotationModel() == null) {
 			return Collections.emptyList();
 		}
-		final Iterator<?> iterator = getAnnotationModel().getAnnotationIterator();
+		final Iterator<Annotation> iterator = getAnnotationModel().getAnnotationIterator();
```
cdietrich: are you sure this in luna compatible?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
```diff
@@ -67,38 +68,81 @@ protected IAnnotationModel getAnnotationModel() {
...
+	protected List<Annotation> sortBySeverity(List<Annotation> annotations) {
+		if (annotations == null)
+			return Collections.emptyList();
+		if (annotations.size() <= 1)
+			return annotations;
+		List<Annotation> sorted = Lists.newArrayList(annotations);
+		Collections.sort(sorted, new Comparator<Annotation>() {
+			@Override
+			public int compare(Annotation a1, Annotation a2) {
+				boolean isError1 = markerAnnotationAccess.isSubtype(a1.getType(), "org.eclipse.ui.workbench.texteditor.error");
```
szarnekow: Can we extract two methods `isError`, `isWarning` probably also `isInfo` and use these is `isHandled`, too? This would avoid the duplicated string constants.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
spoenemann: Do we need to be source compatible? This should not change anything on the bytecode level.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
szarnekow: Using Boolean.compare(..) could further simplify the code, e.g. Comparator.comparing(this::isError, Boolean::compare).thenComparing(this::isWarning, ..)

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
cdietrich: it will give an error to those wo edit with luna as tp.
we have planned to move to oxygen as minimal tp after the release so we would change this then (along with 200 other places)

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
```diff
@@ -67,38 +68,81 @@ protected IAnnotationModel getAnnotationModel() {
 		return sourceViewer.getAnnotationModel();
 	}
 
+	/**
+	 * Find all handled annotations that include the given offset. If there are multiple annotations,
+	 * those that start on the given line number are put first in the list, then those that start one
+	 * line before, etc.
+	 */
 	public List<Annotation> getAnnotations(final int lineNumber, final int offset) {
```
szarnekow: Please add `Does not return lineDiffInfos.` in contrast to `isHandled`

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
spoenemann: Ok, I'll revert that line.

cdietrich: i will have a look if luna already has the generic there

szarnekow: This is not Luna compatible

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
```diff
@@ -67,38 +68,81 @@ protected IAnnotationModel getAnnotationModel() {
...
+							if (lineDiff == 0 || offset >= 0) {
+								// Insert the annotation according to its line number difference
+								int index = lineDiffs.size();
+								while (index > 0 && lineDiffs.get(index - 1) > lineDiff) {
+									index--;
+								}
+								result.add(index, annotation);
+								lineDiffs.add(index, lineDiff);
+							}
+						} catch (Exception x) {
```
szarnekow: What sort of exceptions are ignored here?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
szarnekow: No it's not.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
spoenemann: Probably BadLocationException from `getDocument().getLineOfOffset(start)`. I kept this catch clause as it was before in order not to break anything.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
spoenemann: The sorting algorithm is stable, so it retains the order that was returned by `getAnnotations(int, int)` in case the annotations have the same type. That order is important since it puts those annotations first that start on the same line as the requested position in the document.

Previously we have just taken the order that was given by the annotation model. I haven't noticed any instability there.

spoenemann:
I added a test for each of the two reported problems.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
szarnekow: Good point, I wasn't sure about the default sorting of lists being stable.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hover/AbstractProblemHover.java
szarnekow: Ok

org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/editor/hover/AbstractProblemHoverTest.java
```diff
@@ -98,6 +155,11 @@ protected IRegion getHoverRegionInternal(int lineNumber, int offset) {
 		protected Object getHoverInfoInternal(ITextViewer textViewer, int lineNumber, int offset) {
 			return null;
 		}
+		
+		@Override
```
cdietrich: is this still needed?

cdietrich: you could add a comment that it makes the getter visible to the test and thus is needed

kthoms:
@cdietrich @szarnekow Fine for you?

meysholdt:
thx @spoenemann & reviewers!

Issue #654, opened 2018-05-01
Provide test infrastructure for testing customized DoubleClickStrategyProvider


miklossy:
Some Xtext-based languages ([Eclipse Xtend](https://github.com/eclipse/xtext-xtend/blob/master/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/editor/XtendDoubleClickStrategyProvider.java), [Eclipse GEF DOT](https://bugs.eclipse.org/bugs/show_bug.cgi?id=532244)) require a customized DoubleClickStrategyProvider. 

See also https://www.eclipse.org/forums/index.php/t/373244/.

It would be useful to extend the Xtext test infrastructure to enable a comfortable testing of such customized DoubleClickStrategyProvider behaviour.

cdietrich:
Fixed in 2.14

PR #655, opened 2018-05-01
[#654] Add AbstractEditorDoubleClickTextSelectionTest abstract class.

miklossy:
AbstractEditorDoubleClickTextSelectionTest abstract class to provide a
base infrastructure for testing a customized DoubleClickStrategyProvider
behaviour.

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractEditorDoubleClickTextSelectionTest.xtend
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractEditorDoubleClickTextSelectionTest.xtend
@@ -1,0 +1,103 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ui.testing
+
+import com.google.inject.Inject
+import org.eclipse.jface.text.ITextSelection
+import org.eclipse.swt.SWT
+import org.eclipse.swt.widgets.Event
+import org.eclipse.ui.texteditor.AbstractTextEditor
+import org.eclipse.xtext.resource.FileExtensionProvider
+import org.eclipse.xtext.ui.editor.XtextEditor
+import org.eclipse.xtext.ui.testing.util.IResourcesSetupUtil
+import org.eclipse.xtext.ui.editor.XtextEditorInfo
+
+/**
+ * @since 2.14
+ */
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractEditorDoubleClickTextSelectionTest.xtend
```diff
@@ -0,0 +1,99 @@
...
+	protected def getCursorPosition(CharSequence text) {
+		val cursorPosition = text.toString.indexOf(c)
+		if(cursorPosition == -1) {
+			fail('''
+				The input text
+				�text�
+				must contain the �c� symbols indicating the current cursor position!
+			''')
+		}
+		cursorPosition
```
szarnekow: It's a matter of taste, but explicit returns read better in code blocks.

miklossy: I personally prefer implicit return and use it wherever as possible. What do @cdietrich and @kthoms think?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractEditorDoubleClickTextSelectionTest.xtend
```diff
@@ -0,0 +1,99 @@
...
+	 * @param it
+	 *            The editor's input text. The text must contain the {@link #c}
+	 *            special symbols indicating the current cursor position.
+	 * 
+	 * @param expected
+	 *            The text that is expected to be selected after double clicking
+	 *            in the Xtext editor on the current cursor position.
+	 */
+	def assertSelectedTextAfterDoubleClicking(CharSequence it, String expected) {
+
```
cdietrich: formatting looks wired

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractEditorDoubleClickTextSelectionTest.xtend
cdietrich: if i use implict returns in combination with a non single line statement i specifiy a explicit return type

miklossy:
I updated the commit in the PR and implemented the following improvements:

- Use injection of the XtextEditorInfo object to access the editor id.
- Specify an explicit return type in method signatures using implict returns in combination with a non single line statement.
- Change the default special symbol from '<|>' to '|' to improve the readability of the tests.
- Improve the javadoc formatting.

Thank you both for your feedback!

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractEditorDoubleClickTextSelectionTest.xtend
holgerschill: I prefer explicit returns always since it reads better.

Issue #656, opened 2018-05-02
Reset prompt dialog settings has no margin


Bananeweizen:
The preference to reset the "Don't ask again" dialogs explicitly sets zero pixels margin. That leads to bad layout: 
![image](https://user-images.githubusercontent.com/406876/39523412-ca310ea8-4e15-11e8-9a9c-fd8efcdcdc2f.png)

Relevant is probably https://github.com/eclipse/xtext-eclipse/blob/581a5c827c4c32860d3bfb2fcc046c6fdd80f578/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/preferences/ResetPromptDialogSettingsField.java#L69 and the following line.

cdietrich:
Fixed in 2.14

PR #659, opened 2018-05-04
[#656] Fixed Margin Problem in Reset Preference Page

cdietrich:
[#656] Fixed Margin Problem in Reset Preference Page

Old:
bildschirmfoto vom 2018-05-04 07-53-03

New:
bildschirmfoto vom 2018-05-04 07-53-38


kthoms:
On MacOS there is no difference. The dialog looks fine already without patch. But I think the change does not harm.

Before:
screenshot 64

After:
screenshot 65

PR #671, opened 2018-05-14
Typo in method name. Added getParallelFileSystemAccess as replacement.

kthoms:

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/ParallelBuilderParticipant.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/ParallelBuilderParticipant.java
@@ -304,7 +304,7 @@ public class ParallelBuilderParticipant extends BuilderParticipant {
 					Set<IFile> derivedResources = getDerivedResources(delta, buildContext.outputConfigurations, buildContext.generatorMarkers);
 
 					FileSystemAccessQueue fileSystemAccessQueue = buildContext.fileSystemAccessQueue;
-					IFileSystemAccess2 fsa = getParalleFileSystemAccess(delta, buildContext, derivedResources, fileSystemAccessQueue, buildContext.synchronousFileSystemAccess);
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/ParallelBuilderParticipant.java
```diff
@@ -382,9 +382,19 @@ public void run() {
 
 	/**
 	 * @since 2.9
+	 * @deprecated Use {@link #getParallelFileSystemAccess(Delta, org.eclipse.xtext.builder.IXtextBuilderParticipant.IBuildContext, Set, FileSystemAccessQueue, IFileSystemAccess2)}
 	 */
+	@Deprecated
 	protected IFileSystemAccess2 getParalleFileSystemAccess(final IResourceDescription.Delta delta,
 			final IBuildContext context, Set<IFile> derivedResources, FileSystemAccessQueue fileSystemAccessQueue, IFileSystemAccess2 delegate) {
+		return getParallelFileSystemAccess(delta, context, derivedResources, fileSystemAccessQueue, delegate);
```
cdietrich: * @noextend This class is not intended to be subclassed by clients.
 * @noinstantiate This class is not intended to be instantiated by clients.

but has `@since`
i am not sure how to deal with that api wise @szarnekow what do you think

cdietrich: and i dislike this solution cause its value is zero right now

kthoms: an error is an error. not important for 2.14 though.

szarnekow: We do not need the bridge method. The since tag is on this class to document that it was introduced in that version. The noextend and noinstantiate do have a higher prio.

PR #673, opened 2018-05-16
[eclipse/xtext-core#666] Configure JUnit library container in test proj.

kthoms:
This will add the JUnit library containerto test projects with the selected JUnit version

### DIFF
--- a/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
+++ b/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
@@ -38,11 +38,13 @@ import org.eclipse.xtext.ui.util.PluginProjectFactory;
 import org.eclipse.xtext.ui.util.ProjectFactory;
 import org.eclipse.xtext.ui.wizard.IProjectCreator;
 import org.eclipse.xtext.ui.wizard.IProjectInfo;
+import org.eclipse.xtext.util.JUnitVersion;
 import org.eclipse.xtext.xtext.wizard.AbstractFile;
 import org.eclipse.xtext.xtext.wizard.BinaryFile;
 import org.eclipse.xtext.xtext.wizard.ParentProjectDescriptor;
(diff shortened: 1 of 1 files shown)


kthoms:
Test failure seems unrelated

kthoms:
JU4
screenshot 101

JU5
screenshot 103

org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
```diff
@@ -125,6 +127,11 @@ private void configureJavaProject(ProjectDescriptor descriptor, JavaProjectFacto
 		if (needsBuildshipIntegration(descriptor) && !descriptor.isEclipsePluginProject()) {
 			factory.addClasspathEntries(JavaCore.newContainerEntry(new Path("org.eclipse.buildship.core.gradleclasspathcontainer")));
 		}
+		if (descriptor instanceof TestProjectDescriptor) {
+			JUnitVersion junitVersion = descriptor.getConfig().getJunitVersion();
```
cdietrich: Do we need this in grade and pure maven too

kthoms: Yes. The projects only have a dependency on the JU API, not on the test engines. To execute the tests in Eclipse, we still need the JU container.

kthoms: It might be necessary to add the engine dependencies for the surefire plugin, but the project itself should not depend on them for similar reason as in Tycho.

cdietrich: Ok should be handle this with a different pr / the enhance Junit 5 support with 2.15 issue

kthoms: At least with different PR. If possible for 2.14, I'll try that.

PR #679, opened 2018-05-17
[eclipse/xtext-core#666] Configure JUnit lib container for plain Maven

kthoms:
When using the Maven source layout the JUnit library container must also
be configured for the runtime project.


kthoms:
Before:
screenshot 113

After:
screenshot 115

Issue #681, opened 2018-05-22
DocumentLockerTest / FSSynchronizationTest are flaky


cdietrich:
```
Fehlermeldung

sleep interrupted

Stacktrace

java.lang.InterruptedException: sleep interrupted
	at java.lang.Thread.sleep(Native Method)
	at org.eclipse.xtext.ui.tests.editor.model.DocumentLockerTest.testPriorityReadOnlyCancelsReaders(DocumentLockerTest.java:151)

```

```
java.lang.NullPointerException: null
	at org.eclipse.xtext.builder.impl.AbstractFSSynchronizationTest.testCleanUpDerivedResourcesWithCreateBetween(AbstractFSSynchronizationTest.java:341)
	at org.eclipse.xtext.builder.impl.AbstractFSSynchronizationTest.testCleanUpDerivedResourcesWithCreateBetween(AbstractFSSynchronizationTest.java:255)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
......
```

cdietrich:
the slower the (virtual) hardware the more often this seems to happen.
i propose to ignore the tests for now and investigate what happens

ArneDeutsch:
Here is the stack trace that currently leads to the (sometimes) failing test:

```
java.lang.InterruptedException: null
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1302)
	at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:231)
	at org.eclipse.xtext.ui.tests.editor.model.DocumentLockerTest.testPriorityReadOnlyCancelsReaders(DocumentLockerTest.java:151)
```

It looks as if the test runner thread is interrupted. I can't say why. But it might be that we are bitten by a JUnit bug here: https://github.com/junit-team/junit4/issues/1365

I will insert a `Thread.currentThread.interrupted` at the start of the test case to ensure the interrupted flag is cleared.

ArneDeutsch:
[...]

PR #683, opened 2018-05-24
[#681] Fix race condition.

ArneDeutsch:
There is no guarantee from the VM that a write to an array from one
thread is visible to any other thread at any time. This will happen
especially when there is a high load with high memory pressure. In this
case I guess the started thread passes the "check.set" but this value
does not be shown to the main thread. That waits sometimes for 40sec in
the while loop until the first threads timed out.

Replaced the code with a CountDownLatch that does not suffer from this
issue. Flakyness should be gone.


cdietrich:
thx @ArneDeutsch

PR #687, opened 2018-05-25
Made printing of source code on error overridable

tivervac:
Printing source code is not always something that is acceptable thus there should be a way to disable this behaviour. Extracting the message creation to an overridable function achieves this purpose.


cdietrich:
thx @tivervac

Issue #665, opened 2018-05-08
The isPlatformResource guard in ParallelResourceLoader.loadResource does not work for resources in jars


cdietrich:
`ParallelResourceLoader.loadResource` has a guard that does not load out of sync workspace resources to avoid a deadlock between the loading thread (that would need a workspace lock to do a refresh) and the builder thread (that actually holds the workspace lock)

currently this guard simply does a check for `uri.isPlatformResource()`. but this wont match for a uri like `archive:platform:/resource/Program/extLib/SmartFlow.jar!/org/eclipse/xtext/Xtext.ecore`

so we run right into that deadlock

```
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	at java.lang.Object.wait(Object.java:502)
	at org.eclipse.core.internal.jobs.ThreadJob.waitForRun(ThreadJob.java:313)
	- locked <0x00000004c222ed50> (a java.lang.Object)
	at org.eclipse.core.internal.jobs.ThreadJob.joinRun(ThreadJob.java:202)
	at org.eclipse.core.internal.jobs.ImplicitJobs.begin(ImplicitJobs.java:92)
	at org.eclipse.core.internal.jobs.JobManager.beginRule(JobManager.java:308)
	at org.eclipse.core.internal.resources.WorkManager.checkIn(WorkManager.java:121)
	at org.eclipse.core.internal.resources.Workspace.prepareOperation(Workspace.java:2188)
	at org.eclipse.core.internal.resources.Resource.refreshLocal(Resource.java:1552)
	at org.eclipse.core.internal.resources.File.refreshLocal(File.java:316)
	at org.eclipse.emf.ecore.resource.impl.PlatformResourceURIHandlerImpl$WorkbenchHelper.createPlatformResourceInputStream(PlatformResourceURIHandlerImpl.java:205)
	at org.eclipse.emf.ecore.resource.impl.PlatformResourceURIHandlerImpl.createInputStream(PlatformResourceURIHandlerImpl.java:525)
	at org.eclipse.emf.ecore.resource.impl.ExtensibleURIConverterImpl.createInputStream(ExtensibleURIConverterImpl.java:360)
	at org.eclipse.xtext.resource.XtextResourceSet$1.createInputStream(XtextResourceSet.java:298)
[...]

PR #691, opened 2018-05-27
[#665] Fixed Issue with Parallel Loaders and Out-Of-Sync jar files in workspace

cdietrich:
[#665] Fixed Issue with Parallel Loaders and Out-Of-Sync jar files in workspace

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
@@ -10,6 +10,7 @@
  *******************************************************************************/
 package org.eclipse.xtext.builder.resourceloader;
 
+import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
 import java.util.NoSuchElementException;
@@ -29,6 +30,7 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.emf.common.archive.ArchiveURLConnection;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.common.util.WrappedException;
 import org.eclipse.emf.ecore.resource.Resource;
@@ -37,6 +39,7 @@ import org.eclipse.xtext.ui.resource.IResourceSetProvider;
 import org.eclipse.xtext.util.Triple;
 import org.eclipse.xtext.util.Tuples;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 
 /**
@@ -87,6 +90,38 @@ public class ParallelResourceLoader extends AbstractResourceLoader {
 						// we will recover & load the resource in the main builder thread (that owns the project resource lock)
 						// (if we do try to load resources that are out of sync, we create a deadlock)
 						return null;
+					}
+				} else if (uri.isArchive()) {
+					String authority = uri.authority();
+					// we dont treat recursive/nested jars here cause Xtext does not support them.
+					if (authority != null && authority.startsWith("platform:/resource/")) {
+						class MyArchiveURLConnection extends ArchiveURLConnection {
+							public MyArchiveURLConnection(String url) {
+								super(url);
+							}
+							public String getNestedURI() {
+								try {
+									return getNestedURL();
+								} catch (IOException exception) {
+									return ""; //$NON-NLS-1$
+								}
+							}
+						}
+						MyArchiveURLConnection archiveURLConnection = new MyArchiveURLConnection(uri.toString());
+						String nestedURIString = archiveURLConnection.getNestedURI();
+						if (Strings.isNullOrEmpty(nestedURIString)) {
+							return null;
+						}
+						URI nestedURI = URI.createURI(nestedURIString);
+						if (nestedURI.isPlatformResource()) {
+							IFile file = root.getFile(new Path(nestedURI.toPlatformString(true)));
+							if (!file.isSynchronized(IResource.DEPTH_ZERO)) {
+								// don't bother trying to refresh it, this loading-thread doesn't own the project resource lock
+								// we will recover & load the resource in the main builder thread (that owns the project resource lock)
+								// (if we do try to load resources that are out of sync, we create a deadlock)
+								return null;
+							}
+						}
 					}
 				}
 			}


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
```diff
@@ -88,6 +91,37 @@ protected Resource loadResource(URI uri, ResourceSet localResourceSet, ResourceS
...
+							}
+							public String getNestedURI() {
+								try {
+									return getNestedURL();
+								} catch (IOException exception) {
+									return ""; //$NON-NLS-1$
+								}
+							}
+						}
+						MyArchiveURLConnection archiveURLConnection = new MyArchiveURLConnection(uri.toString());
```
kthoms: Don't like the 'My' prefix. How about an anonymous class here?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
```diff
@@ -88,6 +91,37 @@ protected Resource loadResource(URI uri, ResourceSet localResourceSet, ResourceS
 						// (if we do try to load resources that are out of sync, we create a deadlock)
 						return null;
 					}
+				} else if (uri.isArchive()) {
+					String authority = uri.authority();
+					if (authority != null && authority.startsWith("platform:/resource/")) {
```
kthoms: any way to get `platform:/resource` replaced by constants or methods on URI?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
cdietrich: Same pattern is used a few lines below or above

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
cdietrich: Please give me that constant if you have one

cdietrich:
ping @szarnekow

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
szarnekow: I'd think this should be a recursive to be able to handle nesting. Also we do have similar code in the Storage2URIMapper 

```
if (result.isEmpty() && uri.isArchive()) {
  String authority = uri.authority();
  authority = authority.substring(0, authority.length() - 1);
  URI archiveURI = URI.createURI(authority);
  if (archiveURI.isFile() || archiveURI.isPlatformResource()) {
```

Should we harmonize both locations?

szarnekow: The constants are protected in URI so not generally reachable.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
szarnekow: The anonymous class would not necessarily improve the readability since we need to access the nested URI, e.g. it would become something like
```
String nestedURIString = new ArchiveURLConnection(url) {
  public String getNestedURI() {..}
}.getNestedURI();
```
I'm not fond of this.

szarnekow: But if we strip the exclamation mark from the authority, it should be safe to remove the url connection stuff completely.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
cdietrich: In my Tests recursive does not work at all in Xtext

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
cdietrich: ?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
cdietrich: And I did not find any exampl,e how recursive uri would look like

cdietrich: => i would need help for that since no idea how to test

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
szarnekow: I understood that the ArchiveURLConnection is only necessary to obtain the nestedURI. It would be sufficient to remove the trailing '!' from the authority and use URI.createURI on the remainder. With that at hand, you could start a new attempt to try isInSync. That would also make nested archives work.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
szarnekow: `jar:jar:platform:/resource/project/path/to/file.jar!/within/jar/nested.jar!/my/dsl.file` should be such a URI for nested jars in jars.

cdietrich: I am not sure about that. And have no idea on how to test that. So I think I leave it was won't fix and we keep the workaround in the project forever

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
cdietrich: and i am not sure how to trim prefixes and suffixes so that uri.createuri wont explode

cdietrich: and the api in emf that tells you how a archive uri starts and how no sucks even more that the one to test for platform resource uri

szarnekow: :)
Agreed.

Issue #645, opened 2018-04-22
[wizard] Detect JUnit version and preselect highest


kthoms:
The wizard should preselect JUnit 5 (only) when it is available on the target platform.

ArneDeutsch:
Done.

PR #694, opened 2018-05-28
[#645] Make selection of JUnit Version dependent on presence of plugin.

ArneDeutsch:

### DIFF
--- a/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.xtend
+++ b/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.xtend
@@ -30,6 +30,7 @@ import static org.eclipse.xtext.xtext.ui.wizard.project.Messages.*
 import static org.osgi.framework.Bundle.*
 import org.eclipse.xtext.util.JUnitVersion
 import org.eclipse.swt.widgets.Label
+import org.eclipse.pde.core.plugin.PluginRegistry
 
 class AdvancedNewProjectPage extends WizardPage {
 
@@ -100,11 +101,12 @@ class AdvancedNewProjectPage extends WizardPage {
 					junitVersion4 = Radio [
 						layoutData = new GridData(SWT.LEFT, SWT.CENTER, false, false)
 						text = "4"
-						selection = true
+						selection = !isJUnit5PluginAvailable
 					]
 					junitVersion5 = Radio [
 						layoutData = new GridData(SWT.LEFT, SWT.CENTER, false, false)
 						text = "5"
+						selection = isJUnit5PluginAvailable
 					]
 				]
 			]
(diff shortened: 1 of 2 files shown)


cdietrich:
i am still not convinced. 

- what about maven and gradle only secenarios
- what about target platforms

kthoms:
This will again just make the availability dependent on the installation of JU5 in the available plugins, but does not consider the target platform. JU5 is for test execution only available when it is part of the target platform.

For plain Gradle & Maven setups we can assume that JU5 is available when the JU5 bundles are installed (this is what the check does), and they will be executable through Gradle/Maven then.

So the case that is not clearly supported is when a project with Eclipse setup is configured (so our default).

ArneDeutsch:
Currently I do not follow your concerns with the propsed solution My point of view:
1. The ticket is about the default selection in the Xtext project creation wizard. It is just about the point if the "JUnit 4" or "JUnit 5" is selected by default. The user is free to select the other if the default does not fit him.
1. At the project creation time there is probably no target platform set. Chances are good that the user will use the created target platform or no targte platform at all.
1. The decision which checkbox is selected by default can be influenced then by used eclipse olny. Of course the user might wish to setup a project completly disconnected from what he has in his eclipse installation. He is free to do so. I would call it an "informaed guess" to look at the eclipse installation if JUnit 5 is present to decide if the user might wish to use JUnit 5.

kthoms:
I can follow your argumentation and agree. I would like to hear a 2nd opinion.

ArneDeutsch:
@szarnekow Maybe you can have a look and psot your opinion?

cdietrich:
i'd still pefer to keep junit 4 as default but i am kind a "old fashioned" in that regard so i may not count

szarnekow:
I've a hard time imagining what a user thinks after the first compile when there are errors. Is there a chance to show at least a warning if Junit5 was selected, there is already a TP in Eclipse configured and it does not contain Junit5?

kthoms:
Same is true for Xtext itself. When no TP was activated, any plugin won't compile. And this is default when starting with a fresh workspace.

szarnekow:
Please explain how you would use the wizard without Xtext installed and therefore contained in the standard target platform.

szarnekow:
My bad - I misunderstood your comment I suppose.

szarnekow:
Would it be worthwhile to have the TP check for the wizard in general?

cdietrich:
i still have my doubts that we can make the wizard totally intelligent. we might even need an option to automatically set the target platform if the user requests it in a dialog

ArneDeutsch:
Can we merge this one to resolve https://github.com/eclipse/xtext-eclipse/issues/645 ? Right now simply the JUnit 4 checkbox is preselected, even if JUnit 5 is available in eclipse installation. With the change availablility of JUnit 5 is taken as a hint that JUnit 5 is desired and the correspondig checkbox is selected. From my point of view this is clearly an improvement (even if further improvements e.g. regarding target platform could be discussed).

PR #695, opened 2018-05-28
Fixed typo in PersistentDataAwareDirtyResource

tivervac:
You can just close this ticket if you don't think it's worth deprecating these methods for the typo. Whatever the response, I'll keep it in mind for future contributions.


cdietrich:
PersistentDataAwareDirtyResource is non api.
=> i am not sure if we simply should correct the typo. that do you think @szarnekow

szarnekow:
I did not find any callers on Github. So it's likely safe to rename the method.

cdietrich:
@tivervac could you change the pr accordingly

cdietrich:
many thanks @tivervac

PR #696, opened 2018-05-28
[NONE] Add StatemachineParsingTest to the 'Xtext State-Machine' example.

miklossy:
- Extend the GenerateStatemachine.mwe2 workflow to generate the runtime
test plugin.
- Modify the generated StatemachineParsingTest.xtend file to verify that
the fowler statemachine example can be parsed without any errors.


kthoms:
Thx @miklossy !

PR #697, opened 2018-05-29
[eclipse/xtext-eclipse#681] fixed test not resetting its state

cdietrich:
[eclipse/xtext-eclipse#681] fixed test not resetting its state


cdietrich:
please review

kthoms:
I'd prefer an `@After` method for the clean up

cdietrich:
can you please give a sample? i just repeated that pattern that is already there

kthoms:
```
	@After
	public void resetOutputConfigurations () {
		BuilderPreferenceAccess.Initializer initializer = new BuilderPreferenceAccess.Initializer();
		initializer.setOutputConfigurationProvider(new OutputConfigurationProvider());
		initializer.initialize(preferenceStoreAccess);
	}
```
This should make the pattern from the other method also obsolete.

cdietrich:
updated pr. will see in two hours if this still solves the problem

PR #701, opened 2018-05-31
Bugfix related to move refactoring of non-Xtext resources

meysholdt:
Since the refactoring applies content changes before applying move
changes, the content changes must be applied to the old URI.


meysholdt:
build failure seems unrelated.

cdietrich:
If there are unrelated build failures and there is not already a github issue for it please open one

PR #705, opened 2018-06-04
Fixed Java Model Exception: Java Model Status [ "folder "does not exist]

arshad1990:
problem when opening a file from File search.

Signed-off-by: Arshad <a.adavani@itemis.com>

### DIFF
--- a/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/trace/TraceForTypeRootProvider.java
+++ b/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/trace/TraceForTypeRootProvider.java
@@ -15,6 +15,7 @@ import org.apache.log4j.Logger;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IStorage;
+import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
@@ -40,6 +41,7 @@ import org.eclipse.xtext.ui.generator.trace.IEclipseTrace;
 import org.eclipse.xtext.ui.generator.trace.ITraceForStorageProvider;
 import org.eclipse.xtext.util.Pair;
 import org.eclipse.xtext.util.Tuples;
+import org.eclipse.xtext.util.internal.Nullable;
 
 import com.google.common.base.Charsets;
 import com.google.inject.Inject;
@@ -73,6 +75,10 @@ public class TraceForTypeRootProvider implements ITraceForTypeRootProvider {
 
 	@Inject
 	private ITraceForStorageProvider traceForStorageProvider;
+	
+	@Inject(optional=true) 
+	@Nullable
+	private IWorkspace workspace;
 	
 	private Pair<ITypeRoot, IEclipseTrace> lruCache = null;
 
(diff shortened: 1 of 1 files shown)


cdietrich:
@szarnekow can you please have a second look on that

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/trace/TraceForTypeRootProvider.java
```diff
@@ -180,13 +181,11 @@ private IEclipseTrace createTraceToSource(final ITypeRoot derivedJavaType) {
...
 
 	public IEclipseTrace getTraceToSource(final ICompilationUnit javaFile) {
-		try {
-			IResource resource = javaFile.getUnderlyingResource();
-			if (resource instanceof IStorage)
-				return traceForStorageProvider.getTraceToSource((IStorage) resource);
-		} catch (JavaModelException e) {
-			log.error("Error finding trace to source", e);
-		}
+		IWorkspaceRoot workspace = ResourcesPlugin.getWorkspace().getRoot();
```
szarnekow: Since we do already have a resource at hand (within the javaFile), I doubt that the lookup from scratch does a better job.
Would it make sense to check the JavaModelException for e.isDoesNotExist and skip logging the error in these cases?

arshad1990: Hi Sebastian, When you say better, do you mean Performance or the correctness of the fix provided ? I did some File Search tests and looks like it works fine. Do you see any problems here or you prefer guarding this case and skipping the log ?

cdietrich: well to avoid the javamodelexception we would check the compilation unit for isOpen and call the getUnderlying resource only when it is. this pr targets to get it running even if the file is not in a source folder

szarnekow: I see. In that case I'd move the proposed solution into the catch block.

cdietrich: And what to catch for?

cdietrich: sth like

```
	public IEclipseTrace getTraceToSource(final ICompilationUnit javaFile) {
		IResource resource = null;
		try {
			resource = javaFile.getUnderlyingResource();
		} catch (JavaModelException e) {
			IWorkspaceRoot workspace = ResourcesPlugin.getWorkspace().getRoot();
			resource = workspace.findMember(javaFile.getPath());
		}
		if (resource instanceof IStorage)
			return traceForStorageProvider.getTraceToSource((IStorage) resource);
		return null;
	}
```
?

szarnekow: Yes, something along these lines looks good to me.

szarnekow: Minor: The workspace can be injected.

cdietrich: would need an optional=true + a null check

arshad1990: I have fixed the code according to the review comments.

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/trace/TraceForTypeRootProvider.java
```diff
@@ -73,6 +76,10 @@
 	@Inject
 	private ITraceForStorageProvider traceForStorageProvider;
 	
+	@Inject(optional=true)
```
cdietrich: i had a look. this is not needed. we assume the workspace beeing there all over the place. => you can remove the optional=true and nullable

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/trace/TraceForTypeRootProvider.java
```diff
@@ -180,16 +187,29 @@ private IEclipseTrace createTraceToSource(final ITypeRoot derivedJavaType) {
...
+			}
 		}
+		if (resource instanceof IStorage)
+			return traceForStorageProvider.getTraceToSource((IStorage) resource);
 		return null;
 	}
 
+	private IResource getResourceFromNonSrcFolder(final ICompilationUnit javaFile) {
+		IResource resource = null;
+		if(null != workspace) {
```
cdietrich: null check not needed. see above

arshad1990: Done.

cdietrich:
thx @arshad1990

PR #708, opened 2018-06-11
Prevent NPE in outlines when eObject is null

tivervac:
This can happen when the `getEObjectURI()` is out of sync with the XtextResource.

It occurred to me in the `getChildren()` call when I was adding a new element to my file and then pressing Ctrl+Z at exactly the right time. This made the `getEObjectURI()` return the new element, but it was queried on an XtextResource that no longer had said item.

Of course you can fix this in your implementation by adding a dispatch for `getChildren()` that handles `Void` items, but this would have to be done for every outline implementation and it doesn't seem useful to me at all to try to create an outline node for a null item.

This PR needs API from https://github.com/eclipse/xtext-core/pull/766.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/outline/impl/AbstractOutlineNode.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/outline/impl/AbstractOutlineNode.java
@@ -203,28 +203,24 @@ public abstract class AbstractOutlineNode implements IOutlineNode, IOutlineNode.
 
 	@Override
 	public <T> T readOnly(final IUnitOfWork<T, EObject> work) {
-		if (getEObjectURI() != null) {
-			try {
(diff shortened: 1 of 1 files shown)


cdietrich:
will this break client that have a dispatch for void?

tivervac:
It would no longer jump into void dispatches no. But to have an outline that doesn't "randomly" throws errors, every outline implementation needs to have a dispatch over Void (which probably does nothing).

I see it as a bug fix, but if you're afraid of the backwards compatibility that's ok as well. We might have to update the documentation instead then.

cdietrich:
i could live with that breaking change. @szarnekow what do you think?

szarnekow:
Do we have any meaningful dispatch over Void in one of the outline trees?

cdietrich:
i did not find any (in xtext code base)

cdietrich:
thx @tivervac

Issue #699, opened 2018-05-31
Improve the Formatter of the Xtext State-Machine Example


miklossy:
Currently, the Xtext State-Machine Example Formatter contains only the generated stub based on the AbstractDeclarativeFormatter API. 

- The formatter should be switched to the AbstractFormatter2 API.
- The generated formatter stub should be implemented to improve the statemachine dsl formatting result.

cdietrich:
Fixed in 2.15

PR #712, opened 2018-06-17
[#699] Improve the Formatter of the Xtext State-Machine Example.

miklossy:
- Ensure that the formatter inserts an empty line between the events and
the commands block (if no resetEvents block), between the events and the
states block (if no resetEvents block and or commands block), and
between the resetEvents and the states block (if no commands block).
- Implement corresponding StatemachineFormatterTest test cases.


cdietrich:
Thx @miklossy

Issue #736, opened 2018-07-07
Provide ui test infrastructure for testing the syntactical/semantical highlighting.


miklossy:
It would be useful to extend the Xtext ui test infrastructure to enable the comfortable testing of the syntactical/semantical highlighting behaviour.

See also https://www.eclipse.org/forums/index.php/t/1066860/

PR #737, opened 2018-07-07
[#736] Add syntactical/semantical highlighting ui test infrastructure.

miklossy:
AbstractHighlightingTest abstract class to provide a base infrastructure
for testing the (customized) syntactical/semantical highlighting
behaviour.
- Extend the State-Machine Xtext example by the
StatemachineHighlightingTest to demonstrate the usage of the
AbstractHighlightingTest test class.

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
@@ -1,0 +1,170 @@
+/*******************************************************************************
+ * Copyright (c) 2018 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package org.eclipse.xtext.ui.testing
+
+import com.google.inject.Inject
+import org.eclipse.core.resources.IFile
+import org.eclipse.swt.custom.StyleRange
+import org.eclipse.swt.custom.StyledText
+import org.eclipse.swt.graphics.Color
+import org.eclipse.xtext.resource.FileExtensionProvider
+import org.eclipse.xtext.ui.XtextProjectHelper
+import org.eclipse.xtext.ui.editor.XtextEditorInfo
+import org.eclipse.xtext.ui.editor.utils.TextStyle
+import org.eclipse.xtext.ui.refactoring.ui.SyncUtil
+import org.eclipse.xtext.ui.testing.util.IResourcesSetupUtil
+
+import static extension org.eclipse.xtext.ui.testing.util.IResourcesSetupUtil.addNature
+
+/**
+ * @since 2.15
+ */
+abstract class AbstractHighlightingTest extends AbstractEditorTest {
+
+	@Inject XtextEditorInfo editorInfo
+
+	@Inject extension SyncUtil
+	@Inject extension FileExtensionProvider
+
+	override protected getEditorId() {
+		editorInfo.getEditorId
+	}
+
+	/**
+	 * @param it The editor's input text. The input text must contain the given <code>text</code>.
+	 * @param text The text that's highlighting style is to test.
+	 * @param style The expected highlighting configuration text style used when highlighting the given <code>text</code>. 
+	 */
+	def testHighlighting(CharSequence it, String text, TextStyle style) {
+		testHighlighting(text, style.style, 
+			if (style.color===null) 0 else style.color.red,
+			if (style.color===null) 0 else style.color.green,
+			if (style.color===null) 0 else style.color.blue,
+			if (style.backgroundColor===null) 255 else style.backgroundColor.red,
+			if (style.backgroundColor===null) 255 else style.backgroundColor.green,
+			if (style.backgroundColor===null) 255 else style.backgroundColor.blue
+		)
+	}
+
+	/**
+	 * @param it The editor's input text. The input text must contain the given <code>text</code>.
+	 * @param text The text that's highlighting style is to test.
+	 * @param fontStyle The expected font style used when highlighting the given <code>text</code>. 
+	 * @param foregroundR The red component of the expected foreground color when highlighting the given <code>text</code>.
+	 * @param foregroundG The green component of the expected foreground color when highlighting the given <code>text</code>.
+	 * @param foregroundB The blue component of the expected foreground color when highlighting the given <code>text</code>.
+	 */
+	def testHighlighting(CharSequence it, String text, int fontStyle, int foregroundR, int foregroundG, int foregroundB) {
+		testHighlighting(text, fontStyle, foregroundR, foregroundG, foregroundB, 255, 255, 255)
+	}
+
+	/**
+	 * @param it The editor's input text. The input text must contain the given <code>text</code>.
+	 * @param text The text that's highlighting style is to test.
+	 * @param fontStyle The expected font style used when highlighting the given <code>text</code>. 
+	 * @param foregroundR The red component of the expected foreground color when highlighting the given <code>text</code>.
+	 * @param foregroundG The green component of the expected foreground color when highlighting the given <code>text</code>.
+	 * @param foregroundB The blue component of the expected foreground color when highlighting the given <code>text</code>.
+	 * @param backgroundR The red component of the expected background color when highlighting the given <code>text</code>.
+	 * @param backgroundG The green component of the expected background color when highlighting the given <code>text</code>.
+	 * @param backgroundB The blue component of the expected background color when highlighting the given <code>text</code>.
+	 */
+	def testHighlighting(CharSequence it, String text, int fontStyle,
+		int foregroundR, int foregroundG, int foregroundB, 
+		int backgroundR, int backgroundG, int backgroundB) {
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,171 @@
...
+
+	@Inject XtextEditorInfo editorInfo
+	@Inject extension FileExtensionProvider
+
+	override protected getEditorId() {
+		editorInfo.getEditorId
+	}
+
+	/**
+	 * @param it - The editor's input text. The input text must contain the given <i>text</i>.
```
szarnekow: I think the '-' character does not need to be put after the param name.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,171 @@
...
+		dslFile
+		// when
+		.openInEditor
+		// then
+		.text(text).isHighlightedIn(fontStyle, foregroundR, foregroundG, foregroundB, backgroundR, backgroundG, backgroundB)
+	}
+
+	protected def dslFile(CharSequence content) {
+		val file = IResourcesSetupUtil.createFile(projectName, fileName, fileExtension, content.toString)
+		file.project.addNature(XtextProjectHelper.NATURE_ID)
```
szarnekow: This will unnecessarily slow down all the tests I suppose. The project should have the nature already assigned from the outside.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,171 @@
...
+		file
+	}
+
+	protected def String getProjectName() '''Test'''
```
szarnekow: Please use a more distinct project name.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,171 @@
...
+
+	protected def String getFileName() '''Foo'''
```
szarnekow: And file name

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,171 @@
...
+		/*
+		 * wait for the Xtext framework HighlightingPresenter.updatePresentation()
+		 * to apply the semantic highlighting executed asynchronously
+		 */
+		waitForEventProcessing
+	
+		editor.internalSourceViewer.textWidget
+	}
+
+	protected def text(StyledText styledText, String text) {
```
szarnekow: What does this method do?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,171 @@
...
+			}
+		}
+	}
+
+	protected def isRelevant(String character) {
+		// skipping the whitespace characters
+		character == character.trim 
+	}
+
+	protected def hasFontStyle(StyleRange it, String character, int expected) {
```
szarnekow: This is more an assertFontStyle

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,171 @@
...
+		val actual = fontStyle
+		assertEquals('''Expected font style does not correspond to the actual font style on character �character�''',
+			expected, actual)
+	}
+
+	protected def hasForegroundColor(StyleRange it, String character, Color expected) {
```
szarnekow: Same here and below

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,171 @@
...
+			expected, actual)
+	}
+
+	protected def hasBackgroundColor(StyleRange it, String character, Color expected) {
+		val actual = background ?: new Color(null, 255, 255, 255) // the default background color is white 
+		assertEquals('''Expected background color does not correspond to the actual background color on character �character�''',
+			expected, actual)
+	}
+
+	protected def waitForEventProcessing() {
```
szarnekow: There should be smth available in SyncUtil or close by.

org.eclipse.xtext.xtext.ui.examples/projects/fowlerdsl/org.eclipse.xtext.example.fowlerdsl.ui.tests/src/org/eclipse/xtext/example/fowlerdsl/ui/tests/StatemachineHighlightingTest.xtend
```diff
@@ -0,0 +1,182 @@
...
+	@Test def events_keyword() {
+		'''
+			events
+				doorClosed   D1CL
+				drawerOpened D2OP
+				lightOn      L1ON
+				doorOpened   D1OP
+				panelClosed  PNCL
+			end
+		'''.testHighlighting("events", SWT.BOLD, 127, 0, 85)
```
szarnekow: My gut feeling is, that we'd rather like to check that a certain style is applied from the highlighting configuration instead of checking the colors directly.

miklossy:
Thanks for the review @szarnekow. I updated the PR.

The State-Machine Example still checks the color values directly, but the AbstractHighlightingTest class already provides utility methods to use text styles from the highlighting configuration. I would suggest to leave the State-Machine Example to use the colors directly and add a demonstrator for using text styles to another Xtext Example in a follow up PR.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
szarnekow: Please add at least a check if the nature is already present (or does addNature(..) perform such a short-circuit already?)

miklossy:
Thanks for the feedback, @szarnekow . I guarded the addNature call so that it will be only executed if needed. Since I am also not satisfied how the nature is currently set, I also added a TODO statement to keep track of the remaining work.

```
	/*
	 * TODO: find a better (with good performance) solution
	 * to set the Xtext nature on the test project.
	 */
	val project = file.project
	if(!project.hasNature(XtextProjectHelper.NATURE_ID)) {
		project.addNature(XtextProjectHelper.NATURE_ID)
	}
```

 Since it requires a deeper investigation (and I think also effects other test cases), I would create a separate follow up issue for that.

szarnekow:
LGTM

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,185 @@
...
+
+	@Inject extension SyncUtil
+	@Inject extension FileExtensionProvider
+
+	override protected getEditorId() {
+		editorInfo.getEditorId
+	}
+
+	/**
+	 * @param it The editor's input text. The input text must contain the given <i>text</i>.
```
kthoms: if you refer to classes, methods, files, better use `<code>` instead of `<i>`

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.xtend
```diff
@@ -0,0 +1,185 @@
...
+	 *
+	 * @param text The text that's highlighting style is to test.
+	 *
+	 * @param style The expected highlighting configuration text style used when highlighting the given <i>text</i>. 
+	 */
+	def testHighlighting(CharSequence it, String text, TextStyle style) {
+		testHighlighting(text, style.style, 
+			if (style.color===null) 0 else style.color.red,
```
kthoms: you should be able to use the elvis operator here:
`style.color?.red ?: 0`

miklossy:
Thanks for the feedback @kthoms . I updated the PR regarding your javadoc suggessions. However, I could not use the elvis operator, because of the error message: 

`The operator '?:' is undefined for arguments of type int`

cdietrich:
@kthoms @szarnekow @miklossy what is the state of this pr?

miklossy:
From my side it is done, as soon as it is merged, I will create one follow up PR to another Xtext Example to demonstrate the usage of asserting on text styles instead of colors.

Apart from that I will create a new issue for optimizing the Xtext Nature configuration on the test projects.

szarnekow:
My +1 still holds. Nice.

kthoms:
Thanks, @miklossy

PR #744, opened 2018-07-10
fixed warnings after switch to oxygen as minimal target

cdietrich:
fixed warnings after switch to oxygen as minimal target


cdietrich:
why will this break luna?

kthoms:
Because you make now use of the generics (`<T> T getAdapter(Class<T> adapter)`) that have been introduced after Luna.

szarnekow:
> This will definitely break on Luna.
Do you talk about binary incompatibility? I assume the Java compiler produces bridge methods for the generified methods.

szarnekow:
From my understanding we did not change the code to be source compatible to the older TP. Otherwise we could not have built our stuff. Am I mistaken?

cdietrich:
yes but afaik the code should still work. (see the code snippet i sent you via chat)
but i am ok with setting the minmal version

szarnekow:
So am I. It will definitely not do any harm in the effort to keep the deps straight.

kthoms:
PRs for that are on the way, e.g. https://github.com/eclipse/xtext-eclipse/pull/745

cdietrich:
@szarnekow @kthoms ok to merge?

PR #745, opened 2018-07-11
[#1168] Minimal target platform Oxygen.3a

kthoms:
Update lower bound version constraints on platform related bundles with
versions from an Oxygen.3a platform.

Used an Oxygen.3a Target Platform and selected the proposed versions from it:

screenshot 247


kthoms:
rebased

cdietrich:
still would like to get a "minimal" version of this into 2.16

szarnekow:
What is necessary to make this a mergable pr?

kthoms:
Seems that the versions are outdated a bit. With latest Oxygen target I confirm that the mentioned versions (jdt.core, pde.ui) changed. Have to check this.

cdietrich:
@kthoms @szarnekow anything left?

kthoms:
Should be Ok now?

Issue #755, opened 2018-07-18
Xtext 'Domain-Model Example' failing UI test case.


miklossy:
When creating a fresh Xtext 'Domain-Model Example' (Xtext Complete SDK	2.15.0.v20180716-0209) into a Photon Workspace on Windows, the ContentAssistTest.testTemplateProposal is failing:

![screenshot](https://user-images.githubusercontent.com/20393472/42862546-43ce7ddc-8a60-11e8-8eea-eeb0accc21e5.png)

Expected line ending: Windows style
Actual line ending: Unix sytle

@ArneDeutsch Could you please execute this test case on Windows and give feedback wheter you can reproduce this issue?

cdietrich:
@miklossy i can confirm the failing test on windows.
i am not even sure if this is a test or a template problem

i think the template contains windows style endings.
but when inserted there is linux style endings so i wonder where
the replacement happens and why. maybe already in template reader.

would be interesting to know how other templates e.g from jdt behave in that regard.
but where are they replaced. maybe the templates are "re-platformized" on insert.

@szarnekow @kthoms what do you think

cdietrich:
p.s. i have a kind a deja-vu on a existng github issue or bugzilla but did not find one on the first glance

szarnekow:
I think it should be possible to customize the apply method of the template proposal. It should make sure that it uses the line delimiter as it is present in the document.

szarnekow:
see org.eclipse.xtext.ui.editor.templates.XtextTemplateProposal.apply(ITextViewer, char, int, int)

szarnekow:
Even better: org.eclipse.xtext.ui.editor.templates.XtextTemplateContext.IndentationAwareTemplateTranslator

miklossy:
I agree with @szarnekow that the template proposal should insert the templates using the line endings that already present in the document, independently of the line ending of the templates.xml file. Currently, the template proposal inserts the entity template of the Domain-Model Example using Unix style endings even if the document has Window style line delimiters:
![screenshot2](https://user-images.githubusercontent.com/20393472/42923072-e5b8b31a-8b23-11e8-8869-314e018d3712.png)

holgerschill:
Looks like this is fixed, right?

miklossy:
No, the problem still exists.

cdietrich:
@miklossy are you working on this issue? if not it would be a good candidate i think

ArneDeutsch:
[...]

PR #764, opened 2018-07-26
[#755] Implement additional DomainModel Example Template Proposal test.

miklossy:
- The more tests are detecting the bug, the better the fix will be.


cdietrich:
thx @miklossy

Issue #768, opened 2018-07-31
No Run as Junit Test on XbaseIntegrationTest in Domainmodel Example


cdietrich:

miklossy:
Could you please give more context on this issue?

miklossy:
Same for

- Homeautomation.FormatterTest
- Homeautomation.ParserTest

All these test cases have inherited (no declared) test methods.

cdietrich:
Yes that is that I mean

cdietrich:
@miklossy  is this fixed now?

miklossy:
Yes, ticket can be closed.

cdietrich:
Fixed in 2.15

PR #770, opened 2018-07-31
[#768] Ensure that the Run As -> JUnit Test occurs on inherited tests.

miklossy:
- Modify the JavaElementDelegateJunitLaunch to consider the type
hierarchy (inherited test methods) by determining if an Xtend file
contains a JUnit test method.

### DIFF
--- a/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/launching/JavaElementDelegateJunitLaunch.java
+++ b/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/launching/JavaElementDelegateJunitLaunch.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2012, 2018 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


szarnekow:
Does JDT also create a type hierarchy for this? From my experience, this may be quite expensive.

ArneDeutsch:
The method `containsJUnitTestMethod` seems not consistent for me. It chechs two cases ... the 'old' way of defining a test method using naming conventino and the 'new' one, using an annotation. But for the old one access modifiers, visibility and so on are tested, wheras this is skipped for the annotation. Why? Seems like these checks should be done for the second case as well!?

miklossy:
The JDT [JUnitPropertyTester](https://github.com/eclipse/eclipse.jdt.ui/blob/be311952d25f246966de3de6f5bd80a96dfdb278/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/JUnitPropertyTester.java#L117) uses the CoreTestSearchEngine helper class that delegates to the [JUnit5TestFinder](https://github.com/eclipse/eclipse.jdt.ui/blob/6cd5bd624cf4023ceb02ba3ab8de84a223245eab/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/launcher/JUnit5TestFinder.java#L116-L134) class operating on org.eclipse.jdt.core.dom.ITypeBinding instances to determine if any of the contained method is annotated by JUnit annotations.

miklossy:
I did not recognize any performance problem while creating the type hierarchy.

Even if there is a small performance delay, it will only occur if the selected Xtend file does not contain any test cases. I would leave the implementation as it is in the PullRequest and optimize it later on on demand only if the performance delay becomes observable.

miklossy:
Thanks for the feedback @szarnekow ! I implemented the requested changes and updated the commit in the Pull Request.

cdietrich:
Thx @miklossy

Issue #772, opened 2018-08-02
investigate bogus behaviour of TargetPlatformUtil


cdietrich:
org.eclipse.xtext.ui.testing.util.TargetPlatformUtil.setTargetPlatform(Class<?>) fails when running single tests e.g. in QuickFixTest in Xtend (maybe on photon and later only)

=> xbase lib wont be found

additionally: depending on state of p2.pool etc the tp will explode => maybe its better if we can detect if we are started from eclipse and not from tycho and then skip that.

```
	public static void setTargetPlatform(Class<?> context) throws Exception {
		if (Boolean.getBoolean("eclipse.pde.launch")) return;
```

cdietrich:
Fixed in 2.15

PR #773, opened 2018-08-03
[#772] Don't do anything in TargetPlatformUtil when called from eclipse pde

cdietrich:
[#772] Don't do anything in TargetPlatformUtil when called from eclipse pde

### DIFF
--- a/org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/ui/util/TargetPlatformUtil.java
+++ b/org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/ui/util/TargetPlatformUtil.java
@@ -46,6 +46,9 @@ public class TargetPlatformUtil {
 	 * @since 2.14
 	 */
 	public static void setTargetPlatform(Class<?> context) throws Exception {
+		if (isPdeLaunch()) {
+			return;
+		}
 		Bundle currentBundle = FrameworkUtil.getBundle(context);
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/ui/util/TargetPlatformUtil.java
```diff
@@ -46,6 +46,7 @@ public static void setTargetPlatform() throws Exception {
 	 * @since 2.14
 	 */
 	public static void setTargetPlatform(Class<?> context) throws Exception {
+		if (Boolean.getBoolean("eclipse.pde.launch")) return;
```
szarnekow: Nitpick: I'd extract a method, e.g. isPdeLaunch() and also do a linebreak before the return. But in general this looks about right to me.

LorenzoBettini:
Sorry I came too late.
It's kind of strange it fails when running from Eclipse: setting the target platform should be inoffensive...
I see you already came with a solution. In my programs, I prefer to set the target platform (with something similar to the target platform util) only if a property is set (which I set from the Tycho build), but I guess also this solution is fine.

cdietrich:
@LorenzoBettini the problem here is oomph. the .p2 pool contains 10s of the same plugin that are all pulled into the target platform and then later pde cannot handle that. so basically the target platform does not work at all.

LorenzoBettini:
OK I see

Issue #747, opened 2018-07-11
Consider Archiving Workspace log files


cdietrich:
would be nice to have workspace log files for ui tests archived e.g. to investigate bugs e.g.
`target/work/data/.metadata/.log`

kthoms:
You mean on CI build?

cdietrich:
Yes in Jenkinsfile(s)

cdietrich:
Fixed in 2.15

PR #774, opened 2018-08-03
[#747] archive junit logs

cdietrich:
[#747] archive junit logs


szarnekow:
Perspectively we should use smth like `archiveArtifacts artifacts: 'screenshots/**,build/test/results/*.xml', allowEmptyArchive: true` - otherwise the build may fail if nothing to-be-archived can be found

cdietrich:
imho its intentional (.e.g. when no tests run)

cdietrich:
@szarnekow can you approve this, and open a follow up ticket for that issue? if then it should be fixed everywhere

PR #778, opened 2018-08-07
Made UI TargetURIConverter extends from non ui one

tivervac:
Needs https://github.com/eclipse/xtext-core/pull/810 to be merged first


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/TargetURIConverter.java
```diff
@@ -7,33 +7,11 @@
...
-	public TargetURIs fromIterable(Iterable<URI> uris) {
-		if (uris instanceof TargetURIs) {
-			return (TargetURIs) uris;
-		}
-		TargetURIs result = targetURIProvider.get();
-		result.addAllURIs(uris);
-		return result;
-	}
-	
+public class TargetURIConverter extends org.eclipse.xtext.findReferences.TargetURIConverter {
```
cdietrich: am not sure if we should mark this as deprecated @szarnekow what do you think

PR #779, opened 2018-08-08
[537780] marked optional injection with nullable so that it actually can be optional

cdietrich:
[537780] marked optional injection with nullable so that it actually can be optional


tivervac:
@cdietrich Sinci it is nullable there is a possible NPE in org.eclipse.xtext.ui.refactoring2.ChangeConverter.findOpenEditor(IFile)

I suggest that be fixed as well

cdietrich:
well then leonid either did not test or we did not talk about the same thing.

cdietrich:
=> refactorings shall no be executed headless

tivervac:
Just read the original ticket. The NPE probably doesn't surface because they don't open the editor in a headless environment?

tivervac:
@cdietrich LGTM

PR #783, opened 2018-08-09
[eclipse/xtext-xtend#528] Don't request a build when changing compile…

tivervac:
…r settings that don't require a rebuild

Signed-off-by: Titouan Vervack <tivervac@gmail.com>

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/preferences/OptionsConfigurationBlock.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/preferences/OptionsConfigurationBlock.java
@@ -70,6 +70,9 @@ public abstract class OptionsConfigurationBlock {
 	@Inject
 	protected PreferenceStoreAccessImpl preferenceStoreAccessImpl;
 
+	@Inject
+	private IBuildPreferenceEvaluator buildPreferenceEvaluator;
+
 	private static final String IS_PROJECT_SPECIFIC = "is_project_specific"; //$NON-NLS-1$
 	private static final String SETTINGS_EXPANDED = "expanded"; //$NON-NLS-1$
 	private static final String REBUILD_COUNT_KEY = "preferences_build_requested"; //$NON-NLS-1$
@@ -528,7 +531,7 @@ public abstract class OptionsConfigurationBlock {
 	}
 
 	protected boolean processChanges(IWorkbenchPreferenceContainer container) {
-		boolean needsBuild = !getPreferenceChanges().isEmpty();
+		boolean needsBuild = buildPreferenceEvaluator.isAffectingBuild(getPreferenceChanges());
 		boolean doBuild = false;
 		if (needsBuild) {
 			int count = getRebuildCount();
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/EclipseOutputConfigurationProvider.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/EclipseOutputConfigurationProvider.java
@@ -21,6 +21,7 @@ import org.eclipse.xtext.generator.OutputConfiguration;
 import org.eclipse.xtext.generator.OutputConfiguration.SourceMapping;
 import org.eclipse.xtext.ui.editor.preferences.IPreferenceStoreAccess;
 import org.eclipse.xtext.ui.editor.preferences.PreferenceConstants;
+import org.eclipse.xtext.ui.preferences.XtextBuildPreferenceEvaluator;
 import org.eclipse.xtext.ui.resource.ProjectByResourceProvider;
 
 import com.google.common.base.Strings;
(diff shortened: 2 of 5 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/AffectsBuilderPredicate.java
```diff
@@ -0,0 +1,43 @@
...
+	public boolean apply(Entry<String, ValueDifference<String>> input) {
+		String key = input.getKey();
+		if (key.startsWith(PREFIX)) {
+			String keyName = key.substring(key.lastIndexOf(PreferenceConstants.SEPARATOR) + 1);
+			switch (keyName) {
+				case OUTPUT_CREATE_DIRECTORY:
+				case OUTPUT_OVERRIDE:
+				case OUTPUT_DERIVED:
+				case OUTPUT_CLEANUP_DERIVED:
+				case OUTPUT_CLEAN_DIRECTORY:
```
cdietrich: @szarnekow i am not sure which of these should go where

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderConfigurationBlock.java
```diff
@@ -52,6 +55,8 @@
 public class BuilderConfigurationBlock extends OptionsConfigurationBlock {
 	
 	protected static final String[] BOOLEAN_VALUES = new String[] { IPreferenceStore.TRUE, IPreferenceStore.FALSE };
+
+	private static final Predicate<Entry<String, ValueDifference<String>>> AFFECTS_BUILDER_PREDICATE = new AffectsBuilderPredicate();
```
cdietrich: am not sure if we need a factory for that to be better configurable

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderConfigurationBlock.java
```diff
@@ -388,4 +393,8 @@ public String getPropertyPrefix() {
 		return PROPERTY_PREFIX;
 	}
 
+	@Override
+	protected Predicate<Entry<String, ValueDifference<String>>> getAffectsBuilderPredicate() {
```
cdietrich: OptionsConfigurationBlock is api so i wonder if we should make this method api too

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/preferences/OptionsConfigurationBlock.java
```diff
@@ -581,6 +584,16 @@ private void logError(String text, IOException e) {
 		uiPlugin.getLog().log(status);
 	}
 
+	public Map<String, ValueDifference<String>> getBuildAffectingPreferenceChanges() {
+		Map<String, ValueDifference<String>> changes = getPreferenceChanges();
+		Map<String, ValueDifference<String>> buildAffectingChanges = Maps.filterEntries(changes, getAffectsBuilderPredicate());
+		return buildAffectingChanges;
+	}
+
+	protected Predicate<Entry<String, ValueDifference<String>>> getAffectsBuilderPredicate() {
```
cdietrich: OptionsConfigurationBlock is api so i wonder if we should make this method api too

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderConfigurationBlock.java
szarnekow: I'd rather like to inject this one. Can we use a dedicated type here, though?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderConfigurationBlock.java
szarnekow: It should become API, yes. But I'd prefer a better type signature and easier means to customize it.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/preferences/OptionsConfigurationBlock.java
szarnekow: Should be make this a method `hasBuildAffectingPreferenceChanges()` that delegates to a utility?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderConfigurationBlock.java
szarnekow: Maybe a type with a method `isAffectingBuild(Map<String, ValueDiff<String>>)` would be easier to grasp and allow for easier customizing, too?

org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/XtextUiModule.java
```diff
@@ -184,5 +186,8 @@ public void configureIResourceDescriptionsPersisted(com.google.inject.Binder bin
 	public void configureEditorScope(Binder binder) {
 		binder.bindConstant().annotatedWith(Names.named(XtextEditor.KEY_BINDING_SCOPE)).to("org.eclipse.xtext.xtext.ui.XtextEditorScope");
 	}
-	
+
```
cdietrich: @tivervac why is this here and not in DefaultUiModule?

tivervac: Because I thought that was the place to put it, seems I was wrong ;)

cdietrich:
@szarnekow any ideas how to resolve that without introducing a cycle?

cdietrich:
or we use simply `@ImplementedBy`

tivervac:
@cdietrich Using `@ImplementedBy` we still need to access to xtext.builder from xtext.ui, don't we? If so, that doesn't solve anything

cdietrich:
dont get that: cant everything be moved to builder? what is needed in ui?

cdietrich:
i am clueless. maybe the option and the builder page should be separated so that everything can be moved to builder

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/preferences/OptionsConfigurationBlock.java
```diff
@@ -528,7 +531,7 @@ private void incrementRebuildCount() {
 	}
 
 	protected boolean processChanges(IWorkbenchPreferenceContainer container) {
-		boolean needsBuild = !getPreferenceChanges().isEmpty();
+		boolean needsBuild = buildPreferenceEvaluator.isAffectingBuild(getPreferenceChanges());
```
cdietrich: why is buider here in a generic class

cdietrich: maybe a simple method here and the delegate to the evaluator in subclass

tivervac: @szarnekow requested an injectable API for better customization

szarnekow: This is so much entangled, it's a shame, eg see this code below `container.registerUpdateJob(getBuildJob(getProject()));`

Bindings required for the builder are currently done by the Xtext generator, which I'd like to avoid.

It would be a safe bet if we moved the XtextBuildPreferenceEvaluator and defined the constant for the output config in ui and referenced it from `EclipseOutputConfigurationProvider.OUTPUT_PREFERENCE_TAG`

tivervac:
@szarnekow Is this what you meant?

szarnekow:
Yes exactly. What do you think? Is this a feasible approach?

tivervac:
I think it's a bit nasty to declare those constants in UI instead of in the builder package but it's a lot better than generating it. Seems feasible to me.

tivervac:
@szarnekow Is this ready to be merged now?

cdietrich:
should be done after version bump

szarnekow:
So the last remaining bit would be `@since 2.16`

PR #786, opened 2018-08-14
marked old TargetURIConverter as deprecated

cdietrich:
marked old TargetURIConverter as deprecated


cdietrich:
@szarnekow last week we were not sure how to deal with api changes that move stuff to core. i think we should at least mark the old class as deprecated but i am not sure if we should adapt the clients to use the new (non) API instead of the old one

cdietrich:
an extended version would look like https://github.com/eclipse/xtext-eclipse/pull/787

PR #808, opened 2018-08-28
ValidatorFragment2 generates an overridden version of addComboBox for…

holgerschill:
… Xbase languages xtext-core#771

Depends on PR xtext-core#840


holgerschill:
Green here http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/hs%252Fissue_xtext_core771/3/console

Issue #433, opened 2017-10-23
Builder: Improve Progress Reporting


cdietrich:
we should consider improoving the progress reporting in eclipse during build.
maybe we can facilitate new features in more recent versions once we moved tp to oxygen 

e.g. Progress Monitor shows "refeshing" although refreshing has finished before
The Progress Monitor changes its label to "Refreshing" at the places where we call `IResource.refreshLocal` and keeps that label even after finishing the refresh.
this gives the user the wrong impression all the time is spent on "refreshing"

cdietrich:
to be checked: is this an issue with platform and is the behaviout still there with oxygen

cdietrich:
![progressmonitor](https://user-images.githubusercontent.com/238736/31887229-d4bdfef8-b7f7-11e7-8cb3-aa5e0f82873d.png)

cdietrich:
cc @iloveeclipse

iloveeclipse:
Christian, I would guess that someone in the call chain of refresh code uses monitor in a wrong way. If this would be a platform bug, we would have more clients complaining.

cdietrich:
that is the "correct" way of calling?

iloveeclipse:
Let say after refresh no one opens new task and the old one is not finished

cdietrich:
so its not intended to call refresh within a block of other stuff

iloveeclipse:
See examples how to use API at https://github.com/iloveeclipse/ece2016-progressmonitors.git

iloveeclipse:
> so its not intended to call refresh within a block of other stuff

One must inspect the code to answer this

holgerschill:
Fixed in 2.15

PR #811, opened 2018-08-29
Builder: Improve Progress Reporting #433

holgerschill:


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java
```diff
@@ -590,9 +587,13 @@ protected void handleChangedContents(Delta delta, IBuildContext context, Eclipse
 		saveResourceStorage(resource, fileSystemAccess);
 		if (shouldGenerate(resource, context)) {
 			try {
+				IProgressMonitor fsaMonitor = fileSystemAccess.getMonitor();
 				MonitorBasedCancelIndicator cancelIndicator = new MonitorBasedCancelIndicator(fileSystemAccess.getMonitor());
```
szarnekow: Please use the fsaMonitor for the cancelIndicator, too instead of reading it again from the fileSystemAccess.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java
```diff
@@ -590,9 +587,13 @@ protected void handleChangedContents(Delta delta, IBuildContext context, Eclipse
...
 				GeneratorContext generatorContext = new GeneratorContext();
 				generatorContext.setCancelIndicator(cancelIndicator);
+				// Since we make use of SubMonitor.split instead of SubMonitor.newChild
+				// there is no reason anymore to avoid showing each resource.
+				fsaMonitor.subTask("Compile " + resource.getURI());
```
szarnekow: The URI is not necessarily nice to read. We should probably use only the last segment.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/EclipseResourceFileSystemAccess2.java
```diff
@@ -533,12 +533,15 @@ protected IFile getTraceFile(IFile file) {
...
+		syncIfNecessary(result, this.monitor);
 	}
 
 	private void syncIfNecessary(IResource result, IProgressMonitor progressMonitor) {
 		try {
-			sync(result, IResource.DEPTH_ZERO, progressMonitor);
+			// org.eclipse.core.internal.resources.Resource.refreshLocal(int, IProgressMonitor) starts a new subtask
+			// that we don't want to propagate to the parent monitor.
+			SubMonitor subMonitor = SubMonitor.convert(monitor, 1);
+			sync(result, IResource.DEPTH_ZERO, subMonitor.split(1, SubMonitor.SUPPRESS_ALL_LABELS));
```
szarnekow: Should the call of syncIfNecessary already suppress the labels?

holgerschill: sync comes from the ResourceUtil that we don't want to change -> so no other place to do that and sync always calls Resource.refreshLocal.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java
holgerschill: Done.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java
holgerschill: Done.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/clustering/ClusteringBuilderState.java
```diff
@@ -115,7 +114,12 @@
      */
     @Override
     protected Collection<Delta> doUpdate(BuildData buildData, ResourceDescriptionsData newData, IProgressMonitor monitor) {
-        final SubMonitor progress = SubMonitor.convert(monitor, 100);
+        // We assume that we have 101 ticks to work with:
```
szarnekow: Comment is out of sync with the 100 below?

PR #818, opened 2018-09-05
State-Machine Example QuickfixProvider - Use more meaningful method name.

miklossy:
- Change the methode name from 'capitalizeName' to
'convertNameToFirstLowerCase' to better indicate what the quickfix does.


cdietrich:
thx @miklossy

PR #823, opened 2018-09-11
Added comment about possible null return

tivervac:
Comment is based on org.eclipse.ui.texteditor.AbstractTextEditor.getSourceViewer().
Is it to specific on that interface? If so, what would be better? It can definitely be null


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateEditorSupport.java
```diff
@@ -245,6 +245,9 @@ public boolean isConcurrentEditingIgnored(IDirtyStateEditorSupportClient client)
 	 */
 	public interface IDirtyStateEditorSupportClient {
 
+		/**
+		 * May return <code>null</code before EditorPart creation or after its disposal.
```
cdietrich: type missing >

tivervac:
@cdietrich Fixed the typo

PR #825, opened 2018-09-14
Reduce visibility of StatemachineGenerator helper methods to protected.

miklossy:


miklossy:
Thanks for the feedback @kthoms ! I agree with you that for testing purposes it is better to use protected visibility for the helper methods. I updated the commit accordingly.

PR #826, opened 2018-09-17
Use tryReadOnly instead of readOnly

tivervac:
Didn't replace any occurrences in tests, also avoided a few uses where it could change behavior.


tivervac:
Seems like I have some more work to do, no need to review this until the PR is stable

szarnekow:
What is the status here?

tivervac:
I was having issues getting jenkins to pick up my branch on xtext-core, will try to get it fixed

tivervac:
Requires [this PR in xtext-core](https://github.com/eclipse/xtext-core/pull/966)

tivervac:
Ouch, I forgot about this, I'll see if I can get it back up to speed

cdietrich:
@tivervac @szarnekow what is the state of this pr

kthoms:
rebased

tivervac:
If the build is green this should be mergeable.
@kthoms I don't see build information, is this due to the switch to eclipse ci?

PR #835, opened 2018-09-29
Xtext State-Machine Example: Eliminate discouragedReference warnings.

miklossy:
- Make the Formatter2 API classes accessible by defining an explicit
access rule on the org.eclipse.xtext.formatting2 package.


miklossy:
@kthoms Thanks for making me aware of the explicit access rules. I updated the PR accordingly.

Issue #828, opened 2018-09-21
Update Xtext Setup


cdietrich:
Xtext.setup uses outdated IBuilds for 4.9 and offers no entry for 4.10/2018-12

cdietrich:
cc @kthoms

cdietrich:
@kthoms can you work on this or provide steps what i have to do.
i get failures on this all the time

cdietrich:
Fixed in 2.16

PR #846, opened 2018-10-05
[#828] Updated Setup to 4.9 final. introduced 4.10-ibuilds

cdietrich:
[#828] Updated Setup to 4.9 final. introduced 4.10-ibuilds

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -56,14 +56,14 @@
     <annotation
         source="http://www.eclipse.org/oomph/setup/GlobalVariable"/>
     <choice
+        value="2018-12"
+        label="Eclipse 2018-12 - 4.10"/>
+    <choice
         value="2018-09"
         label="Eclipse 2018-09 - 4.9"/>
     <choice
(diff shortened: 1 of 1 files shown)


releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -884,7 +887,40 @@
           <repository
               url="${p2.draw2d}"/>
           <repository
-              url="http://download.eclipse.org/eclipse/updates/4.9-I-builds"/>
+              url="http://download.eclipse.org/eclipse/updates/4.9-I"/>
```
kthoms: remove

kthoms:
Modular target "Photon - Integration" can be removed

releng/org.eclipse.xtext.contributor/Xtext.setup
kthoms: The additional eclipse/updates repository is only needed when the base platform target should be overridden. The Eclipse Platform already defines the typical main repos for each release. Only for the upcoming release we need the I-build repo additionally.

cdietrich:
updated the pr @kthoms

Issue #311, opened 2017-07-24
Extract core functionality - support core eclipse platform (i.e. headless cli tools)


yoosiba:
Eclipse Platform separates UI and non-UI bundles which allows to build headless applications based on it. Xtext (as _a framework for development of programming languages_) should support headless tools for languages. Currently xtext-eclipse depends on both UI and non-UI dependencies which makes providing headless tooling for the custom language very limited.

Example: creating headless tool (i.e. compiler) for xtext based language. The compiler can be based on Eclipse Platform (`org.eclipse.core...`), but once it needs to use some xtext-eclipse type (e.g. `IBuilderState`) it has transitive depency to UI bundles (`org.eclipse.ui`, `org.eclipse.xtext.ui`, ...). It is possible to use some workarounds, for example:
 * provide alternative implementations for xtext-eclipse internals (i.e. copy&paste classes from xtext-eclipse and their dependencies from eclipse platform, maintain them across eclipse and xtext releases)
 * use execution environment that does not validate bundle dependencies at runtime and do not provide UI dependencies, _hope_ that no code execution path calls into missing dependency

But all of those are just workarounds for xtext-eclipse limitation.

Extracting xtext-eclipse-core with no dependencies to UI would allow better CLI tools that can take advantage of both Xtext framework and Eclipse Platform without need for dirty hacks.

meysholdt:
Xtext does have support for headless incremental build/compile already: see `org.eclipse.xtext.build.IncrementalBuilder`. Is that what you've been looking for?

cdietrich:
we have standalone and we have eclipse.
we dont have headless eclipse

meysholdt:
I see. This discussion came up several times over the last years and we decided to not address this issues since the cost of breaking *a lot* of API outweighed the benefits. 

Instead we put the effort in extracting a crore that does not depend on Eclipse, neither UI nor headless: http://typefox.io/xtext-core-less-is-more

cdietrich:
Yes and even if you want a headless eclipse it's not too much you need to copy and pasting to a custom headlessuimodule and activator

yoosiba:
I understand inconvenience of doing that, or small ROI for current user base.

[...]

PR #855, opened 2018-10-11
[#311] Use consistent encoding in CompilationTestHelper

szarnekow:


kthoms:
Content OK, but the commit message must use "eclipse/xtext-extras#311" for a cross repo reference. #311 is wrong linked.

kthoms:
updated

Issue #836, opened 2018-10-01
CodeMining that works on Photon does not work on 2018-09.


miklossy:
CodeMining on Photon:
![codemining_photon_ok](https://user-images.githubusercontent.com/20393472/46279465-72057b80-c569-11e8-8288-8ac1588eb1b2.gif)

CodeMining on 2018-09:
![codemining_2018-09_nok](https://user-images.githubusercontent.com/20393472/46279473-7893f300-c569-11e8-9bd5-ac5ed3d47dc5.gif)

The reason is that the [AbstractXtextCodeMiningProvider](https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.ui.codemining/src/org/eclipse/xtext/ui/codemining/AbstractXtextCodeMiningProvider.java#L144) creates a org.eclipse.jface.text.Position object setting only the offset but not the length.

With the [commit](https://github.com/eclipse/eclipse.platform.text/commit/d3f94a16e823f8bf57175e439539192c51c86267) on the eclipse.platform.text the behavior has been changed that all positions that's length is 0 considered as [deleted](https://github.com/eclipse/eclipse.platform.text/blob/master/org.eclipse.jface.text/src/org/eclipse/jface/text/source/inlined/InlinedAnnotationDrawingStrategy.java#L348).

The solution would be to modify the AbstractXtextCodeMiningProvider to set not only the offset, but also the length of the Position object.

cdietrich:
We need tests

miklossy:
It is quite difficult to cover this behaviour with junit tests, since the annotations are [draw](https://github.com/eclipse/eclipse.platform.text/blob/master/org.eclipse.jface.text/src/org/eclipse/jface/text/codemining/AbstractCodeMining.java#L146) on the StyledText canvas.

kthoms:
Merged for 2.16

cdietrich:
@miklossy 

with the greeting grammar

and

```
package org.xtext.example.mydsl1.ui.codemining

import org.eclipse.jface.text.BadLocationException
import org.eclipse.jface.text.IDocument
import org.eclipse.jface.text.codemining.ICodeMining
import org.eclipse.xtext.resource.XtextResource
import org.eclipse.xtext.util.CancelIndicator
import org.eclipse.xtext.util.IAcceptor
import org.xtext.example.mydsl1.myDsl.Model
import org.eclipse.xtext.nodemodel.util.NodeModelUtils
[...]

PR #856, opened 2018-10-11
[#836] Provide better fix for CodeMining functionality bug on Photon.

miklossy:
- Set the lenght of a Position object to '1' instead of the lenght of
the content text.


cdietrich:
thx @miklossy

PR #868, opened 2018-10-20
[eclipse/xtext-core#903] Build against 2018-12

cdietrich:
[eclipse/xtext-core#903] Build against 2018-12


kthoms:
Shall we really make this intermediate step in the example? The snapshots will have to be removed anyway for the final release.

cdietrich:
well how else can we test against the milestones with all the equinox.ds removal the platform team did

kthoms:
For testing this can be done manually, couldn't it? But if you prefer 2 commits here then let it be.

cdietrich:
Can you do testing on a regular basis? I have the feeling I am the only one that does

Issue #281, opened 2017-06-08
Bogus behaviour of mixing xtext and non xtext projects


cdietrich:
assume you have a java project a (no xtext nature)
and a java project b with xtext nature
b has dependency to a

if you change something in a, b will not rebuild the xtend stuff and so
i see only the errors copied from java to xtend

and if there are unstructural changes from non xtext projects in the queue,
xtext will allways tell the builder infrastructure "needsRebuild"
which then will call all builders up to 10 times

cdietrich:
see https://github.com/eclipse/xtext/issues/2429 as well

cdietrich:
reason is that the "toBeBuilt" is empty. `org.eclipse.xtext.builder.impl.XtextBuilder.incrementalBuild(IResourceDelta, IProgressMonitor)`

cdietrich:
caused by `org.eclipse.xtext.builder.impl.javasupport.JdtQueuedBuildData.doNeedRebuild(JavaBuilderState, (UnconfirmedStructuralChangesDelta)=>void)` having a project guard.

szarnekow:
Fixed

PR #872, opened 2018-10-31
[540613] Use a scheduling rule for Xtext projects only

kthoms:
Allows other projects to be built in parallel. Before the change the
XtextBuilder did use a workspace scheduling rule which prevents other
projects to build.

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
@@ -9,6 +9,7 @@ package org.eclipse.xtext.builder.impl;
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
 
@@ -18,6 +19,7 @@ import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.IResourceDeltaVisitor;
 import org.eclipse.core.resources.IStorage;
+import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -25,8 +27,10 @@ import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.ProgressMonitorWrapper;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.jobs.IJobChangeEvent;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.core.runtime.jobs.JobChangeAdapter;
+import org.eclipse.core.runtime.jobs.MultiRule;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.ResourceSet;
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=540613

cdietrich:
i wonder how to test that. if i understand correct we have to set something > 1 here
bildschirmfoto vom 2018-10-31 08-51-12

cdietrich:
=> i would prefer an additional option e.g. system property to opt out for that. i dont see how we with our current kappa can reasonably test this for unwanted side effects in the next 4 weeks so i want to have a fallback

cdietrich:
see e.g. https://github.com/eclipse/xtext-eclipse/issues/281

cdietrich:
https://github.com/eclipse/reddeer/wiki/Set-plugin-preference-prior-Eclipse-start

mduft:
It is not easy to test at all. Currently Eclipse is VERY defensive with enabling parallel builds. Even when setting a max parallel build count > 1 it will require a patched JavaBuilder (which currently as well uses a not good scheduling rule). Actually, the PDE DSBuilder also requires patching right now (change merged for 4.10 already though). Then there might be other builders in your workspace (which again might use sub-optimal rules). Having a single builder with a null or workspace root rule in the workspace will basically disable parallel builds.

mduft:
If you think that even the defensive approach chosen now is too hardcore, I'd go with an opt-in eclipse preference. It can be set by the brave, and all others are not affected.

Still I think the approach is defensive enough. Even PDE can cope with that, and its definitely not the best code i've seen so far xD

kthoms:
It would be great to have a preference option, e.g. "Locking strategy - Workspace (default), Xtext Projects, Project). WDYT?

cdietrich:
am not suire if i want to have a preference or a system property

kthoms:
A preference (e.g. Experimental features) would allow users to test better around with that and choose also the most aggressive project rule when it does not harm. When we know that it is stable enough the preference can disappear in one of the next releases again.
A system property would require too deep knowledge for users to test the feature.

cdietrich:
well for a preference we need a place where to set it in ui which we currently dont have

cdietrich:
as i said: what if user does not install xtext.xtext.ui

kthoms:
I would rather place that in Shared UI.

cdietrich:
no i mean "Xtext" is the menu point for the DSL called `Xtext`

mduft:
the actual preference value should be stored with the builder (so you can set it manually from an .epf file even without UI - we do this all the time :)). The two "best" places for a dedicated preference page setting that preference would be IMHO either

1) General > Workspace > Build > Xtext
2) Xtext > Builder

I would not look for such an option anywhere else TBH.

cdietrich:
and xtext.ui.share is no ui component. we should make sure it still works headless

cdietrich:
as discussed:

- introduce preference to build only xtext projects non parallel and/or all projects in parallel
- add tycho config from my branch + enable xtext opt in there
- adapt oomph to activate both parallel builds and build non xtext projects in parallel
- after some tests with nightlys introduce gui for pref

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -374,4 +381,11 @@ protected void doClean(ToBeBuilt toBeBuilt, IProgressMonitor monitor) throws Cor
 		}
 	}
 	
+	@Override
+	public ISchedulingRule getRule(int kind, Map<String, String> args) {
+		return new MultiRule(Arrays.stream(
```
szarnekow: I'm not terribly sure about the injected workspace. `getProject().getWorkspace()` should be sufficient

mduft: using workspace (root) as a rule will block all other builders (e.g. JDT, PDE, ...) from using the "new" parallel build capabilities as soon as there is a single Xtext project.

szarnekow: To be more explicit:
Instead of `workspace.getRoot().getProjects().etc` I'd rather opt for `getProject().getWorkspace().getRoot().getProjects().etc`

kthoms:
Replaced by #893

PR #876, opened 2018-11-05
https://github.com/eclipse/xtext-xtend/issues/372 Refactoring.

miklossy:
- Eliminate duplicated code in the Javadoc hovering/hyperlinking
functionality by introducing the JavaDocHelper utility class.

### DIFF
--- a/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/hover/XbaseDispatchingEObjectTextHover.java
+++ b/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/hover/XbaseDispatchingEObjectTextHover.java
@@ -7,8 +7,6 @@
  *******************************************************************************/
 package org.eclipse.xtext.xbase.ui.hover;
 
-import java.util.List;
-
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.jface.text.IRegion;
 import org.eclipse.jface.text.ITextViewer;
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/util/JavaDocHelper.xtend
```diff
@@ -0,0 +1,69 @@
...
+	@Inject extension IScopeProvider
+
+	@Inject extension IJavaDocTypeReferenceProvider
+
+	@Inject extension IQualifiedNameConverter
+
+	/**
+	 * Returns the referenced elements in javadoc comment at a given <i>offset</i>.
+	 */
+	def Pair<EObject, IRegion> getReferencedElementInJavaDoc(XtextResource resource, int offset) {
```
szarnekow: If we introduce a new abstraction here with the JavaDocHelper, I'd like it to return a "better" type than a pair. Even if it is only a small data object that has an EObject and a region, it would make sense to encapsulate and name that better than `Pair<EObject, IRegion>`.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/util/JavaDocHelper.xtend
```diff
@@ -0,0 +1,69 @@
...
+		val parseResult = resource.parseResult
+		if(parseResult !== null) {
+			val rootNode = parseResult.rootNode
+			val node = rootNode.findLeafNodeAtOffset(offset)
+			val semanticObject = node.findActualSemanticObjectFor
+			if(semanticObject !== null) {
+				val scope = semanticObject.getScope(TypesPackage.Literals.JVM_PARAMETERIZED_TYPE_REFERENCE__TYPE)
+				val replaceRegions = node.computeTypeRefRegions
+				for(replaceRegion : replaceRegions) {
```
szarnekow: Poor choice of local variable names

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/util/JavaDocHelper.xtend
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @author miklossy - Initial contribution and API
+ * 
+ * @since 2.16
+ */
+class JavaDocHelper {
+
+	@Inject extension IScopeProvider
+
+	@Inject extension IJavaDocTypeReferenceProvider
```
szarnekow: Judging from the names of the types, the method `getReferencedElementInJavaDoc` belongs to the IJavaDocTypeReferenceProvider - which would deserve a better name, too since it does not know anything about types at a first glance.

@miklossy @cdietrich 
What do you think about a new default method on `IJavaDocTypeReferenceProvider` that returns something like `class EObjectInComment { EObject object; ITextRegion region }` and returns nothing in its default impl (in the default method of the interface). The MultiLineComment thingy would have a template method `getEReference():EReference` that returns null and has the bigger part of the logic as it is implemented in this new helper right now. We could have a special binding for Xbase that is aware of `TypesPackage.Literals.JVM_PARAMETERIZED_TYPE_REFERENCE__TYPE` and than we would not need a new helper class and spread the handling of comments accross different abstractions.

miklossy:
I like the idea of @szarnekow , created a PR in xtext-core: https://github.com/eclipse/xtext-core/pull/948

miklossy:
@cdietrich you are absolutly right. I will update this PR as soon as https://github.com/eclipse/xtext-extras/pull/329 is merged.

miklossy:
PR updated as requested to reuse functionality provided by the IJavaDocTypeReferenceProvider.

cdietrich:
thx @miklossy

Issue #839, opened 2018-10-03
Provide ui test infrastructure for testing Quickfixes.


miklossy:
It would be useful to extend the Xtext ui test infrastructure to enable the comfortable testing of Quickfixes.
See also https://www.eclipse.org/forums/index.php/t/1066662/

kthoms:
New class AbstractQuickfixTest is introduced

miklossy:
Adding the following QuickfixTest test case to the DomainModel Example
```
	@Test def fix_unused_imports2() {
		'''
			import java.util.Date
			import java.util.List
			
			entity Blog {
				posts : List<Post>
			}
			
			entity Post {}
		'''.testQuickfixesOn(IMPORT_UNUSED,
			new Quickfix(
				"Organize imports",
				"Organizes the whole import section. Removes wildcard imports as well as duplicates and unused ones.",
				'''
					import java.util.List
					
					entity Blog {
						posts : List<Post>
					}
					
					entity Post {}
				'''
			)
		)
	}
```
and execute it results in a test failure.
![screenshot](https://user-images.githubusercontent.com/20393472/48271692-2c2bb500-e43d-11e8-8ee9-e4f44c77e373.png)

There might be a problem in the [AbstractQuickfixTest](https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractQuickfixTest.xtend) class.

PR #891, opened 2018-11-09
[#839] Ensure the Xbase organize import quickfix can be properly tested.

miklossy:
- Modify the AbstractQuickfixTest to use the IResourceSetProvider to
obtain a new resource while testing the IssueModification effect.
- Implement corresponding QuickfixTest DomainModel Example test cases.

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractQuickfixTest.xtend
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractQuickfixTest.xtend
@@ -10,6 +10,7 @@ package org.eclipse.xtext.ui.testing
 import com.google.inject.Inject
 import com.google.inject.Injector
 import org.eclipse.core.resources.IFile
+import org.eclipse.core.resources.IProject
 import org.eclipse.core.runtime.NullProgressMonitor
 import org.eclipse.emf.common.util.URI
 import org.eclipse.emf.ecore.util.EcoreUtil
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractQuickfixTest.xtend
```diff
@@ -194,8 +198,7 @@ abstract class AbstractQuickfixTest extends AbstractEditorTest {
 		val in = new StringInputStream(Strings.emptyIfNull(model))
 		val uri = URI.createURI("") // creating an in-memory EMF Resource
 		
-		val rs = injector.getInstance(XtextResourceSet)
-		rs.setClasspathURIContext(AbstractQuickfixTest)
+		val rs = project.get
```
cdietrich: would prefer a non extension syntax here for better readability

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractQuickfixTest.xtend
```diff
@@ -194,8 +198,7 @@ abstract class AbstractQuickfixTest extends AbstractEditorTest {
 		val in = new StringInputStream(Strings.emptyIfNull(model))
 		val uri = URI.createURI("") // creating an in-memory EMF Resource
 		
-		val rs = injector.getInstance(XtextResourceSet)
-		rs.setClasspathURIContext(AbstractQuickfixTest)
```
cdietrich: this is imho a api/behaviour change. am not sure if we want that?

cdietrich: @szarnekow what do you think?

cdietrich: but i am not sure if the current impl makes any sense when project and rs dont fit together

miklossy:
Updated the PR not to use extension syntax for the IResourceSetProvider member. I think we do not need to care about api/behaviour change, since this class has not been released yet (is going to be released with 2.16 for the first time).

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractQuickfixTest.xtend
szarnekow: Since we now get the resourceSet associated with the project, it will already have the correct classpathURIContext. Looks about right to me.

PR #893, opened 2018-11-12
[540613] Enable parallel building

kthoms:
Makes the scheduling rule of the XtextBuilder configurable. By default
the workspace rule is used, which prevents parallel building of other
projects. This here introduces two additional more relaxed strategies
that enable parallel building of other projects:
- all_xtext_projects: Defines a scheduling rule for all projects woth
Xtext nature
- project: Defines a scheduling rule for the current built project
Enable parallel building in the tests by pluginCustomization and
activate the all_xtext_projects scheduling rule.

Enable all_xtext_projects for test execution by plugin_customization.ini
and in Xtext.setup.

Also-by: Karsten Thoms <karsten.thoms@itemis.de>


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=540613

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
```diff
@@ -32,6 +32,36 @@
 	 * Name of a preference for configuring whether the builder participant is enabled or not.
 	 */
 	public static final String PREF_AUTO_BUILDING = "autobuilding"; //$NON-NLS-1$
+	/**
+	 * Name of a preference defining the scheduling rule strategy for the builder.
+	 * @since 2.16
+	 */
+	public static final String PREF_SCHEDULING_RULE = "schedulingrule"; //$NON-NLS-1$
```
cdietrich: can you add a `@Beta`

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
```diff
@@ -32,6 +32,36 @@
...
+			this.name = name;
+		}
+		
+		static SchedulingOption fromName (String name) {
+			for (SchedulingOption opt: values()) {
+				if (opt.name.equals(name)) {
+					return opt;
+				}
+			}
+			throw new IllegalArgumentException(name);
```
cdietrich: log a warning and default?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
```diff
@@ -56,6 +86,7 @@ public void initialize(IPreferenceStoreAccess preferenceStoreAccess) {
 
 		protected void initializeBuilderPreferences(IPreferenceStore store) {
 			store.setDefault(PREF_AUTO_BUILDING, true);
+			store.setDefault(PREF_SCHEDULING_RULE, SchedulingOption.WORKSPACE.name);
```
cdietrich: am not sure if this will be affected by https://bugs.eclipse.org/bugs/show_bug.cgi?id=449010

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: Would be a really bad programming error if you'd try to get a value for a non existing name

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
```diff
@@ -109,4 +140,22 @@ public void setAutoBuildEnabled(Object context, boolean enabled) {
...
+		if (preferenceStore.contains(PREF_SCHEDULING_RULE)) {
+			return SchedulingOption.fromName(preferenceStore.getString(PREF_SCHEDULING_RULE));
+		}
+		return SchedulingOption.WORKSPACE;
+	}
+	
+	/**
+	 * @since 2.16
+	 */
+	public void setSchedulingRule (Object context, SchedulingOption opt) {
```
cdietrich: what is this needed for?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: OK

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: If you'd like to set this preference programmatically. I'd like to have that symetric. Option 'autoBuildEnabled' is also settable.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -82,6 +86,9 @@
 	@Inject 
 	private OperationCanceledManager operationCanceledManager;
 	
+	@Inject
+	private BuilderPreferenceAccess preferenceAccess;
```
cdietrich: @szarnekow isnt this a language specific thing?

cdietrich: 1) No implementation for java.lang.String annotated with @com.google.inject.name.Named(value=languageName) was bound.
  while locating java.lang.String annotated with @com.google.inject.name.Named(value=languageName)
    for parameter 0 at org.eclipse.xtext.ui.editor.preferences.PreferenceStoreAccessImpl.setLanguageNameAsQualifier(Unknown Source)
  at org.eclipse.xtext.ui.editor.preferences.PreferenceStoreAccessImpl.setLanguageNameAsQualifier(Unknown Source)
  while locating org.eclipse.xtext.ui.editor.preferences.IPreferenceStoreAccess
    for parameter 0 at org.eclipse.xtext.builder.preferences.BuilderPreferenceAccess.setPreferenceStoreAccess(Unknown Source)
  at org.eclipse.xtext.builder.preferences.BuilderPreferenceAccess.setPreferenceStoreAccess(Unknown Source)

cdietrich: => we should not use preference access for this?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
cdietrich: see other place. this is a global and not a language specific option

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
szarnekow: True. This should use the standard APIs to access the preference.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -373,5 +380,17 @@ protected void doClean(ToBeBuilt toBeBuilt, IProgressMonitor monitor) throws Cor
 			progress.worked(1);
 		}
 	}
-	
+
+	@Override
+	public ISchedulingRule getRule(int kind, Map<String, String> args) {
+		switch (preferenceAccess.getSchedulingRule(getProject())) {
+			case WORKSPACE: return getProject().getWorkspace().getRoot();
```
szarnekow: Super experimental would be the 'null' rule.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
szarnekow: Why not `SchedulingOption.valueOf(..)`?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
```diff
@@ -32,6 +33,37 @@
...
+	 * Options for scheduling rules.
+	 * @since 2.16
+	 */
+	public static enum SchedulingOption {
+		WORKSPACE ("workspace"),
+		// all Xtext projects in the workspace
+		XTEXT_PROJECTS ("all_xtext_projects"), //$NON-NLS-1$
+		// current built project
+		PROJECT ("project");
+		private String name;
```
szarnekow: Enums do already have a name. We should not overload that.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
szarnekow: I'd go for the `SchedulingOption.name()`

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
```diff
@@ -109,4 +142,22 @@ public void setAutoBuildEnabled(Object context, boolean enabled) {
 		preferenceStore.setValue(PREF_AUTO_BUILDING, enabled);
 	}
 
+	/**
+	 * @since 2.16
+	 */
+	public SchedulingOption getSchedulingRule (Object context) {
```
szarnekow: This should not be an option that can be set on a project level.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
szarnekow: See above. From my understanding, this should be a workspace option and not per project.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ */
+public class ParallelBuildEnabledTest {
+
+	@Test
+	public void testIsParallelBuildEnabledInWorkspace() {
+		int maxConcurrentProjectBuilds = DefaultScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES).getInt("maxConcurrentBuilds", -1);
+		assertEquals("parallel build is not enabeld", 8, maxConcurrentProjectBuilds);
```
tivervac: enabled*

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
```diff
@@ -32,6 +33,37 @@
 	 * Name of a preference for configuring whether the builder participant is enabled or not.
 	 */
 	public static final String PREF_AUTO_BUILDING = "autobuilding"; //$NON-NLS-1$
+	/**
+	 * Name of a preference defining the scheduling rule strategy for the builder.
+	 * @since 2.16
```
tivervac: Nitpicking here, shouldn't the description and `@since` be separated by a newline?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
```diff
@@ -32,6 +33,37 @@
...
+	public static final String PREF_SCHEDULING_RULE = "schedulingrule"; //$NON-NLS-1$
+	/**
+	 * Options for scheduling rules.
+	 * @since 2.16
+	 */
+	public static enum SchedulingOption {
+		WORKSPACE ("workspace"),
+		// all Xtext projects in the workspace
+		XTEXT_PROJECTS ("all_xtext_projects"), //$NON-NLS-1$
+		// current built project
```
tivervac: Built = past tense, what exactly does "current built project" mean?
I think what you meant to say was "The currently building project"

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
kthoms: "was not enabled" might express it better

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: correct, we usually do this for methods

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: This is indeed better

kthoms:
Can't work on this further until thursday. If someone wants to update the branch in the meantime feel free to do so.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Test;
+
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ */
+public class ParallelBuildEnabledTest {
+
+	@Test
+	public void testIsParallelBuildEnabledInWorkspace() {
+		int maxConcurrentProjectBuilds = DefaultScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES).getInt("maxConcurrentBuilds", -1);
```
cdietrich: Should be instance scope if I think over it

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
kthoms: Why not? When we are already at it, it makes sense to add this.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: Agreed. I wanted to have the values lower case and the enum literal upper case. But it is OK to have the value upper case I guess.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: > am not sure if this will be affected by https://bugs.eclipse.org/bugs/show_bug.cgi?id=449010

Was not aware of that. Good point. For testing we should avoid the initialization then.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: Agreed.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
kthoms: Will remove this from BuilderPreferenceAccess and use IPreferenceStoreAccess in the XtextBuilder instead.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
kthoms: The value is defined in pluginCustomization.ini. There is no way to use a constant. The test here tests if the pluginCustomization.ini has been picked up successfully.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/BuilderPreferenceAccess.java
kthoms: Using valueOf() now. I had a mix of upper/lower case before, that was the reason. Comment is obsolete now.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
cdietrich: public static final String PREF_MAX_CONCURRENT_BUILDS = "maxConcurrentBuilds"; //$NON-NLS-1$

am not sure if this is available in oxygen already

cdietrich: an its intended to have a arbitrary number for our tests here

kthoms: I'd expect this to be ignored then. The test will fail then.
We should move to JUnit 5. This will allow us to exclude such tests depending on environment.

kthoms: > an its intended to have a arbitrary number for our tests here

I don't get this comment. What do you mean?

cdietrich: for reading it works anyway. i mean if the contant to be used in the test is there

cdietrich: well i randonly chose 8 inside the pluginCustomization ini. could be 2,6,5 and 7 as well

kthoms:
Please review again. The preference is now picked up in the test and tests are green.
https://ci.eclipse.org/xtext/job/xtext-eclipse/job/PR-893/5/testReport/org.eclipse.xtext.builder.tests/ParallelBuildEnabledTest/

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
cdietrich: still holds true

kthoms: sorry, missed that. updated.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -57,7 +65,31 @@
...
+		WORKSPACE,
+		// all Xtext projects in the workspace
+		ALL_XTEXT_PROJECTS,
+		// the currently building project
+		PROJECT,
+		// null scheduling rule
+		NULL;
+	}
+	
+	private IPreferenceStore preferenceStore = new ScopedPreferenceStore(InstanceScope.INSTANCE,
```
cdietrich: same here. totally unsure how to access correctly

kthoms: but this should access the InstanceScope here, shouldn't it?

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.Test;
+
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ */
+public class ParallelBuildEnabledTest {
+
+	@Test
+	public void testIsParallelBuildEnabledInWorkspace() {
+		int maxConcurrentProjectBuilds = InstanceScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES).getInt("maxConcurrentBuilds", -1);
```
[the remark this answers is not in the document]
kthoms: not on oxygen, but this test would fail then also? We might enable it only when org.eclipse.core.resources is at least 3.13. WDYT?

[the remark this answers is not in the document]
cdietrich: so simply add comment that hints to the constant

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
cdietrich: i have no idea how to correctly obtain a thing to read here. find only outdated or you can do 1000 ways documentation

kthoms: ScopedPreferenceStore is used quite common. E.g. 
org.eclipse.ui.internal.ide.ChooseWorkspaceData.readPersistedData()
org.eclipse.jdt.internal.ui.preferences.JavaBasePreferencePage.JavaBasePreferencePage()

cdietrich: yes org.eclipse.ui.plugin.AbstractUIPlugin.getPreferenceStore()

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.junit.Test;
+import org.osgi.framework.Version;
+
+/**
+ * @author Christian Dietrich - Initial contribution and API
+ */
+public class ParallelBuildEnabledTest {
+
+	@Test
+	public void testIsParallelBuildEnabledInWorkspace() {
```
cdietrich: shouldnt this be

```
public class ParallelBuildEnabledTest {

	@Test
	public void testIsParallelBuildEnabledInWorkspace() {
		// preference and constant ResourcesPlugin#PREF_MAX_CONCURRENT_BUILDS only available on o.e.core.resources >= 3.13 
		if (Platform.getBundle(ResourcesPlugin.PI_RESOURCES).getVersion().compareTo(new Version(3,13,0)) < 0) {
			return; // running on too old platform
		}
		int maxConcurrentProjectBuilds = new ScopedPreferenceStore(InstanceScope.INSTANCE, ResourcesPlugin.PI_RESOURCES).getInt("maxConcurrentBuilds");
		assertEquals("parallel build was not enabled", 8, maxConcurrentProjectBuilds);
	}
	
	@Test
	public void testBuilderSchedulingRuleEnabledInWorkspace() {
		String schedulingRule = new ScopedPreferenceStore(InstanceScope.INSTANCE, Activator.PLUGIN_ID).getString(XtextBuilder.PREF_SCHEDULING_RULE);
		assertEquals("non-default scheduling rule was not enabled", SchedulingOption.ALL_XTEXT_PROJECTS.name(), schedulingRule);
	}
	
}
```

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -373,5 +405,29 @@ protected void doClean(ToBeBuilt toBeBuilt, IProgressMonitor monitor) throws Cor
...
+
+	/**
+	 * @since 2.16
+	 */
+	private SchedulingOption getSchedulingOption () {
+		String schedulingRuleName = preferenceStore.getString(PREF_SCHEDULING_RULE);
+		if (schedulingRuleName.isEmpty()) {
+			return SchedulingOption.WORKSPACE;
+		}
+		return SchedulingOption.valueOf(preferenceStore.getString(PREF_SCHEDULING_RULE));
```
cdietrich: nitpicky: why not use schedulingRuleName

kthoms: Is it really just a name? It is more a strategy.

cdietrich: got me wrong. why calling  SchedulingOption.valueOf(preferenceStore.getString(PREF_SCHEDULING_RULE)); twice

kthoms: OK, makes sense. I thought you were referring to the preference's name. Will update.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/tests/ParallelBuildEnabledTest.java
cdietrich: updated the pr according to my proposal

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
cdietrich: updated the pr

cdietrich:
@kthoms @szarnekow from my pov this PR is ok now. what do you think.
and: should i do the plugincustomizationn thingy for xtend as well?

kthoms:
The change in Xtend was already prepared. I opened the PR now: https://github.com/eclipse/xtext-xtend/pull/635

From my side OK now, we should start testing this in our devenvs now properly and add the pref page maybe when fine with another PR.

@szarnekow Please approve when OK for you.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -373,5 +405,29 @@ protected void doClean(ToBeBuilt toBeBuilt, IProgressMonitor monitor) throws Cor
...
+				.toArray(ISchedulingRule[]::new));
+			default: throw new IllegalArgumentException();
+		}
+	}
+
+	/**
+	 * @since 2.16
+	 */
+	private SchedulingOption getSchedulingOption () {
+		String schedulingRuleName = preferenceStore.getString(PREF_SCHEDULING_RULE);
```
szarnekow: Does this always do i/o or is this using cached information?

kthoms: I assume this is uncached, would have to test that. Do you think it would be worth to store the value and listen to changes? Would it make sense to enable IPreferenceStoreAccess to be used for such scenarios? (Currently it can only be used in project context)

kthoms: When we would like to enable change by a pref page, we have to react on changes.

cdietrich: Yes we have preference change listeners all over the place

kthoms: Ok, then I‘ll add a field that updates on change

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -86,6 +93,47 @@ public IResourceSetProvider getResourceSetProvider() {
...
+		if (schedulingRuleName.isEmpty()) {
+			schedulingOption = SchedulingOption.WORKSPACE;
+		} else {
+			schedulingOption = SchedulingOption.valueOf(schedulingRuleName);
+		}
+		
+		preferenceStore.addPropertyChangeListener(e -> {
+			if (PREF_SCHEDULING_RULE.equals(e.getProperty())) {
+				SchedulingOption.valueOf(e.getNewValue().toString());
```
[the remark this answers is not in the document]
kthoms: oh dear, of course :/

releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -363,6 +375,8 @@
         url="${p2.xtext}"/>
     <repository
         url="${p2.mwe2}"/>
+    <repository
```
cdietrich: dont forget to change back again

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -86,6 +93,47 @@ public IResourceSetProvider getResourceSetProvider() {
...
+		// the currently building project
+		PROJECT,
+		// null scheduling rule
+		NULL;
+	}
+	
+	private SchedulingOption schedulingOption;
+	
+	public XtextBuilder () {
+		IPreferenceStore preferenceStore = new ScopedPreferenceStore(InstanceScope.INSTANCE,
```
cdietrich: am not sure of we should move this to a post constuct `@Inject` annotated method. @szarnekow what do you think?

szarnekow: My thoughts on this:
Right now there will be one XtextBuilder per project thus one listener per project. Maybe we should move this preference handling to another class that is annotated with `@Singleton`. This would be injected anyway. Also it would be easier to test if the cached preference is updated after setting a value on the store.

cdietrich: :+1:

kthoms: Good idea. I will implement this.

kthoms: Updated

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -85,6 +93,58 @@
...
 		return resourceSetProvider;
 	}
+
+	/**
+	 * 
+	 * @since 2.16
+	 */
+	@Singleton
+	@Beta
+	static class BuilderPreferences {
```
szarnekow: Please make this public (but still `@Beta`) so others can customize it

releng/org.eclipse.xtext.contributor/Xtext.setup
szarnekow: See here

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
kthoms: done

kthoms:
@cdietrich Also OK for you now?

cdietrich:
build is red. do we need a rebase?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -85,6 +93,58 @@
...
+	@Inject
+	private BuilderPreferences preferences;
+	
+	/**
+	 * Options for scheduling rules.
+	 * 
+	 * @since 2.16
+	 */
+	public static enum SchedulingOption {
+		WORKSPACE,
```
cdietrich: do we want proper javadoc for these?

kthoms: Not sure if we keep this configurable in the future. Assumption is that one scheduling rule is enough. We just don't know yet the best strategy.
Or do you mean just convert the single-line comments to Javadoc?

cdietrich: yes i mean have the enum values properly documented

PR #919, opened 2018-12-04
#755 Fix failing UI tests in domain model example.

ArneDeutsch:
- Use IndentationAwareTemplateTranslator even for empty document
- Fix Use IndentationAwareTemplateTranslator such that line ending from
system is used (instead of line ending taken from template).

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/templates/XtextTemplateContext.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/templates/XtextTemplateContext.java
@@ -9,6 +9,7 @@ package org.eclipse.xtext.ui.editor.templates;
 
 import org.eclipse.jface.text.BadLocationException;
 import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.IDocumentExtension4;
 import org.eclipse.jface.text.IRegion;
 import org.eclipse.jface.text.Position;
 import org.eclipse.jface.text.templates.DocumentTemplateContext;
@@ -21,9 +22,8 @@ import org.eclipse.xtext.scoping.IScopeProvider;
 import org.eclipse.xtext.ui.editor.contentassist.ContentAssistContext;
 
 /**
- * Represents an extended version of class {@link DocumentTemplateContext} to provide additional Xtext related
- * information and services for resolving a <code>Template</code>. Furthermore it fixes the indentation
- * of the applied template.
+ * Represents an extended version of class {@link DocumentTemplateContext} to provide additional Xtext related information and services for
+ * resolving a <code>Template</code>. Furthermore it fixes the indentation of the applied template.
  * 
  * @author Michael Clay - Initial contribution and API
  * @author Sebastian Zarnekow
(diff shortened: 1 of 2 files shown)


ArneDeutsch:
The test fail because they check, that if the templates are translated the line breaks are used that are defined in the templates. I have changed the logic such that the line seperator from the system is used instead. My fix would be to relax the assertion in the test. @szarnekow would you please have a look at the class `XtextTemplateContextTest` and verify that you agree if it is a good idea to use the system line seperator instead the one defined in the template?

ArneDeutsch:
Developed an even better solution. Will try to read the first newline delimiter in the document, use that if present and only use system delimiter if no delimiter zert existent. Should work best this way. Will comit in short time.

ArneDeutsch:
Please review again.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/templates/XtextTemplateContext.java
```diff
@@ -88,37 +87,40 @@ protected TemplateTranslator createTemplateTranslator() {
 			int offset = getStart();
 			IRegion lineRegion = getDocument().getLineInformationOfOffset(offset);
 			String line = getDocument().get(lineRegion.getOffset(), lineRegion.getLength());
+			String lineDelimiter = getDocument().getLineDelimiter(0);
```
szarnekow: There is an extension interface IDocumentExtension4 that offers `getDefaultLineDelimiter`. That's probably to be preferred.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/templates/XtextTemplateContext.java
```diff
@@ -88,37 +87,40 @@ protected TemplateTranslator createTemplateTranslator() {
...
 
 		@Override
 		public TemplateBuffer translate(Template template) throws TemplateException {
 			return translate(template.getPattern());
 		}
-		
+
 		@Override
 		public TemplateBuffer translate(String string) throws TemplateException {
 			String withIndentation = string.replaceAll("(\r\n?)|(\n)", "$0" + indentation);
```
szarnekow: "$0" is the original line delimiter. Maybe you simply want to replace it by the line delimiter as passed into the constructor?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/templates/XtextTemplateContext.java
```diff
@@ -88,37 +87,40 @@ protected TemplateTranslator createTemplateTranslator() {
...
 	public static class IndentationAwareTemplateTranslator extends TemplateTranslator {
-	
+
 		private final String indentation;
+		private final String lineDelimiter;
 
-		public IndentationAwareTemplateTranslator(String indentation) {
+		public IndentationAwareTemplateTranslator(String indentation, String lineDelimiter) {
 			this.indentation = indentation;
+			this.lineDelimiter = lineDelimiter;
```
szarnekow: if the given line delimiter is null, you can use the System's default.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/templates/XtextTemplateContext.java
ArneDeutsch: Done.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/templates/XtextTemplateContext.java
ArneDeutsch: Done.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/templates/XtextTemplateContext.java
ArneDeutsch: Done.

PR #927, opened 2018-12-13
eclipse/xtext-xtend#655 Fix exception.

ArneDeutsch:
Fix exception in case a file has already organized imports.


ArneDeutsch:
This is only a fix as small part of the ticket. We should merge it right away, even if we do not implement the full feature rigt away.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/imports/MultiImportOrganizer.java
```diff
@@ -75,7 +75,7 @@
 					}
 					TextEdit textEdit = replaceConverter.convertToTextEdit(replace);
```
szarnekow: Why does it produce a null value?

tivervac: If replace.isEmpty then textEdit is null

tivervac: Don't think it matters a lot

ArneDeutsch: Adapted as suggested. Makes sense.

Issue #796, opened 2018-08-23
ResourceDescriptionUpdaterTest.testIndependentProjects is flaky


cdietrich:
ResourceDescriptionUpdaterTest.testIndependentProjects is flaky

```
Fehlermeldung

source

Stacktrace

java.lang.AssertionError: source
	at org.eclipse.xtext.builder.impl.ResourceDescriptionUpdaterTest.testIndependentProjects(ResourceDescriptionUpdaterTest.java:61)
```

http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/cd_core_issue822b/lastCompletedBuild/testReport/org.eclipse.xtext.builder.impl/ResourceDescriptionUpdaterTest/testIndependentProjects/
http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/cd_issue1266_asm62/6/testReport/junit/org.eclipse.xtext.builder.impl/ResourceDescriptionUpdaterTest/testIndependentProjects/

cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/master/161/testReport/junit/org.eclipse.xtext.builder.impl/ResourceDescriptionUpdaterTest/testIndependentProjects/

cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/master/223/testReport/junit/org.eclipse.xtext.builder.impl/ResourceDescriptionUpdaterTest/testIndependentProjects/

ArneDeutsch:
Guess it is because AbstractParticipatingBuilderTest is not thread safe. Will make it so and hope it will prevent the flakynes.

ArneDeutsch:
Class is now thread safe. Hopefully test is now stable.

PR #932, opened 2019-01-03
#796 Make AbstractParticipatingBuilderTest thread safe

ArneDeutsch:
Because multiple threads (main + build thread) may access state
concurrently.

### DIFF
--- a/org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/AbstractParticipatingBuilderTest.java
+++ b/org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/AbstractParticipatingBuilderTest.java
@@ -44,35 +44,35 @@ public abstract class AbstractParticipatingBuilderTest extends AbstractBuilderTe
 	}
 	
 	@Override
-	public void build(IBuildContext context, IProgressMonitor monitor) throws CoreException {
+	public synchronized void build(IBuildContext context, IProgressMonitor monitor) throws CoreException {
 		if (logging) {
 			invocationCount++;
 			this.context = context;
 		}
 	}
 	
-	public int getInvocationCount() {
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/AbstractParticipatingBuilderTest.java
```diff
@@ -44,35 +44,35 @@ public void tearDown() throws Exception {
 	}
 	
 	@Override
-	public void build(IBuildContext context, IProgressMonitor monitor) throws CoreException {
+	public synchronized void build(IBuildContext context, IProgressMonitor monitor) throws CoreException {
```
szarnekow: This is usually executed in a job / workspace lock. Why does it help to make it synchronized?

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/AbstractParticipatingBuilderTest.java
```diff
@@ -44,35 +44,35 @@ public void tearDown() throws Exception {
...
 		if (logging) {
 			invocationCount++;
 			this.context = context;
 		}
 	}
 	
-	public int getInvocationCount() {
+	public synchronized int getInvocationCount() {
 		return invocationCount;
```
szarnekow: What I can imagine is that invocationCount must be volatile.

cdietrich:
@szarnekow @ArneDeutsch what is the state of this pr

szarnekow:
> @szarnekow @ArneDeutsch what is the state of this pr

I'd like to understand the reasoning behind the changes.

szarnekow:
As discussed offline: Please go ahead with this.

PR #933, opened 2019-01-03
Xtext eclipse 813 automated ui tests

ArneDeutsch:


cdietrich:
this change makes no sense to me. i dont expect changes here

ArneDeutsch:
I think you are right. Seems these are all leftovers from the time when the tests were residing in the Eclipse project. Will close this request.

PR #935, opened 2019-01-08
[eclipse/xtext#1249] use bom for version management

szarnekow:
Target files must be reset before merge.


szarnekow:
> How about the Domainmodel example? We could use the BOM here, too.

Let's do this after this effort was merged to master. I'd like to keep the cosmetic changes to a bare minimum now.

szarnekow:
Does only add a comment. Merging now.

PR #936, opened 2019-01-09
[eclipse/xtext#1365] added support for 2019-03 to oomph

cdietrich:
[eclipse/xtext#1365] added support for 2019-03 to oomph


cdietrich:
@kthoms since i dont know how to handle oomph editors properly can you please have a closer look

PR #942, opened 2019-01-16
[https://github.com/eclipse/xtext-xtend/issues/671]

miklossy:
Add Quickfix to remove unused local variables.


org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/quickfix/XbaseQuickfixProvider.java
```diff
@@ -236,6 +238,20 @@ public void apply(EObject element, IModificationContext context) throws Exceptio
...
 		}
 	}
+	
+	@Fix(IssueCodes.UNUSED_LOCAL_VARIABLE)
+	public void removeUnusedLocalVariable(final Issue issue, IssueResolutionAcceptor acceptor) {
+		// use the same label, description and image
+		// to be able to use the quickfixes (issue resolution) in batch mode
+		String label = "Remove.";
+		String description = "Remove the unused local variable.";
+		String image = "fix_indent.gif";
```
holgerschill: Maybe use fix_local_var.gif instead.

holgerschill: We will use the remove icon from jdt instead.

miklossy: Thanks for the feedback, it is a good idea to use the same symbols as jdt uses. I updated the PR as requested.

PR #947, opened 2019-01-25
Improve ResourceSetBasedAllContainersStateProvider.getResourc…

AndersDala:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=427770

Short:
    Implement solution mentioned in above BUG report.

Longer:
    Bug (?) can be triggered by e.g.:

    IResourceServiceProvider sp = IResourceServiceProvider.Registry.INSTANCE.getResourceServiceProvider(fileUri);
    IResourceSetProvider provider = sp.get(IResourceSetProvider.class);
    IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
    // ... file etc
    IProject project = file.getProject();
    // Use the provider to get the resource set for the given project
    // FIXME how to get the existing XtextResourceSet? Below just creates a new rs. :-(
    ResourceSet rs = provider.get(project);
    // ... run build to generate index used by namespace etc
    BuildRequest request = new BuildRequest();
    request.setResourceSet(rs);
    request.setBaseDir(UriUtil.createFolderURI(new File(baseDir)));
    request.setDirtyFiles(allFiles);
    // ... copy paste code ...
    indexState = incrementalBuilder.build(request, languages).getIndexState();

    Hints for doing this correctly/cleaner is highly appreciated!

    After the code above has been triggered doing a Project Clean will
    trigger method from title to be called with a CurrentDescriptions.

Note that this commit goes together with a commit in xtext-core (see
https://github.com/eclipse/xtext-core/pull/1002).

Signed-off-by: Anders Dahlberg <anders.xb.dahlberg@ericsson.com>

### DIFF
--- a/org.eclipse.xtext.builder/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.builder/META-INF/MANIFEST.MF
@@ -43,7 +43,7 @@ Export-Package: org.eclipse.xtext.builder,
  org.eclipse.xtext.builder.trace;x-friends:="org.eclipse.xtext.xbase.ui,org.eclipse.xtend.ide",
  org.eclipse.xtext.builder.trace.impl;x-internal:=true,
  org.eclipse.xtext.builder.trace.util;x-internal:=true
-Require-Bundle: org.eclipse.xtext,
+Require-Bundle: org.eclipse.xtext;bundle-version="2.17.0",
  org.eclipse.xtext.util,
  org.eclipse.emf.ecore;bundle-version="2.10.2";visibility:=reexport,
  com.google.guava;bundle-version="[14.0.0,22.0.0)",
--- a/org.eclipse.xtext.builder/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.builder/META-INF/MANIFEST.MF
@@ -43,7 +43,7 @@ Export-Package: org.eclipse.xtext.builder,
  org.eclipse.xtext.builder.trace;x-friends:="org.eclipse.xtext.xbase.ui,org.eclipse.xtend.ide",
  org.eclipse.xtext.builder.trace.impl;x-internal:=true,
  org.eclipse.xtext.builder.trace.util;x-internal:=true
-Require-Bundle: org.eclipse.xtext;bundle-version="2.17.0",
+Require-Bundle: org.eclipse.xtext,
  org.eclipse.xtext.util,
  org.eclipse.emf.ecore;bundle-version="2.10.2";visibility:=reexport,
  com.google.guava;bundle-version="[14.0.0,22.0.0)",
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/clustering/CurrentDescriptions.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/clustering/CurrentDescriptions.java
@@ -22,6 +22,8 @@ import org.eclipse.xtext.resource.IResourceDescription;
 import org.eclipse.xtext.resource.IResourceDescriptions;
 import org.eclipse.xtext.resource.impl.ResourceDescriptionsData;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 /**
  * @author Sebastian Zarnekow - Initial contribution and API
  * @author Thomas Wolf <thomas.wolf@paranor.ch> - Performance optimization and JavaDoc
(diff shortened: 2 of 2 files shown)


AndersDala:
> When the bug can be reproduced with above code, could add a test to `org.eclipse.xtext.builder.tests`?

I will try!

org.eclipse.xtext.builder/META-INF/MANIFEST.MF
```diff
@@ -43,7 +43,7 @@ Export-Package: org.eclipse.xtext.builder,
  org.eclipse.xtext.builder.trace;x-friends:="org.eclipse.xtext.xbase.ui,org.eclipse.xtend.ide",
  org.eclipse.xtext.builder.trace.impl;x-internal:=true,
  org.eclipse.xtext.builder.trace.util;x-internal:=true
-Require-Bundle: org.eclipse.xtext,
+Require-Bundle: org.eclipse.xtext;bundle-version="2.17.0",
```
szarnekow: We usually do not maintain plugin versions.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/clustering/CurrentDescriptions.java
```diff
@@ -26,14 +26,16 @@
...
+public class CurrentDescriptions extends AdapterImpl implements IResourceDescriptions.IResourceSetAware {
 
 	/**
 	 * New index.
 	 */
 	private final ResourceDescriptionsData newData;
 	
 	private BuildData buildData;
+
+	private ResourceSet resourceSet;
```
szarnekow: I'm not terribly sure, but AdapterImpl.getTarget already yields the resource set. So calling that one and casting should be good enough.

AndersDala: Ok, I will change to using getTarget()! Should I add null-checks and/or cast-checks while doing that to make sure that getTarget() is not null or another type?

szarnekow: I think a simple cast but an explicit Preconditions.checkNotNull() would be good to have, yes. 👍

org.eclipse.xtext.builder/META-INF/MANIFEST.MF
cdietrich: @AndersDala please remove this

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/clustering/CurrentDescriptions.java
AndersDala: Will do, busy days ahead but will squeeze it in somewhere. :-)

org.eclipse.xtext.builder/META-INF/MANIFEST.MF
AndersDala: Will do!

cdietrich:
@szarnekow what do you think?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/clustering/CurrentDescriptions.java
```diff
@@ -138,6 +140,16 @@ public boolean isAdapterForType(Object type) {
...
+	@Override
+	public ResourceSet getResourceSet() {
+		Object target = getTarget();
+		ResourceSet resourceSet = (ResourceSet) checkNotNull(target);
+		return resourceSet;
+	}
+
 	/**
 	 * Context-aware instance of our index.
 	 */
```
szarnekow: This impl should probably also become IResourceSetAware

AndersDala: Not sure I understand, it already is, or?

szarnekow: Looks like I cannot add a comment to a line outside of the diff. I was refering to `public static class ResourceSetAware implements IResourceDescriptions.IContextAware` below

AndersDala: Ah ok, got it now I think. So have that implement IResourceSetAware and implement using similar approach as the existing setContext with the delegate etc?

cdietrich:
thx @AndersDala

PR #955, opened 2019-01-28
eclipse#953: Correct File Exist Error Message

jmenzies12:


kthoms:
Thx @jmenzies12 
Just for curiosity: You submitted the change by editing the file on GitHub only, right? I'm asking because usually this fails the Eclipse IP validation when users do this. They need to sign an Eclipse CLA and sign-off their commits. Your commit is signed-off and I assume you have a valid CLA, correct? What did you do to sign-off, just added this as comment, or did you actually check out the sources and did it there and pushed it again?

jmenzies12:
@kthoms 
Initially edited it on GitHub but it wouldn't pick up the sign off so had to check out the source locally and do it via the eclipse IDE. Which did then pick up the sign off.

I do have an eclipse account and have signed the ECA. Let me know if you need anything else

cdietrich:
so it looks like github only edits are still not possible

PR #957, opened 2019-01-30
Fix ClassCastException in Compare View.

AndersDala:
Bug can be reproduced by e.g. doing <right-click> --> Compare with HEAD
revision on an xtend file in a Git repo.
Or select one xtext document and a non-xtext document and do a compare.

Signed-off-by: Anders Dahlberg <anders.xb.dahlberg@ericsson.com>

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/syntaxcoloring/TokenScanner.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/syntaxcoloring/TokenScanner.java
@@ -141,8 +141,8 @@ public class TokenScanner extends AbstractTokenScanner {
 			// User might have selected a non-xtext document in e.g. a compare operation.
 			// Return an empty iterable, this will disable syntax highlighting
 			// for the "non-xtext editor".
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/syntaxcoloring/TokenScanner.java
```diff
@@ -134,6 +137,13 @@ public void setRange(IDocument document, final int offset, final int length) {
 	}
 
 	protected Iterable<ILexerTokenRegion> getTokens(IDocument document) {
+		if (!(document instanceof XtextDocument)) {
+			// User might have selected a non-xtext document in e.g. a compare operation.
+			// Return an empty iterable, this will disable syntax highlighting
+			// for the "non-xtext editor".
+			Iterator<ILexerTokenRegion> iterator = Collections.<ILexerTokenRegion>emptyIterator();
```
kthoms: `return Collections.emptyList()`

AndersDala: ...D'Oh. Thanks. :-)

PR #964, opened 2019-02-04
Made ClosedProjectsQueue more robust

szarnekow:


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/ClosedProjectsQueue.java
```diff
@@ -59,7 +59,12 @@ public boolean isEmpty() {
 		 * Add the tasks again to the top of the queue.
 		 */
 		public void reschedule() {
-			insert(projectNames, toBeBuilt);
+			Set<URI> toBeDeleted = toBeBuilt.getToBeDeleted();
```
cdietrich: is this executed single threaded

szarnekow: Yes, it's running in the builder with a lock.

PR #970, opened 2019-02-06
Refactoring the DomainmodelHyperlinkHelper in the Domain-Model Example.

miklossy:
- Use the inherited EObjectAtOffsetHelper instance instead of injecting
an own one.


cdietrich:
thx @miklossy

PR #971, opened 2019-02-07
made non vararg invocation clear

cdietrich:

### DIFF
--- a/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/imports/TypeChooser.java
+++ b/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/imports/TypeChooser.java
@@ -62,7 +62,7 @@ public class TypeChooser {
 		IStructuredContentProvider contentProvider = new ContentProvider();
 		Dialog dialog = new Dialog(shell, new LabelProvider(labelProvider), contentProvider);
 		dialog.setInput(candidateTypes);
-		dialog.setInitialSelections(new JvmDeclaredType[] { candidateTypes.get(0) });
+		dialog.setInitialSelections((Object[])new JvmDeclaredType[] { candidateTypes.get(0) });
 		int result = dialog.open();
 		if(originalSelection != null)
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/imports/TypeChooser.java
```diff
@@ -62,7 +62,7 @@ public JvmDeclaredType choose(final List<JvmDeclaredType> candidateTypes, Iterab
 		IStructuredContentProvider contentProvider = new ContentProvider();
 		Dialog dialog = new Dialog(shell, new LabelProvider(labelProvider), contentProvider);
 		dialog.setInput(candidateTypes);
-		dialog.setInitialSelections(new JvmDeclaredType[] { candidateTypes.get(0) });
+		dialog.setInitialSelections((Object[])new JvmDeclaredType[] { candidateTypes.get(0) });
```
cdietrich: @szarnekow is this correct to indicate we call the in later eclipse versions variadic param in a non variadic way?

szarnekow: Actually I'd think passing an array of JvmDeclaredType did already indicate this and would be source compatible to a varargs-change.

cdietrich: well jdt complains.
see your workspace ;)

szarnekow: e.g. `Arrays.asList(new String[] { "" });` this is a `List<String>` and not a `List<String[]>`

szarnekow: My example was broken since Arrays.asList uses a type parameter.

PR #973, opened 2019-02-08
Updated .classpath files

szarnekow:


org.eclipse.xtext.builder/.classpath
```diff
@@ -2,13 +2,9 @@
...
-	<classpathentry kind="src" path="xtend-gen"/>
-	<classpathentry kind="src" path="emf-gen">
-		<attributes>
-			<attribute name="ignore_optional_problems" value="true"/>
-		</attributes>
-	</classpathentry>
-	<classpathentry kind="src" path="deprecated"/>
-	<classpathentry kind="output" path="bin"/>
+	<classpathentry kind="src" path="emf-gen/"/>
+	<classpathentry kind="src" path="src/"/>
```
kthoms: why emf-gen before src?

szarnekow: Seems to depend on the order in the build.properties (which does make sense). I'll reorder it there.

kthoms: FYI The order can be defined in the build path and affects the order in which the containers are listed in the navigator tree.

szarnekow: See next commit

PR #985, opened 2019-02-20
Add null guards

kthoms:
Although happened in a misconfigured project it would be better to guard this instead of throwing NPE.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/XtextProjectHelper.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/XtextProjectHelper.java
@@ -12,6 +12,8 @@ import org.eclipse.core.resources.ICommand;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 
+import com.google.common.base.Preconditions;
+
 /**
  * @author Sebastian Zarnekow - Initial contribution and API
  */
@@ -23,6 +25,7 @@ public class XtextProjectHelper {
 	private static final Logger log = Logger.getLogger(XtextProjectHelper.class);
 
 	public static boolean hasNature(IProject project) {
+		Preconditions.checkNotNull(project);
 		try {
 			if (project.isAccessible()) {
 				return project.hasNature(NATURE_ID);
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/XtextProjectHelper.java
```diff
@@ -24,7 +24,7 @@
 
 	public static boolean hasNature(IProject project) {
```
szarnekow: Where did the project come from? E.g. which piece of code called hasNature with a null argument? Usually projects are obtained from the workspace and are not null.

kthoms: We are working on containers with external files. In that specific case Storage2UriMapperImpl returns a collection of mappings from a storage to a null project. This causes the NPE in org.eclipse.xtext.builder.impl.QueuedBuildData.queueURI(URI) when the helper is invoked with a null project.
As said, the project itself is buggy ATM, but the NPE here might not be the right answer also.

szarnekow: It depends. An exception helped you to easily spot that you called this method transitively with a null project. A simple false wouldn't have helped, I suppose. I'm not convinced.

kthoms: Then Preconditions#checkNotNull might worth consideration. It effectively also throws a NPE, but from the stack one could see that this is actually a guarded check.

szarnekow: Doesn't make any difference at all at this exact line of code where we dereference the argument and only the argument immediately and where we do not store the argument in a field. But please go ahead and do that.

cdietrich:
@kthoms @szarnekow what is the state of this pr?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/XtextProjectHelper.java
```diff
@@ -37,7 +40,7 @@ public static boolean hasNature(IProject project) {
 	 * @since 2.4
 	 */
 	public static boolean hasBuilder(IProject project) {
-		if (project.isAccessible()) {
+		if (project != null && project.isAccessible()) {
```
szarnekow: I don't like this here. We should bail out eagerly since it's most likely a caller problem.

kthoms: Updated to use `Preconditions.checkNotNull(project)`

Issue #986, opened 2019-02-20
BuilderDeltaConverter leaks JDT JavaModelCache.childrenCache 


sergio-otero:
I've been making performance optimizations on a pretty big project with almost 10.000 source files and after solving XText related problems (for example limiting scope), i'm stuck with an OutOfMemory from JDT itself when the generated java files from my DSL are compiled.

I've been narrowing down the problem and i've found 2 issues:

* OutOfMemory when the files appear due to an interaction with XText BuilderDeltaConverter 
* OutOfMemory when doing a full build because of JDT batch compilation of 2.000 files: this can be solved with -DmaxCompiledUnitsAtOnce=200

The first one only happens with XText plugins installed, even when the project is a simple java project with no xtext nature

Steps to reproduce:

* Open an Eclipse 2018-02 without XText Plugins. Use "-DmaxCompiledUnitsAtOnce=200" in eclipse.ini to avoid the other JDT problem
* Create a simple Java Project and paste the class GenCode below
* Execute GenCode to generate 10.000 files with 2.000 variables each
* Connect JVisualVM to Eclipse with VisualGC
* Refresh the project to let eclipse index the new files
* Look at OldGen occupation growing to 800Mb but going down to 400Mb after finishing (it may need a "perform gc" from JVisualVM)
* Then repeat the same experiment but with XText Plugins installed in Eclipse (happens at least in Eclipse Neon and 2018-12)
* In this case, memory goes up to several Gb and if it has enough to finish, it does not go down less than 2.4Gb. It happens event wild autobuild off
* Doing a heapdump shows that almost all memory is due to org.eclipse.jdt.internal.core.SourceFieldElementInfo and it's root path points to JavaModelManager->JavaModelCache->childrenCache (see attached jdt-dump.png)
* With the help of some breakpoints, i've seen that XText BuilderDeltaConverter somehow triggers that cache been filled when it's designed to hold only information about opened files. See the stacktrace at the bottom

My DSL does not need to link java types like xtend, but it doesn't seem posible to disable BuilderDeltaConverter looking at all the generated java code of some projects

Issue #70 is related, but in my case the java code is inside an Xtext project, so if BuilderDeltaConverter is changed to avoid looking at projects without XText Nature, it will still affect me.

I have a proposed fix in org.eclipse.xtext.common.types.ui.notification.DeltaConverter for convertNewTypes and convertChangedCompilationUnit

```
	/**
	 * @since 2.5
	 */
[...]

PR #987, opened 2019-02-21
[eclipse/xtext-eclipse#986] BuilderDeltaConverter JDT leak

sergio-otero:
Signed-off-by: Sergio Otero Lopez <sergio.otero@gmail.com>

### DIFF
--- a/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/notification/DeltaConverter.java
+++ b/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/notification/DeltaConverter.java
@@ -189,13 +189,15 @@ public class DeltaConverter {
 		
 		ICompilationUnit compilationUnit = (ICompilationUnit) delta.getElement();
 		try {
-			doForeachCompilationType(compilationUnit, type -> {
-				if (isDerived(type))
-					return;
+
+			boolean hasDerivedType = doForeachCompilationUnitChildType(compilationUnit, true, type -> {
 				String typeName = type.getFullyQualifiedName();
 				URI topLevelURI = uriHelper.createResourceURIForFQN(typeName);
 				convertChangedTypeAndChildren(result, typeNames, type, topLevelURI);
 			});
+			
+			if (hasDerivedType)
+				return;
 			
 			convertRemovedTypes(typeNames, result);
 		} catch (JavaModelException e) {
@@ -205,14 +207,19 @@ public class DeltaConverter {
 		}
 	}
 	
-	private void doForeachCompilationType(ICompilationUnit compilationUnit, Consumer<IType> consumer) throws JavaModelException {
+	private boolean doForeachCompilationUnitChildType(ICompilationUnit compilationUnit, boolean stopAtFirstDerivedType, Consumer<IType> consumer) throws JavaModelException {
 		IOpenable openableUnit = compilationUnit.getOpenable();
 		boolean wasOpen = openableUnit.isOpen();
 
 		try {
 			for (IType type : compilationUnit.getTypes()) {
-				consumer.accept(type);
+				if (stopAtFirstDerivedType && isDerived(type))
+					return true;
+				
+				consumer.accept(type); 
 			}
+			
+			return false;
 		} finally {
 			boolean isOpen = openableUnit.isOpen();
 			if (!wasOpen && isOpen) {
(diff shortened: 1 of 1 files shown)


sergio-otero:
I've signed the ECA after making the PR

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/notification/DeltaConverter.java
```diff
@@ -187,21 +189,38 @@ protected void convertChangedCompilationUnit(IJavaElementDelta delta, List<IReso
 		
 		ICompilationUnit compilationUnit = (ICompilationUnit) delta.getElement();
 		try {
-			for (IType type : compilationUnit.getTypes()) {
+			doForeachCompilationType(compilationUnit, type -> {
 				if (isDerived(type))
```
szarnekow: Here we do have a changed semantic: formally, we would return from the method if the typ was a derived type. With the change, only the lambda is left and the processing would continue for the remaining types in the compilation unit. Also the convertRemovedTypes below would kick in. Could you please sort this out?

sergio-otero: I think it's ok now

szarnekow:
@cdietrich Something for 2.17? What do you think?

cdietrich:
If you and Karsten find time to test

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/ui/notification/DeltaConverter.java
```diff
@@ -187,21 +189,46 @@ protected void convertChangedCompilationUnit(IJavaElementDelta delta, List<IReso
 		
 		ICompilationUnit compilationUnit = (ICompilationUnit) delta.getElement();
 		try {
-			for (IType type : compilationUnit.getTypes()) {
-				if (isDerived(type))
-					return;
+
+			boolean hasDerivedType = doForeachCompilationUnitChildType(compilationUnit, true, type -> {
```
szarnekow: Hmm I'm not terribly convinced yet. The common case would be, that we do not have 100s and 1000s of CU. Also I'd expect that the CU is accessed in subsequent operations again, e.g. during the Xtext build. Eager removal from the caches does not seem to be the right thing to do here. Should this be guarded against the number of CU in the delta? Or against the size of the existing cache.
A system property could mitigate these concerns. If the property was set, we'd do the eager cache cleanup. Otherwise not.

What do you think?

sergio-otero: I've been making some tests and it's quite complex to understand and as you say, not beneficial in all the cases

In a project with 2 java files, one referencing another, JDT does not put/remove anything in the cache, but is linking them appropriately
When a file with a reference to another is open, both end up in the cache, and after closing it, the referenced java is still in there (in can get removed if it's open and then closed)

In another test with an xtend file (GenTestXtend0) referencing a java class (GenTestJava0_0) is more complex, because they are added in different cases:
* GenTestXtend0.java: when the xtend file is compiled to java and is gets added by DeltaConverter
* GenTestJava0_0: when it appears and is detected by DeltaConverter (in my case, generated by a main file) and also when building the xtend scope of GenTestXtend0 (when saving xtend, clean, etc.)

The fix solves GenTestJava0_0 being added by DeltaConverter, so it would use less memory if it's not referenced by an xtext model like in my case, but it will be added twice if it's referenced.

I've made a test with 1000 xtend files referencing 10 unique java files with 200 variables and 10 common java files referenced by all the xtend files.

With and without the fix, the max memory used (1.4Gb) and the total time seems to be the same (51s). The end memory is 370Mb (GenTestJava* are not cleared from cache and GenTestXtend*.java does not occupy a lot)

The same test but without xtend files is different
Max memory is similar with or without the fix: 400-500Mb
End memory is 300Mb without the fix and 150Mb with the fix (GenTestJava* are cleared from cache)

Anyway, maybe the root cause is using CompilationUnit + JavaModelCache to obtain Java type info? 
Somehow JDT does not use it to do a full clean of a java project, but i don't know how

cdietrich:
@szarnekow @sergio-otero what is the state of this pr

PR #996, opened 2019-02-26
Fixed indentation, improved coding in 'ResourceRelocationProcessor'

sailingKieler:
previous coding could be considered buggy although the behavior is
correct

Signed-off-by: Christian Schneider <christian.schneider@typefox.io>


sailingKieler:
The essential part is this: https://github.com/eclipse/xtext-eclipse/pull/996/files?utf8=%E2%9C%93&diff=unified&w=1

Apart from that Gitpod was so kind and removed some trailing WS, which was not on purpose ;-)

PR #1000, opened 2019-03-02
Refactoring the test case accesses the workspace root.

miklossy:
- Eliminate duplicated code by reusing the
org.eclipse.xtext.ui.testing.util.IResourcesSetupUtil.root() helper
method whenever the test cases need access to the workspace root.

### DIFF
--- a/org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/builderState/EclipseResourceFileSystemAccessTest.java
+++ b/org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/builderState/EclipseResourceFileSystemAccessTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2011, 2019 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -14,11 +14,11 @@ import java.util.List;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
-import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.xtext.builder.EclipseResourceFileSystemAccess;
 import org.eclipse.xtext.builder.tests.BuilderTestLanguageInjectorProvider;
 import org.eclipse.xtext.testing.InjectWith;
 import org.eclipse.xtext.testing.XtextRunner;
+import org.eclipse.xtext.ui.testing.util.IResourcesSetupUtil;
 import org.eclipse.xtext.ui.testing.util.TestedWorkspaceWithJDT;
 import org.eclipse.xtext.util.IAcceptor;
 import org.junit.Assert;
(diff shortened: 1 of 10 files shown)


miklossy:
I agree that deprecating the IResourcesSetupUtil and replacing that by an injected one is a good idea. However refactoring the test base to reuse the IResourcesSetupUtil before would help by the designing how the newly created injectable utility should look like.

cdietrich:
am not sure it test fails are related

miklossy:
If I execute the test cases one by one locally on my computer, they pass. But if I execute all the test cases residing in the `org.eclipse.xtext.xtext.ui.tests` plugin, I get the same test failures locally on my computer. 

The reason for the test failures could be similar to:

> unsere ide injector providers stellen die registry nicht wieder her
>
> wenn der call graph dann so ausschaut:
> 
> - runtime injector provider nimmt snapshot
> - bundle wird aktiviert und injector abgefragt
> - runtime injector provider dreht registry zurück
> - nächstes mal macht der activator nix mehr, da injector bereits da ist
> tot und verderben

@szarnekow has already provided a [fix](https://github.com/eclipse/xtext-xtend/pull/735) to Xtend and if I recall it correctly, he is about to provided a [general solution](https://github.com/eclipse/xtext-xtend/pull/735#issuecomment-467622343).

cdietrich:
@miklossy you can create a test suite (locally)so that the tests run in the same order than on server

```
@RunWith(Suite.class)
@SuiteClasses({
	XtextGrammarQuickfixProviderTest.class,
	XtextGrammarQuickfixTest.class
// TODO will with testcases as they run on server
	
})
public class ViechSuite {

}
```

org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/Activator.java
```diff
@@ -62,5 +64,14 @@ private static void setDefault(Activator activator) {
 	public static ImageDescriptor getImageDescriptor(String path) {
 		return imageDescriptorFromPlugin(PLUGIN_ID, path);
 	}
+	
+	/**
+	 * TODO: remove this workaround as soon as a clear solution is
+	 * provided to make the UiInjectorProvider registry aware.
+	 */
+	@Override
```
cdietrich: @szarnekow @miklossy what is this needed for?

miklossy: The PR contains actually two commits. The first commit contains the refactoring, but that lead to some failing XtextGrammarQuickfixProviderTest test cases due to the fact that the UiInjectorProvider has not been registry aware. If that problem has already been solved, that workaround is not necessary any more and the second commit can be removed from the PR.

cdietrich:
@szarnekow @miklossy how shall we proceed? close this pr?

szarnekow:
I'd prefer the change with createInjector to be removed. Otherwise the PR looks ok.

cdietrich:
@miklossy any news?

miklossy:
I will update the PR soon.

miklossy:
I updated the PR as requested: I removed the commit with the workaround so that the PR contains only one commit.

miklossy:
For the [refactoring](https://github.com/eclipse/xtext-eclipse/pull/1000#pullrequestreview-209867244) I created a separate issue #1092

cdietrich:
thx @miklossy

PR #1007, opened 2019-03-07
[#1006] isReadOnly should consider XtextReadonlyEditorInput inputs

kthoms:
When element is an instance of XtextReadonlyEditorInput isReadOnly() can
directly return true

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/XtextDocumentProvider.java
@@ -51,6 +51,7 @@ import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.editors.text.FileDocumentProvider;
 import org.eclipse.xtext.parser.IEncodingProvider;
 import org.eclipse.xtext.resource.XtextResource;
+import org.eclipse.xtext.ui.editor.XtextReadonlyEditorInput;
 import org.eclipse.xtext.ui.editor.quickfix.IssueResolutionProvider;
 import org.eclipse.xtext.ui.editor.quickfix.XtextResourceMarkerAnnotationModel;
 import org.eclipse.xtext.ui.editor.validation.AnnotationIssueProcessor;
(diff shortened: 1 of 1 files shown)


ArneDeutsch (commented):
Looks a bit like a worksround for me. Maybe there is a bug/missing feature in `org.eclipse.xtext.ui.editor.utils.EditorUtils#createEditorInput` that needs to be handled instead? Do you know for what kind of `IStorage` objects this happens? Maybe it is just a missing case in that method ...

kthoms:
It is `EditorUtils#createEditorInput` that decides which editor input is to be created. This can be XtextReadonlyEditorInput, or the user could customize that set such an input.
The problem is then later that the XtextDocumentProvider will decide that this is still not read-only.
BTW: We started manipulating the creation of the editor input and then realized that the XtextDocumentProvider might be the better place to solve.

ArneDeutsch:
It is a bit weird that the IStorage object tells that it is NOT readOnly, but we consider it to be readOnly for Xtext. If someone else is calling `((IStorage) element).getStorage().isReadOnly()` he will get a different answer then from `IDocumentProviderExtension.isReadOnly(element)`. But probably this is exactly what is correct here ... the element itself may be edited, but the document generated from it may not ...

cdietrich:
@ArneDeutsch by why do we create the readonlyeditorinput?

ArneDeutsch:
I guess because we don't know if we can edit it because it is neither a IFile, a File or from a JAR, hence we just choose "readOnly" ... but thats exactly my point. Would be interesting for what kind of IStorage objects we assume it is "readOnly" because probably we can handle them different. Probable they are editable instead and we (needlessy) make them readonly with this change.

Interesting observation: right now we create this input but actually do nothing with it. Except from test cases there is no reference to `XtextReadonlyEditorInput`. Hence I would say it is probably just a bad name ... the responsibiliy to decide if something is editable is not the one from the IEditorInput but from IStorage instead.

ArneDeutsch:
Actually I start to doubt if eclipse/xtext#2479 is valid. Of course I do not know the issue that triggered the change, maybe there are good reasons. But from the ticket alone I would say the change is needless, just driven by a misleading class name.

cdietrich:
@szarnekow maybe you remember why it was done in the first place and if this change makes sense.

cdietrich:
wont fix

PR #1008, opened 2019-03-07
[eclipse/xtext-xtend#395] Try With Resources

epoell:
Support Content Proposals of used resources within expression block

Signed-off-by: Eva Poell <epoell@itemis.com>


epoell:
Hello,
I need another hint. I cannot find the right place for a testcase that gives something like `try (val a = new StringReader("Test")) { val b = ` as input and expects the usual list, including `a` as Content Proposals.

I guess it should be in a class, that is extending `AbstractXbaseContentAssistInBlockTest` but I'm not sure and I do not find fitting classes. The one you mentioned above seems to contain unused methods. In my attempts, the debugger would not halt at breakpoints in these methods of  `org.eclipse.xtext.xbase.ui.tests.editor.ContentAssistTests.xtend`.

Thanks already!
Eva

szarnekow:
Hi @epoell,

it will not be possible to have these test cases in the Xbase tests, since Xbase does not allow this syntax. You'll have to add these to the subtypes of the ContentAssistTest in the Xtend-UI-Tests.

epoell:
These tests succeed locally in my workspace and actually do not look connected to what I committed. Could you start a new build or test run?

epoell:
Thanks Christian!

ArneDeutsch:
@epoell I merged this PR accidently. It is already reverted ... but would you be so kind and create the PR again? Sorry for the trouble!

epoell:
Hello,
I'm sorry, I couldn't come around to do so until now.
And I'm not sure how to do it. I tried to add a new Pull Request for the brnach ep_issue395 into master, but there are no changes and hence no button to issue a Pull Request...
Nothing to compare

szarnekow:
I think we'll have to revert the revert.

PR #1012, opened 2019-03-08
mark abstract classes as abstract

cdietrich:
mark abstract classes as abstract

do you think this is ok api wise?


szarnekow:
> do you think this is ok api wise?

Yes

Issue #1022, opened 2019-03-24
AbstractQuickfixTest.quickfixesAreOffered is order dependent


cdietrich:
AbstractQuickfixTest.quickfixesAreOffered is order dependent

https://ci.eclipse.org/xtext/job/xtext-eclipse/job/cd_isssue991/1/testReport/junit/org.eclipse.xtext.xtext.ui.editor.quickfix/XtextGrammarQuickfixTest/test_fix_empty_keyword/

cdietrich:
Fixed in 218

PR #1023, opened 2019-03-24
[eclipse/xtext-eclipse#1022] sort quickfixes by label to have a stable order if multiple quickfixes are available

cdietrich:
[eclipse/xtext-eclipse#1022] sort quickfixes by label to have a stable order if multiple quickfixes are available


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractQuickfixTest.xtend
```diff
@@ -116,7 +116,7 @@ abstract class AbstractQuickfixTest extends AbstractEditorTest {
 		val originalText = document.get
 		val issue = document.getValidationIssue(issueCode)
 
-		val actualIssueResolutions = issue.getResolutions
+		val actualIssueResolutions = issue.getResolutions.sortBy[label]
```
miklossy: I would sort not only the actual, but also the expected quickfixes.

cdietrich: Well there is no additional value in that. Thus I keep it one side

cdietrich: You can come up with a second or after this one

miklossy: If you sort the expected values here, there is no need to modify the concrete test cases.

cdietrich: Can you please come up with ape and we will close this one

Issue #1044, opened 2019-04-06
Use IRegion interface instead of the Region class in the AbstractHyperlink class.


miklossy:
Currently, the [hyperlinkRegion](https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hyperlinking/AbstractHyperlink.java#L20) field in the org.eclipse.xtext.ui.editor.hyperlinking.AbstractHyperlink class has the `org.eclipse.jface.text.Region` type.

I suggest to use the `org.eclipse.jface.text.IRegion` type instead.

miklossy:
Fixed in 2.18.

PR #1045, opened 2019-04-06
[#1044] AbstractHyperlink: Usage of IRegion instead of Region.

miklossy:

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hyperlinking/AbstractHyperlink.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hyperlinking/AbstractHyperlink.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2009, 2019 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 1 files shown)


cdietrich:
nothing on github,
but there are also customizations in xbase.ui and usages in domainmodel. am not sure if compatibe

miklossy:
@cdietrich Thanks for checking that! The usages in the xbase.ui and in the domainmodel are compatible with the change, otherwise there would be some compile errors.

cdietrich:
Domainnodel is not part of the build

miklossy:
Ah ok, you are right. I checked that again and the [DomainmodelHyperlinkHelper](https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.ui/src/org/eclipse/xtext/example/domainmodel/ui/navigation/DomainmodelHyperlinkHelper.java#L72) invokes the setHyperlinkRegion method with a Region parameter. The PR changes that parameter from Region to IRegion, and since Since all Region objects are instances of IRegion, it remains compatible.

szarnekow:
@cdietrich Are we aiming at source compatibility or binary compatibility?

cdietrich:
we aim to both. but as always it is a matter of compromise

szarnekow:
I'd keep `setHyperlinkRegion(Region)` and deprecate it.

miklossy:
@szarnekow Good idea! I updated the PR as requested.

PR #1052, opened 2019-04-15
Also consider the listener job to be a manual build job

szarnekow:


cdietrich:
TODO restart build as soon as server is usable again.
test failure does not seem to be related

szarnekow:
Build was green. Merging.

szarnekow:
Test failures can be reproduced locally with test suite that mimics the order of test classes on the server. Investigating before merge.

szarnekow:
org.eclipse.xtext.ui.tests.core.resource.Storage2UriMapperTest does not remove the project that it creates.

PR #1060, opened 2019-04-18
[refactoring2] modified 'ResourceRelocationProcessor' to switch to the persisted index if required

sailingKieler:
…more specifically, if a custom resource relocation strategy impl demands for relying on the persisted resource index.

enables the avoidance of heap pollution and resultant performance decrease due to creating the live scope resource descriptions of the affected resources before executing the actual modifications; can easily happen if the amount of files being moved or renamed is too large

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/participant/ResourceRelocationProcessor.xtend
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/participant/ResourceRelocationProcessor.xtend
@@ -28,6 +28,7 @@ import org.eclipse.xtext.ide.refactoring.ResourceRelocationChange
 import org.eclipse.xtext.ide.refactoring.ResourceRelocationContext
 import org.eclipse.xtext.ide.refactoring.ResourceRelocationContext.ChangeType
 import org.eclipse.xtext.ide.serializer.IChangeSerializer
+import org.eclipse.xtext.resource.impl.ResourceDescriptionsProvider
 import org.eclipse.xtext.ui.refactoring2.ChangeConverter
 import org.eclipse.xtext.ui.refactoring2.LtkIssueAcceptor
 import org.eclipse.xtext.ui.refactoring2.ResourceURIConverter
@@ -75,8 +76,13 @@ class ResourceRelocationProcessor {
 		
 		val changeSerializer = changeSerializerProvider.get();
 		val resourceSet = resourceSetProvider.get(project)
-		liveScopeResourceSetInitializer.initialize(resourceSet)
+
 		val context = new ResourceRelocationContext(type, uriChanges, issues, changeSerializer, resourceSet)
+		val persistedIndexUsageRequested = context.isPersistedIndexUsageRequested()
+		
+		// TODO check preconditions like all editors being saved if 'persistedIndexUsageRequested' == true
+		
+		initializeResourceSet(persistedIndexUsageRequested, context)
 		executeParticipants(context, subMonitor.split(1)) // requires effort of loading affected resources
 		
 		val changeConverter = changeConverterFactory.create(name, [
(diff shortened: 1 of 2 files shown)


releng/org.eclipse.xtext.target/org.eclipse.xtext.target-oxygen.target
```diff
@@ -32,7 +32,7 @@
 		<unit id="org.eclipse.xtext.testlanguages.ide.source" version="0.0.0"/>
 		<unit id="org.eclipse.xtext.tests" version="0.0.0"/>
 		<unit id="org.eclipse.xtext.tests.source" version="0.0.0"/>
-		<repository location="https://ci.eclipse.org/xtext/job/xtext-core/job/master/lastStableBuild/artifact/build/p2-repository/"/>
+		<repository location="https://jenkins.eclipse.org/xtext/job/xtext-core/job/cs-refactoringIndexing2/lastStableBuild/artifact/build/p2-repository/"/>
```
cdietrich: dont forget to revert

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/participant/ResourceRelocationProcessor.xtend
```diff
@@ -90,6 +96,24 @@ class ResourceRelocationProcessor {
 		modificationApplicationMonitor.done
 		return changeConverter.change
 	}
+	
+	protected def boolean isPersistedIndexUsageRequested(ResourceRelocationContext context) {
```
cdietrich: can you add sinces to these two

releng/org.eclipse.xtext.target/org.eclipse.xtext.target-oxygen.target
sailingKieler: sure :-)

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/participant/ResourceRelocationProcessor.xtend
sailingKieler: right, thanks for that hint

sailingKieler: tags are in there, now

sailingKieler:
a successful build ran here: https://ci.eclipse.org/xtext/job/xtext-eclipse/job/cs-refactoringIndexing2/

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring2/participant/ResourceRelocationProcessor.xtend
```diff
@@ -90,6 +96,30 @@ class ResourceRelocationProcessor {
 		modificationApplicationMonitor.done
 		return changeConverter.change
 	}
+	
+	/**
+	 * @since 2.18
+	 */
+	protected def boolean isPersistedIndexUsageRequested(ResourceRelocationContext context) {
+		val strategies = strategyRegistry.strategies
+		val persistedIndexUsageRequested = strategies.exists[
```
szarnekow: I think that the live scope is mandated if at least one strategy requires the live scope. As of now a single persisted descr request will break other strategies that rely on live scoping

sailingKieler: That is indeed true, and for that reason I tried it the other way round in the first place, but as @cdietrich [correctly remarked](https://github.com/eclipse/xtext-core/commit/265e4c97e3689747cf051503f5d21e870c019b06#commitcomment-33214817) that would imply a change in the default behavior.

We also discussed that compromise internally as all of the possible options of introducing that switch do have some drawbacks.

That's why I put the detailed API documentation in https://github.com/eclipse/xtext-core/blob/247a38a20e5d0db666a981dad661b89ed771da83/org.eclipse.xtext.ide/src/org/eclipse/xtext/ide/refactoring/IResourceRelocationStrategy.xtend#L32-L54, which hopefully makes the consequences of returning `true` clear (I on purpose tried to be kind of formal there).

Consequently, languages should not just override the method with a `return true` body but evaluate the information hold by the provided `ResourceRelocationContext` for making a proper decision, as long as they don't live in a closed world setup.

PR #1062, opened 2019-05-03
Prevent glitchy outline select

tivervac:
1. Open an Xtend editor
2. Open the outline view
3. Type something in the Xtend editor such that the outline refreshes
4. After typing quickly press (and hold) shift and tap the left arrow button to start selecting text
5. Note how the selection jumps around resulting in the last few characters not being selected

I don't really see why the outline would like to change the `SourceViewer`'s selection when it refreshes but feel free to let me know if I'm missing something


tivervac:
I did just notice there's a commit (f86fd57a62f5afa9b63eb5c8c9c0b202e898aa34) associated to these two lines however.

cdietrich:
does this change reopen bug 439686?

tivervac:
Don't think it does. From what I see in that ticket, the outline used to lose its selection when the page refreshed, with my patch it doesn't seem to lose selection.

cdietrich:
thx @tivervac

PR #1071, opened 2019-05-08
Make DiagnosticTreeIterator return a new iterator every time

tivervac:
DiagnosticTreeIterator always returns the same iterator when you call
iterator(). This means it can't be iterated twice.

The `Iterable` interface does not force you to return a new `Iterator` every time but this is common practice. If this change doesn't get accepted because of this reason, it would be nice to at least have a comment on `iterator()`.


org.eclipse.xtext.junit4.tests/src/org/eclipse/xtext/junit4/DiagnosticTreeIterableTest.java
```diff
@@ -0,0 +1,26 @@
+package org.eclipse.xtext.junit4;
```
szarnekow: Please add a copyright header.

tivervac: Done

cdietrich:
@szarnekow @tivervac ping. what is the state. how shall we continue

Issue #1066, opened 2019-05-06
xtext 2.17: Tons of exceptions logged at ProjectOpenedOrClosedListener.processClosedProjects


trancexpress:
We are trying out `xtext 2.17.1` as an update from `xtext 2.16` for our product.

During our ARTs we see tons and tons of exceptions with the following stack trace:

```
Unexpected log from plugin: org.eclipse.core.runtime, message: 'An internal error occurred during: "Removing project .org.eclipse.jdt.core.external.folders from Xtext index.".'
org.eclipse.core.runtime.AssertionFailedException: null argument:
	at org.eclipse.core.runtime.Assert.isNotNull(Assert.java:88)
	at org.eclipse.core.runtime.Assert.isNotNull(Assert.java:76)
	at org.eclipse.core.runtime.Path.initialize(Path.java:644)
	at org.eclipse.core.runtime.Path.<init>(Path.java:211)
	at org.eclipse.core.internal.resources.WorkspaceRoot.getProject(WorkspaceRoot.java:148)
	at abc.XYZ.doUpdate(XYZ.java:300)
	at org.eclipse.xtext.builder.builderState.AbstractBuilderState.update(AbstractBuilderState.java:116)
	at org.eclipse.xtext.builder.impl.ProjectOpenedOrClosedListener.processClosedProjects(ProjectOpenedOrClosedListener.java:275)
	at org.eclipse.xtext.builder.impl.ProjectOpenedOrClosedListener$RemoveProjectsJob.runInWorkspace(ProjectOpenedOrClosedListener.java:104)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:42)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

After investigating, we observed that we have a class which overrides the method  `org.eclipse.xtext.builder.clustering.ClusteringBuilderState.doUpdate(BuildData, ResourceDescriptionsData, IProgressMonitor)`; the first call of our implementation is:

```
IProject project = workspace.getRoot().getProject(buildData.getProjectName());
```

This code is called `null` with project name by `org.eclipse.xtext.builder.impl.ProjectOpenedOrClosedListener.processClosedProjects(IProgressMonitor)` in the following code block:

```
try {
	ResourceSet resourceSet = createResourceSet();
	BuildData buildData = new BuildData(null, resourceSet, task.getToBeBuilt(), queuedBuildData, false,
[...]

PR #1076, opened 2019-05-21
[#1066] Pass all participating projects along with BuildData

szarnekow:
[#1066] Pass all participating projects along with BuildData

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuildData.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuildData.java
@@ -14,6 +14,9 @@ import java.util.Set;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.resource.ResourceSet;
 import org.eclipse.xtext.resource.IResourceDescription;
+import org.eclipse.xtext.util.Strings;
+
+import com.google.common.collect.ImmutableSet;
 
 /**
  * @author Sebastian Zarnekow - Initial contribution and API
@@ -23,6 +26,15 @@ public class BuildData {
 	private final ToBeBuilt toBeBuilt;
 	private final QueuedBuildData queuedBuildData;
 	private final String projectName;
+	/**
+	 * If projects are deleted from the workspace, they are not participating in a
+	 * dedicated build but may be processed along with other project builds.
+	 * 
+	 * The participating projects may be obtained via {@link #getParticipatingProjectNames()}
+	 * 
+	 * @since 2.18
+	 */
+	private final ImmutableSet<String> participatingProjectNames;
 	private final ResourceSet resourceSet;
 	private final boolean indexingOnly;
 	private final SourceLevelURICache sourceLevelURICache;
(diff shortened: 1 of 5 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuildData.java
```diff
@@ -45,7 +57,26 @@ public BuildData(String projectName, ResourceSet resourceSet, ToBeBuilt toBeBuil
...
+		this(participatingProjectNames.iterator().next(), resourceSet, toBeBuilt, queuedBuildData, indexingOnly, buildRequestor, participatingProjectNames);
+	}
+	
+	/**
+	 * @since 2.18
+	 */
+	public BuildData(String projectName, ResourceSet resourceSet, ToBeBuilt toBeBuilt, QueuedBuildData queuedBuildData, boolean indexingOnly, Runnable buildRequestor, Set<String> participatingProjectNames) {
 		this.projectName = projectName;
+		if (participatingProjectNames.contains(projectName) || projectName == null) {
+			this.participatingProjectNames = ImmutableSet.copyOf(participatingProjectNames);
```
cdietrich: am not sure if i like the mix of guava and std java api for set immuntability

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -319,15 +320,62 @@ protected void incrementalBuild(IResourceDelta delta, final IProgressMonitor mon
...
 
-	private void addInfosFromTask(Task task, final ToBeBuilt toBeBuilt) {
+	private static Boolean mustCallDeprecatedDoBuild = null;
+	private static boolean wasDeprecationWarningLogged = false;
+	
+	/**
+	 * @since 2.18
+	 */
+	protected void addInfosFromTaskAndBuild(Task task, final ToBeBuilt toBeBuilt, BuildType buildType, IProgressMonitor monitor)
+			throws CoreException {
```
cdietrich: highly sophisticated :)

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -319,15 +320,62 @@ protected void incrementalBuild(IResourceDelta delta, final IProgressMonitor mon
...
+				doBuild(toBeBuilt, task.getProjectNames(), monitor, buildType);
+			}
+		}
+		
+	}
+
+	/**
+	 * @since 2.18
+	 */
+	protected boolean isDoBuildSpecialized(Class<?> c) {
```
cdietrich: i wonder if we can do something not to keep this stuff forever e.g. announcing a removal in 2020

szarnekow: I'll mark it as deprecated immediately.

cdietrich: (same for the others that are for compatibility)

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
szarnekow: I tried to break only few things.

What I was also thinking about: changing the constructor of BuildData to smth like `BuildData(Map<Key<T>, T>)` to become more forward compatible.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
szarnekow: I think I caught the relevant candidates.

szarnekow:
@cdietrich What do you think?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/ProjectOpenedOrClosedListener.java
```diff
@@ -263,15 +263,22 @@ protected void processClosedProjects(IProgressMonitor monitor) {
...
 		String taskName = Messages.ProjectOpenedOrClosedListener_RemovingProject + projectNames
 				+ Messages.ProjectOpenedOrClosedListener_FromIndex;
 		monitor.setTaskName(taskName);
 		SubMonitor progress = SubMonitor.convert(monitor, 1);
 		try {
 			ResourceSet resourceSet = createResourceSet();
-			BuildData buildData = new BuildData(null, resourceSet, task.getToBeBuilt(), queuedBuildData, false,
-					BuildManagerAccess::requestBuild);
+			BuildData buildData = new BuildData(
+					"",
```
cdietrich: null?

szarnekow: If you use `workspace.getProject(buildData.getProjectName())` you'd get an assertion error. With the empty string, you'll find a project that does not exist. We do have a test that fails if we use null here. I'm not really passionate - changing it into one or the other direction would be ok'ish. But the empty string is more failsafe I suppose.

cdietrich: ok. can you please add a javadoc comment to BuildData.getProjectName

PR #1077, opened 2019-05-21
[eclipse/xtext#1460] Two phase lock of the workspace in BuilderState.load

szarnekow:
[eclipse/xtext#1460] Two phase lock of the workspace in BuilderState.load

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/AbstractBuilderState.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/AbstractBuilderState.java
@@ -13,6 +13,10 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Set;
 
+import org.apache.log4j.Logger;
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.IWorkspaceRunnable;
+import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.SubMonitor;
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/AbstractBuilderState.java
```diff
@@ -52,22 +58,48 @@
...
+					workspace.run(runnable, null, IWorkspace.AVOID_UPDATE, null);
+				}
+			} catch(CoreException e) {
+				log.error(e.getMessage(), e);
+			}
+		}
+	}
+	
+	@SuppressWarnings("restriction")
+	private boolean isWorkspaceLocked() {
```
kthoms: In Storage2UriMapperJavaImpl the helper method is called isWorkspaceLockedByCurrentThread. Shall they both be named equally?
This makes use of internal API. I did not see a way to achieve the same without these internals. Would it make sense to request an API method for future use, e.g. on IWorkspace?

cdietrich: good catch. and there should be one method only

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/AbstractBuilderState.java
```diff
@@ -52,22 +58,48 @@
...
+		try {
+			return ((org.eclipse.core.internal.resources.Workspace) workspace).getWorkManager().isLockAlreadyAcquired();
+		} catch (CoreException e) {
+			return false;
```
cdietrich: _workManager may become null on shutdown. = > CoreExceptionso i am thinking if returning false is correct here

PR #1083, opened 2019-05-27
[#1460] Handle shutdown scenario properly

szarnekow:
[#1460] Handle shutdown scenario properly

### DIFF
--- a/org.eclipse.xtext.ui.shared/src/org/eclipse/xtext/ui/shared/internal/SharedModule.java
+++ b/org.eclipse.xtext.ui.shared/src/org/eclipse/xtext/ui/shared/internal/SharedModule.java
@@ -64,6 +64,7 @@ import org.eclipse.xtext.ui.shared.JdtHelper;
 import org.eclipse.xtext.ui.shared.contribution.ISharedStateContributionRegistry;
 import org.eclipse.xtext.ui.util.IJdtHelper;
 import org.eclipse.xtext.ui.workspace.EclipseProjectConfigProvider;
+import org.eclipse.xtext.ui.workspace.WorkspaceLockAccess;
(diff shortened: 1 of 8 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/workspace/WorkspaceAccess.java
```diff
@@ -0,0 +1,40 @@
...
+@SuppressWarnings("restriction")
+public interface WorkspaceAccess {
+	
+	enum Result {
+		YES,
+		NO,
+		SHUTDOWN
+	}
+
+	static Result isWorkspaceLockedByCurrentThread(IWorkspace workspace) {
```
cdietrich: am not sure if i like the static here (for patchability)

szarnekow: I'll make this injectable.

org.eclipse.xtext.ui.shared/src/org/eclipse/xtext/ui/shared/internal/SharedModule.java
```diff
@@ -203,6 +204,7 @@ protected void configure() {
 				expose(ReferenceSearchViewPage.class);
 			}
 		});
+		bind(WorkspaceLockAccess.class);
```
cdietrich: i wonder if we have any scheme for ordering the stuff here

szarnekow: I did not find an obvious spot, so I put it at the end.

PR #1091, opened 2019-06-07
[eclipse/xtext#1463] Configure timeout & timestamps

Hejado:
Signed-off-by: Nico Prediger <mail@nicoprediger.de>


cdietrich:
thx @Hejado

PR #1098, opened 2019-06-11
increased timeout cause ci.eclipse.org gets more and more crappy

cdietrich:


cdietrich:
cc @Hejado

Issue #1108, opened 2019-06-19
Strange positioning of warnings/errors in wizard


cdietrich:
Strange positioning of warnings/errors in wizard:
am not sure if this happens on mac and windows too

![Bildschirmfoto vom 2019-06-19 21-31-54](https://user-images.githubusercontent.com/238736/59794689-d2943e80-92d9-11e9-8535-64a0b8c7a5e8.png)

cdietrich:
=> seems to be connected to low resolution

cdietrich:
resizing helps. so it looks there is a redraw missing or something like that

cdietrich:
i wonder why we have a status widget on page 2 and use the header on page 1

cdietrich:
fixed

miklossy:
I still observe a problem regarding the layouting on my computer:
![screencast](https://user-images.githubusercontent.com/20393472/60487340-fb381300-9c9f-11e9-8ef3-956f8b47347f.gif)

cdietrich:
can you please analyze and debug this on windows?

miklossy:
The layouting looks properly (at least on my computer) when setting the `widthHint` of the image layoutdata and the `heighHint` of the link layout data. After merging the PR I will test this on several computers.

miklossy:
Fixed in 2.19

PR #1109, opened 2019-06-19
[#1108] fixed layout issue in 2nd new project page status widget

cdietrich:


cdietrich:
have no idea if this is the correct way. maybe somebody with deeper swt knowledge can have a look
@ArneDeutsch @miklossy

PR #1115, opened 2019-06-25
[eclipse/xtext#1459] Change p2.buildship value to use Buildship 3.x

kthoms:
On current Eclipse installations Buildship 3.x is already picked from
the Eclipse SimRel repository. Current Buildship 3.1.0 is installable
also on Eclipse Oxygen, so changing the Buildship repository URL allows
to use the same Buildship version on Eclipse 4.7 and above.

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -139,7 +139,7 @@
       xsi:type="setup:VariableTask"
       type="URI"
       name="p2.buildship"
-      value="https://download.eclipse.org/buildship/updates/e47/releases/2.x"/>
+      value="https://download.eclipse.org/buildship/updates/e411/releases/3.x"/>
   <setupTask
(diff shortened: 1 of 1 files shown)


releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -139,7 +139,7 @@
       xsi:type="setup:VariableTask"
       type="URI"
       name="p2.buildship"
-      value="https://download.eclipse.org/buildship/updates/e47/releases/2.x"/>
+      value="http://download.eclipse.org/buildship/updates/e411/releases/3.x"/>
```
cdietrich: is there no https?

kthoms: Already changed, sorry.

kthoms: Copied the location from the issue

Issue #1132, opened 2019-07-09
XtextGrammarQuickfixProvider.workbench is not @Nullable


cdietrich:
```
1) null returned by binding at org.eclipse.xtext.ui.shared.SharedStateModule.configureIWorkbench(SharedStateModule.java:140)
 but org.eclipse.xtext.xtext.ui.editor.quickfix.XtextGrammarQuickfixProvider.workbench is not @Nullable
  while locating org.eclipse.ui.IWorkbench
    for field at org.eclipse.xtext.xtext.ui.editor.quickfix.XtextGrammarQuickfixProvider.workbench(Unknown Source)
  while locating org.eclipse.xtext.xtext.ui.editor.quickfix.XtextGrammarQuickfixProvider
  while locating org.eclipse.xtext.ui.editor.quickfix.IssueResolutionProvider
    for field at org.eclipse.xtext.ui.editor.validation.MarkerCreator.resolutionProvider(Unknown Source)
  while locating org.eclipse.xtext.ui.editor.validation.MarkerCreator
    for field at org.eclipse.xtext.ui.validation.DefaultResourceUIValidatorExtension.markerCreator(Unknown Source)
  while locating org.eclipse.xtext.ui.validation.DefaultResourceUIValidatorExtension
  while locating org.eclipse.xtext.ui.validation.IResourceUIValidatorExtension
```

kthoms:
fixed in 2.19

PR #1133, opened 2019-07-09
[#1132] marked workbench as nullable

cdietrich:
[#1132] marked workbench as nullable


kthoms:
Sorry for asking: What's the difference between optional injections with and without that annotation?
When should optional injections annotated additionally with Nullable? I found only 3 usages of Nullable yet, but many optional Injects.

cdietrich:
guice does not like optional only it needs to be nullable too. we already had this at different places. https://stackoverflow.com/questions/4887196/guice-nullable-annotation

kthoms:
Thanks for clarifying. Are there other places where Nullable would make sense then, or do we change only when we hit such places like here?

cdietrich:
in the past we fixed only places we hit

PR #1142, opened 2019-07-17
Extend IProjectConfig and related interfaces

mmews-n4:
see https://github.com/eclipse/xtext-core/pull/1209

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/workspace/EclipseProjectConfig.xtend
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/workspace/EclipseProjectConfig.xtend
@@ -8,6 +8,7 @@
 package org.eclipse.xtext.ui.workspace
 
 import com.google.inject.Inject
+import com.google.inject.Singleton
 import java.util.Set
 import org.eclipse.core.resources.IProject
 import org.eclipse.core.resources.IWorkspaceRoot
@@ -20,9 +21,6 @@ import org.eclipse.xtext.workspace.IProjectConfigProvider
 import org.eclipse.xtext.workspace.ISourceFolder
 import org.eclipse.xtext.workspace.IWorkspaceConfig
 import org.eclipse.xtext.workspace.ProjectConfigAdapter
-
-import static extension org.eclipse.xtext.util.UriUtil.*
-import com.google.inject.Singleton
 
 @Singleton
 class EclipseProjectConfigProvider implements IProjectConfigProvider {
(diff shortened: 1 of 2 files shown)


cdietrich:
TODO temp. use the stable build repo built by your upstream PR (once its green) in oxygengen target to see if this is actually green here

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/workspace/EclipseProjectConfig.xtend
```diff
@@ -66,8 +64,12 @@ class EclipseProjectConfig implements IProjectConfig {
 		emptySet
 	}
 
+	override getOutputFolders() {
```
szarnekow: This would be a good default impl, if we decide to keep that method on the interface.

mmews-n4: Thanks. Right, but in favour for keeping the PR as small as possible, I removed the output folders again and work with another solution.

mmews-n4:
@kittaakos  Should I also squash the two commits here? What about the build problems (are they valid?)?

kittaakos:
> Should I also squash the two commits here

In general, yes: having `add output folder`, then `remove output folder` in the history does not have much value.
Regarding the build failure, I can see this in the logs:
```
Caused by: org.jenkinsci.plugins.workflow.steps.MissingContextVariableException: Required context class hudson.FilePath is missing
Perhaps you forgot to surround the code with a step that provides this, such as: node
	at org.jenkinsci.plugins.workflow.steps.StepDescriptor.checkContextAvailability(StepDescriptor.java:266)
	at org.jenkinsci.plugins.workflow.cps.DSL.invokeStep(DSL.java:260)
	... 58 more
```
If you squash and push, maybe the build will magically repair itslelf :) Do not know, let's see if it fails after the squash.

mmews-n4:
Still there are failing builds. Are they valid or bogus? I cannot see how I can fix it. Do you have any idea?

cdietrich:
As I said you have to change the oxygen target to use stuff from your Xtext core
https://ci.eclipse.org/xtext/job/xtext-core/view/change-requests/job/PR-1209/lastSuccessfulBuild/artifact/build/p2-repository/

mmews-n4:
Screenshot 2019-07-19 at 11 46 32

See above that the build is green. I undo the last commit and squash again to reduce number of commits again.

mmews-n4:
Screenshot 2019-07-19 at 12 03 26

Now, two builds are green, which would be all builds I think.

cdietrich:
@mmews-n4 are you sure you reverted the right thing?

mmews-n4:
I wasn't, but now it should be.

BTW: Thanks for your hint regarding the oxygen target. The first time you noted it (see https://github.com/eclipse/xtext-eclipse/pull/1142#issuecomment-512210567) I didn't know what you mean.

PR #1146, opened 2019-07-21
[#1044] Fix warning: the method setHyperlinkRegion(Region) is deprecated

miklossy:
- Modify the org.eclipse.xtext.ui.editor.hyperlinking.HyperlinkHelper
class to use the setHyperlinkRegion(IRegion) instead of the deprecated
setHyperlinkRegion(Region) method of the AbstractHyperlink class.
- Fix the deprecation warnings also in the derived classes.

### DIFF
--- a/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtHyperlinkFactory.java
+++ b/org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtHyperlinkFactory.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2009, 2019 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -10,6 +10,7 @@ package org.eclipse.xtext.common.types.xtext.ui;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.ui.JavaElementLabels;
+import org.eclipse.jface.text.IRegion;
 import org.eclipse.jface.text.Region;
 import org.eclipse.xtext.common.types.JvmIdentifiableElement;
 import org.eclipse.xtext.common.types.access.TypeResource;
@@ -34,6 +35,30 @@ public class JdtHyperlinkFactory {
 		return to instanceof JvmIdentifiableElement && to.eResource() instanceof TypeResource;
 	}
 
+	/**
+	 * @since 2.19
+	 */
+	public void createHyperlink(IRegion region, EObject to, IHyperlinkAcceptor acceptor) {
+		if (region instanceof Region) {
+			createHyperlink((Region) region, to, acceptor);
+		} else {
+			JvmIdentifiableElement element = (JvmIdentifiableElement) to;
+			IJavaElement javaElement = javaElementFinder.findElementFor(element);
+			if (javaElement == null)
+				return;
+			String label = JavaElementLabels.getElementLabel(javaElement, JavaElementLabels.ALL_DEFAULT);
+			JdtHyperlink result = jdtHyperlinkProvider.get();
+			result.setHyperlinkRegion(region);
+			result.setHyperlinkText(label);
+			result.setJavaElement(javaElement);
+			acceptor.accept(result);
+		}
+	}
+
+	/**
+	 * @deprecated use {@link #createHyperlink(IRegion, EObject, IHyperlinkAcceptor)} instead.
+	 */
+	@Deprecated
 	public void createHyperlink(Region region, EObject to, IHyperlinkAcceptor acceptor) {
 		JvmIdentifiableElement element = (JvmIdentifiableElement) to;
 		IJavaElement javaElement = javaElementFinder.findElementFor(element);


org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtHyperlinkFactory.java
```diff
@@ -34,7 +34,7 @@ public boolean canHandle(EObject to) {
 		return to instanceof JvmIdentifiableElement && to.eResource() instanceof TypeResource;
 	}
 
-	public void createHyperlink(Region region, EObject to, IHyperlinkAcceptor acceptor) {
```
cdietrich: looks like unwanted api changes to be. maybe adding a second method. doing reflection to find out what is overriden, mark old method deprecated, ....? eclipse api rules are bad

cdietrich: @szarnekow any comments / ideas

miklossy: I updated the PR introducing a second method with the `IRegion` parameter and deprecated the method with the `Region` parameter.

cdietrich: did you check if there are callers of the method in xtext codebase? and if yes how to ensure calling the old/new method depending on what is overriden by client?

miklossy: I found only one place, the [XtendHyperlinkHelper](https://github.com/eclipse/xtext-xtend/blob/master/org.eclipse.xtend.ide/src/org/eclipse/xtend/ide/hyperlinking/XtendHyperlinkHelper.java#L75)
screenshot
That deprecation can be addressed by declaring/casting the actual parameter from `Region` to `IRegion`. I will address this in a separate PR in the Xtend code base as soon as this one gets merged.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/navigation/XbaseHyperLinkHelper.java
```diff
@@ -171,8 +172,9 @@ protected void createHyperlinksInJavaDoc(XtextResource resource, int offset, IHy
 		EObjectInComment eObjectReferencedInComment = javaDocTypeReferenceProvider.computeEObjectReferencedInComment(resource, offset);
 		if(eObjectReferencedInComment != null) {
 			EObject target = eObjectReferencedInComment.getEObject();
-			ITextRegion region = eObjectReferencedInComment.getRegion();
-			createHyperlinksTo(resource, new Region(region.getOffset(), region.getLength()), target, acceptor);
+			ITextRegion textRegion = eObjectReferencedInComment.getRegion();
+			IRegion region = new Region(textRegion.getOffset(), textRegion.getLength());
+			createHyperlinksTo(resource, region, target, acceptor);
```
cdietrich: this is what concerns me

cdietrich: :(

miklossy: Actually the PR contains the following refactoring:
using the code
```
ITextRegion textRegion = eObjectReferencedInComment.getRegion();
IRegion region = new Region(textREgion.getOffset(), textRegion.getLength());
createHypelrinksTo(resource, region, target, acceptor);
```
instead of the code
`createHyperlinksTo(resource, new Region(region.getOffset(), region.getLength());`
the deprecation warning is gone.

@cdietrich Could you please elaborate a little bit more what exactly concerns you about this?

cdietrich: if a user has customized createHyperlinksTo(Region) does the code still work?

miklossy: Ah, ok, I understand.

If the user has customized the createHyperlinksTo(Region) method and calls that like 
```
Region region = ...
createHyperlinksTo(region);
```
that method call will be bound to the customized createHyperlinksTo(Region) method.
```
But if the user calls the customized method like
IRegion region = ...
createHyperlinksTo(region);
```
that method call will be bound to the newly introduced (inherited) createHyperlinksTo(IRegion) method and the customized one will not be invoked.

So if we consider this, this change is indeed an API break. But the user will get a warning that a deprecated method is overridden/customized, so he(she (I hope) will understand what to do. 

@cdietrich  If you think that this PR causes more problem than it solves, then feel free to close it without merging it.

cdietrich: @miklossy i wonder if you can apply a pattern like

https://github.com/eclipse/xtext-eclipse/blob/7b4d7e44c930eb8e9cee724f983d31cd36632b15/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java#L549

to address the issue

miklossy: Thanks for the hint! Let me think about it....

cdietrich: This also applies for the default helper that was not adapted

org.eclipse.xtext.common.types.ui/src/org/eclipse/xtext/common/types/xtext/ui/JdtHyperlinkFactory.java
```diff
@@ -34,7 +35,18 @@ public boolean canHandle(EObject to) {
 		return to instanceof JvmIdentifiableElement && to.eResource() instanceof TypeResource;
 	}
 
+	/**
+	 * @deprecated use {@link #createHyperlink(IRegion, EObject, IHyperlinkAcceptor)} instead.
+	 */
+	@Deprecated
```
szarnekow: I share the concerns raised by @cdietrich - this change is hard to test for backwards compatibility since createHyperlink may both be called or overridden by clients.
In fact I would prefer the new, correct signature to delegate to the old, now deprecated signature if the type matches. Otherwise do it's own thingy.

miklossy:
Thanks for the feedback @szarnekow ! I updated the PR as suggested. The updated PR only changes the JdtHyperlinkFactory class. I think it is enough for the first step to see if we are on the right track.

Issue #1139, opened 2019-07-16
ISE: Cannot use getDescription if this dirty resource is not connected to a document


cdietrich:
we observer following ISE when deleting a project with a dirty open file

```
	java.lang.IllegalStateException: Cannot use getDescription if this dirty resource is not connected to a document
	at org.eclipse.xtext.ui.editor.DocumentBasedDirtyResource.getDescription(DocumentBasedDirtyResource.java:135)
	at org.eclipse.xtext.ui.editor.DirtyStateEditorSupport.descriptionsChanged(DirtyStateEditorSupport.java:485)
	at org.eclipse.xtext.resource.impl.AbstractResourceDescriptionChangeEventSource.notifyListeners(AbstractResourceDescriptionChangeEventSource.java:52)
	at org.eclipse.xtext.ui.notification.StateChangeEventBroker.descriptionsChanged(StateChangeEventBroker.java:20)
	at org.eclipse.xtext.resource.impl.AbstractResourceDescriptionChangeEventSource.notifyListeners(AbstractResourceDescriptionChangeEventSource.java:52)
	at org.eclipse.xtext.builder.impl.DirtyStateAwareResourceDescriptions.globalDescriptionsChanged(DirtyStateAwareResourceDescriptions.java:98)
	at org.eclipse.xtext.builder.impl.DirtyStateAwareResourceDescriptions$GlobalStateListener.descriptionsChanged(DirtyStateAwareResourceDescriptions.java:114)
	at org.eclipse.xtext.resource.impl.AbstractResourceDescriptionChangeEventSource.notifyListeners(AbstractResourceDescriptionChangeEventSource.java:52)
	at org.eclipse.xtext.builder.builderState.AbstractBuilderState.update(AbstractBuilderState.java:159)
	at org.eclipse.xtext.builder.impl.ProjectOpenedOrClosedListener.processClosedProjects(ProjectOpenedOrClosedListener.java:282)
	at org.eclipse.xtext.builder.impl.ProjectOpenedOrClosedListener$RemoveProjectsJob.runInWorkspace(ProjectOpenedOrClosedListener.java:104)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:42)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

in the debugger i can see that

```
[...]

PR #1147, opened 2019-07-22
[#1139] added better locking to document based dirty resource

cdietrich:
[#1139] added better locking to document based dirty resource

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateEditorSupport.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateEditorSupport.java
@@ -479,10 +479,15 @@ public class DirtyStateEditorSupport implements IResourceDescription.Event.Liste
 	
 	@Override
 	public void descriptionsChanged(final IResourceDescription.Event event) {
-		if (!getDirtyResource().isInitialized())
-			return;
+		DocumentBasedDirtyResource resource = getDirtyResource();
+		IResourceDescription description;
+		synchronized(resource) {
+			if (!resource.isInitialized())
+				return;
+			description = resource.getDescription();
+		}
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateEditorSupport.java
```diff
@@ -479,13 +479,16 @@ public void markEditorClean(IDirtyStateEditorSupportClient client) {
...
-		if (!getDirtyResource().isInitialized())
-			return;
-		for(IResourceDescription.Delta delta: event.getDeltas()) {
-			if (delta.getOld() == getDirtyResource().getDescription() || delta.getNew() == getDirtyResource().getDescription())
+		DocumentBasedDirtyResource resource = getDirtyResource();
+		synchronized(resource) {
+			if (!resource.isInitialized())
 				return;
+			for(IResourceDescription.Delta delta: event.getDeltas()) {
+				if (delta.getOld() == resource.getDescription() || delta.getNew() == resource.getDescription())
```
cdietrich: i am not sure if it is better to store resource.getDescription in a variable and end synchronization here

szarnekow: Yes, that would make sense indeed. The description should be a copy and it *should* be ok to work with that one in the loop without a sync block.

szarnekow:
Approved given the change you proposed within `descriptionsChanged`

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/DirtyStateEditorSupport.java
cdietrich: @szarnekow updated pr. ok now?

szarnekow:
LGTM

Issue #400, opened 2017-10-06
ProjectClasspathChangeListener / BuildScheduler / BuildJob / AutoBuild race


cdietrich:
http://services.typefox.io/open-source/jenkins/job/xtext-eclipse/job/cd_photon/4/testReport/ has failing tests e.g.

org.eclipse.xtext.builder.impl.SimpleProjectsIntegrationTest.testNewlyAddedReexportedSource 
org.eclipse.xtext.builder.impl.SimpleProjectsIntegrationTest.testReexportedJarRemoved

the same fail on my local machine (agains oxygen)
on others it does not

it seems to be a race how builds and forget builds are executed

good case:

```
Begin testNewlyAddedReexportedSource
reallyWaitForAutoBuild
reallyWaitForAutoBuild
reallyWaitForAutoBuild
reallyWaitForAutoBuild
reallyWaitForAutoBuild
reallyWaitForAutoBuild
forget foo FULL
forget bar FULL
forget baz FULL
reallyWaitForAutoBuild
forget bar FULL
build baz FULL
build foo FULL
build bar FULL
reallyWaitForAutoBuild
forget baz FULL
build baz FULL
forget baz FULL
forget baz FULL
waitForBuild 1
reallyWaitForAutoBuild
build baz FULL
doUpdate: baz
queueURI: platform:/resource/baz/src/Baz.buildertestlanguage
assert 1
reallyWaitForAutoBuild
forget foo FULL
build foo FULL
doUpdate: foo
queueURI: archive:platform:/resource/foo/foo.jar!/foo/Foo.buildertestlanguage
queueURI: platform:/resource/baz/src/Baz.buildertestlanguage
build bar AUTO
build baz AUTO
doUpdate: baz
waitForBuild 2
reallyWaitForAutoBuild
assert 2
End testNewlyAddedReexportedSource
reallyWaitForAutoBuild
doUpdate: baz
doUpdate: foo
```

bad case

```
Begin testNewlyAddedReexportedSource
reallyWaitForAutoBuild
reallyWaitForAutoBuild
reallyWaitForAutoBuild
reallyWaitForAutoBuild
reallyWaitForAutoBuild
reallyWaitForAutoBuild
forget foo FULL
forget bar FULL
forget baz FULL
reallyWaitForAutoBuild
forget bar FULL
build bar FULL
build baz FULL
build foo FULL
reallyWaitForAutoBuild
forget baz FULL
build baz FULL
forget baz FULL
[...]

PR #1156, opened 2019-08-05
[#281] Improved interaction between Xtext builder and JDT

szarnekow:
[#281] Improved interaction between Xtext builder and JDT

### DIFF
--- a/org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/IntegrationTest.java
+++ b/org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/IntegrationTest.java
@@ -14,6 +14,7 @@ import static org.junit.Assert.*;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Iterator;
 
+import org.eclipse.core.resources.IBuildConfiguration;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IMarker;
@@ -22,6 +23,7 @@ import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IStorage;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.resources.ResourceAttributes;
+import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
@@ -36,6 +38,7 @@ import org.eclipse.xtext.builder.tests.builderTestLanguage.BuilderTestLanguagePa
 import org.eclipse.xtext.resource.IReferenceDescription;
 import org.eclipse.xtext.ui.XtextProjectHelper;
 import org.eclipse.xtext.ui.resource.IResourceUIServiceProvider;
+import org.eclipse.xtext.ui.testing.util.JavaProjectSetupUtil;
 import org.eclipse.xtext.ui.testing.util.JavaProjectSetupUtil.TextFile;
 import org.eclipse.xtext.util.StringInputStream;
 import org.junit.After;
@@ -209,8 +212,6 @@ public class IntegrationTest extends AbstractBuilderTest {
 		assertEquals(1, countMarkers(bar_file));
 	}
 
-	// TODO fix https://github.com/eclipse/xtext-eclipse/issues/400
-	@Ignore("TODO fix https://github.com/eclipse/xtext-eclipse/issues/400")
 	@Test public void testTwoFilesInTwoReferencedProjectsAddNature() throws Exception {
 		foo_project = createJavaProjectWithRootSrc("foo");
 		removeNature(foo_project.getProject(), XtextProjectHelper.NATURE_ID);
@@ -285,8 +286,6 @@ public class IntegrationTest extends AbstractBuilderTest {
 		assertEquals(printMarkers(bar_file), 0, countMarkers(bar_file));
 	}
 
-	// TODO fix https://github.com/eclipse/xtext-eclipse/issues/400
-	@Ignore("TODO fix https://github.com/eclipse/xtext-eclipse/issues/400")
 	@Test public void testChangeReferencedFile() throws Exception {
 		createTwoFilesInTwoReferencedProjects();
 
@@ -301,8 +300,6 @@ public class IntegrationTest extends AbstractBuilderTest {
 		assertEquals(0, countMarkers(bar_file));
 	}
 
-	// TODO fix https://github.com/eclipse/xtext-eclipse/issues/400
-	@Ignore("TODO fix https://github.com/eclipse/xtext-eclipse/issues/400")
 	@Test public void testDeleteReferencedFile() throws Exception {
 		createTwoFilesInTwoReferencedProjects();
 
@@ -316,6 +313,86 @@ public class IntegrationTest extends AbstractBuilderTest {
 		build();
 		assertEquals(0, countMarkers(foo_file));
 		assertEquals(0, countMarkers(bar_file));
+	}
+	
+	@SuppressWarnings("restriction")
+	@Test public void testBuildOrderIsCorrect() throws Exception {
+		foo_project = createJavaProjectWithRootSrc("foo");
+		bar_project = createJavaProjectWithRootSrc("bar");
+
+		org.eclipse.core.internal.resources.Workspace workspace =
+				(org.eclipse.core.internal.resources.Workspace) ResourcesPlugin.getWorkspace();
+		IBuildConfiguration[] buildOrder = workspace.getBuildOrder();
+		assertEquals(bar_project.getProject(), buildOrder[0].getProject());
+		assertEquals(foo_project.getProject(), buildOrder[1].getProject());
+		// add a classpath entry and a project reference
+		addProjectReference(bar_project, foo_project);
+		
+		buildOrder = workspace.getBuildOrder();
+		assertEquals(foo_project.getProject(), buildOrder[0].getProject());
+		assertEquals(bar_project.getProject(), buildOrder[1].getProject());
+	}
+	
+	@SuppressWarnings("restriction")
+	@Test public void testBuildOrderIsWrong() throws Exception {
+		foo_project = createJavaProjectWithRootSrc("foo");
+		bar_project = createJavaProjectWithRootSrc("bar");
+
+		org.eclipse.core.internal.resources.Workspace workspace =
+				(org.eclipse.core.internal.resources.Workspace) ResourcesPlugin.getWorkspace();
+		IBuildConfiguration[] buildOrder = workspace.getBuildOrder();
+		assertEquals(bar_project.getProject(), buildOrder[0].getProject());
+		assertEquals(foo_project.getProject(), buildOrder[1].getProject());
+		// here we do only add a classpath entry and no core.resources project reference
+		JavaProjectSetupUtil.addProjectReference(bar_project, foo_project);
+		
+		buildOrder = workspace.getBuildOrder();
+		assertEquals(bar_project.getProject(), buildOrder[0].getProject());
+		assertEquals(foo_project.getProject(), buildOrder[1].getProject());
+	}
+	
+	@Test
+	public void testJavaChangeTriggersBuild() throws Exception {
+		foo_project = createJavaProject("foo");
+		IJavaProject zonkProject = createJavaProject("zonk");
+		bar_project = createJavaProjectWithRootSrc("bar");
+
+		addProjectReference(bar_project, zonkProject);
+		IClasspathEntry classpathEntry = JavaCore.newProjectEntry(foo_project.getPath(), true);
+		workspace.addToClasspath(zonkProject, classpathEntry);
+
+		IFolder javaPackage = foo_project.getProject().getFolder("src").getFolder("pack");
+		javaPackage.create(true, true, null);
+		IFile javaFile = javaPackage.getFile("Type.java");
+		javaFile.create(new StringInputStream("package pack; public class Type {}"), true, monitor());
+
+		IFolder dslFolder = bar_project.getProject().getFolder("src");
+		IFile dslFile = dslFolder.getFile("Dsl.typesAssistTest");
+		dslFile.create(new StringInputStream("default pack.Type"), true, monitor());
+
+		build();
+		assertEquals(printMarkers(dslFile), 0, countMarkers(dslFile));
+
+		javaFile.setContents(new StringInputStream("package pack; class X {}"), true, true, null);
+
+		build();
+		// Xtext proxy validation and EObjectValidator proxy validation
+		assertEquals(2, countMarkers(dslFile));
+
+		javaFile.setContents(new StringInputStream("package pack; public class Type {}"), true, true, null);
+
+		build();
+		assertEquals(0, countMarkers(dslFile));
+
+		workspace.removeClasspathEntry(zonkProject, classpathEntry);
+
+		build();
+		// Xtext proxy validation and EObjectValidator proxy validation
+		assertEquals(2, countMarkers(dslFile));
+
+		workspace.addToClasspath(zonkProject, classpathEntry);
+		build();
+		assertEquals(0, countMarkers(dslFile));
 	}
 
 	@Test public void testUpdateOfReferencedFile() throws Exception {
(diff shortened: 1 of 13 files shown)


szarnekow:
@cdietrich What do you think about this approach?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -561,6 +604,9 @@ private boolean isMethodSpecialized(Class<?> subclassOfBuilder, String name, Cla
 	}
 
 	/**
+	 * @param monitor the progress monitor to use for reporting progress to the user. It is the caller's responsibility
```
cdietrich: wonder if this is relevant for the new doClean too

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/JdtToBeBuiltComputer.java
```diff
@@ -8,40 +8,53 @@
 package org.eclipse.xtext.builder.impl.javasupport;
```
cdietrich: wonder if this should fix the tests in IntegrationTest and SimpleProjectsIntegrationTest that are ignored through https://github.com/eclipse/xtext-eclipse/issues/400

at least the IntegrationTest are still constantly red

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/SimpleProjectDependencyGraph.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @author Sebastian Zarnekow - Initial contribution and API
+ * 
+ * @since 2.19
+ */
+@Singleton
+public class SimpleProjectDependencyGraph {
+
+	private final Map<IProject, Set<IProject>> dependencies = Maps.newConcurrentMap();
```
cdietrich: no bidimap intentional?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -537,6 +553,33 @@ protected void addInfosFromTaskAndClean(ToBeBuilt toBeBuilt, Task task, IProgres
...
 	
+	/**
+	 * @since 2.19
+	 */
+	protected void pollQueuedBuildData() {
+		boolean needRebuild = false;
+		if (pollQueuedBuildData(getProject())) {
+			needRebuild = true;
+		}
+		for(IProject project: interestingProjects) {
```
cdietrich: interestingProjects is private. so maybe we should make it protected too or add a (final) getter

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/JdtToBeBuiltComputer.java
```diff
@@ -69,6 +82,12 @@
 	@Inject
 	private IStorage2UriMapperJdtExtensions jdtUriMapperExtension;
 	
+	/**
+	 * @since 2.19
+	 */
+	@Inject
+	private SimpleProjectDependencyGraph projectDependencyGraph;
```
cdietrich: maybe protected or add getter

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/JdtToBeBuiltComputer.java
```diff
@@ -193,4 +214,55 @@ public boolean isRejected(IFolder folder) {
...
+		try {
+			IClasspathEntry[] entries = javaProject.getExpandedClasspath();
+			for (int i = 0, l = entries.length; i < l; i++) {
+				IClasspathEntry entry = entries[i];
+				IPath path = entry.getPath();
+				IProject p = null;
+				switch (entry.getEntryKind()) {
+					case IClasspathEntry.CPE_PROJECT :
+						p = workspaceRoot.getProject(path.lastSegment()); // missing projects are considered too
+						if (((ClasspathEntry) entry).isOptional() && !JavaProject.hasJavaNature(p)) // except if entry is optional
```
cdietrich: maybe we need to calculate transitive hull / reexports here too

szarnekow: The logic is the same as for the Java builder. Could be, that we need another participant for plugin project semantics, though.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/SimpleProjectDependencyGraph.java
szarnekow: Hmm that could simplify getDependentXtextProjects. Should be changed, yes.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/JdtToBeBuiltComputer.java
szarnekow: 👍

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/JdtToBeBuiltComputer.java
szarnekow: I hope so.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
szarnekow: 👍

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/SimpleProjectDependencyGraph.java
szarnekow: The impl of `#put` becomes harder if we make this bidi. Since we need getDependentXtextProjects only when the classpath changes, I'll keep it as is.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/SimpleProjectsIntegrationTest.java
```diff
@@ -351,7 +350,7 @@ public void testReexportedSource() throws Exception {
 	}
 	
 	// TODO fix https://github.com/eclipse/xtext-eclipse/issues/400
-	@Ignore("TODO fix https://github.com/eclipse/xtext-eclipse/issues/400")
+//	@Ignore("TODO fix https://github.com/eclipse/xtext-eclipse/issues/400")
```
szarnekow: Delete commented line

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/IToBeBuiltComputerContribution.java
```diff
@@ -59,4 +62,17 @@
...
+	/**
+	 * Add all the projects to the result, that may affect the resources in this projects.
+	 * The XtextBuilder will return the collected set of projects as the interesting projects after
+	 * each build round.
+	 * 
+	 * @see IncrementalProjectBuilder#build
+	 * 
+	 * @since 2.19
+	 */
+	default void addInterestingProjects(IProject thisProject, Set<IProject> result) {
```
cdietrich: am not sure if i like this name but i dont have a better yet

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -261,7 +267,22 @@ public boolean isCanceled() {
...
-		return getProject().getReferencedProjects();
+		return getReferencedProjects();
+	}
+
+	/**
+	 * @since 2.19
+	 */
+	protected IProject[] getReferencedProjects() throws CoreException {
+		interestingProjects = toBeBuiltComputer.getInterestingProjects(getProject());
+		return interestingProjects.toArray(new IProject[0]);
```
cdietrich: unusal pattern. know that better als list.toArray(new Type[list.size])

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/ProjectClasspathChangeListener.java
```diff
@@ -59,7 +66,11 @@ public void elementChanged(ElementChangedEvent event) {
 								.filter(Predicates.notNull())
 								.transform(IJavaProject::getProject).toSet();
 						dirtyStateManager.notifyListeners(new CoarseGrainedChangeEvent());
-						scheduleBuildIfNecessary(projects);
+						Set<IProject> mutableProjects = Sets.newHashSet(projects);
```
cdietrich: i wonder why we no longer care about order here

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/TestedWorkspaceWithJDT.java
```diff
@@ -187,21 +195,49 @@ public IClasspathEntry addJarToClasspath(IJavaProject project, IFile jarFile) {
...
+	/*
+	 * This is a poor workaround since the build order is not properly updated when a 
+	 * Java project reference is added.
+	 */
+	@Deprecated
+	private void addDynamicProjectReference(IProject from, IProject to) throws CoreException {
+		IProjectDescription description = from.getDescription();
+		List<IProject> dynamicReferences = new ArrayList<>(Arrays.asList(description.getDynamicReferences()));
+		dynamicReferences.add(to);
+		description.setDynamicReferences(dynamicReferences.toArray(new IProject[0]));
```
cdietrich: IProject[dynamicReferences.size]

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/TestedWorkspaceWithJDT.java
```diff
@@ -187,21 +195,49 @@ public IClasspathEntry addJarToClasspath(IJavaProject project, IFile jarFile) {
...
+	/*
+	 * This is a poor workaround since the build order is not properly updated when a 
+	 * Java project reference is added.
+	 */
+	@Deprecated
+	private void removeDynamicProjectReference(IProject from, IProject to) throws CoreException {
+		IProjectDescription description = from.getDescription();
+		List<IProject> dynamicReferences = new ArrayList<>(Arrays.asList(description.getDynamicReferences()));
+		dynamicReferences.removeAll(Collections.singleton(to));
+		description.setDynamicReferences(dynamicReferences.toArray(new IProject[0]));
```
cdietrich: see above

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/IToBeBuiltComputerContribution.java
szarnekow: It's the term used by `IncrementalProjectBuilder.getInterestingProjects`

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
szarnekow: https://shipilev.net/blog/2016/arrays-wisdom-ancients/

szarnekow:
Please take another look @cdietrich

szarnekow:
I added a test. Please re-examine.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/IntegrationTest.java
```diff
@@ -317,6 +314,84 @@ protected void createTwoFilesInTwoReferencedProjects() throws Exception {
...
+	@Test public void testBuildOrderIsWrong() throws Exception {
+		foo_project = createJavaProjectWithRootSrc("foo");
+		bar_project = createJavaProjectWithRootSrc("bar");
+
+		org.eclipse.core.internal.resources.Workspace workspace =
+				(org.eclipse.core.internal.resources.Workspace) ResourcesPlugin.getWorkspace();
+		IBuildConfiguration[] buildOrder = workspace.getBuildOrder();
+		assertEquals(bar_project.getProject(), buildOrder[0].getProject());
+		assertEquals(foo_project.getProject(), buildOrder[1].getProject());
+		JavaProjectSetupUtil.addProjectReference(bar_project, foo_project);
```
cdietrich: having a comment that indicates that this sets a classpath ref would be nice

szarnekow: Done

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -306,9 +327,7 @@ protected void incrementalBuild(IResourceDelta delta, final IProgressMonitor mon
 		final SubMonitor progress = SubMonitor.convert(monitor, Messages.XtextBuilder_CollectingResources, 10);
 		progress.subTask(Messages.XtextBuilder_CollectingResources);
 		
-		if (queuedBuildData.needRebuild(getProject())) {
-			needRebuild();
-		}
+		pollQueuedBuildData();
```
cdietrich: @szarnekow this leads to a full build immediately following a project.build(clean)
is this intended?

cdietrich: we have deltas for projects that are no related at
all.
but still the full build gets called for the project we called clean on

cdietrich: the problematic call is the one in clean.
unfortunately 
the taskqueue ist nailed down so hard to exchange.
will try with subclassing

szarnekow: I can't follow yet.
There is a clean on some unrelated project. And then you see a full build here?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -495,6 +516,8 @@ protected boolean isOpened(IResourceDelta delta) {
 	protected void clean(IProgressMonitor monitor) throws CoreException {
 		SubMonitor progress = SubMonitor.convert(monitor, 10);
 		try {
+			pollQueuedBuildData();
```
szarnekow: So this is the problematic call?

cdietrich: yes. there is an unrelated project with deltas (on generated code)
this call leads to a needsRebuild.
the rebuild detects that the project that was cleaned has no lastcurrentbuildstate
and thus does an immediate fullbuild on the cleaned build

szarnekow: If I understand this correctly, some code / callgraph lead to unrelated projects being in the SimpleProjectDependencyGraph?

cdietrich: i dont see how simpleProjectdependencygraph is involved.
needsrebuild is globally.

cdietrich: and org.eclipse.xtext.builder.impl.javasupport.JdtQueuedBuildData.doNeedRebuild(JavaBuilderState, Procedure1<? super UnconfirmedStructuralChangesDelta>)
does not care about any graphs

szarnekow: Understood. The offending full-build trigger is not visible in this change set here but is 'legacy' from `JdtQueuedBuildData.doNeedRebuild` - we may need the simpleProjectdependencygraph and check, if the project without delta is even related to the "asking" project. Would that make sense?

cdietrich: cant we simply skip the clean part?

szarnekow: True

PR #1165, opened 2019-08-16
[eclipse/xtext#1425] show number of matches in ReferenceSearchViewPage

riederm:
The Find-References page now shows number of matches as a suffix in the
label that describes the search result.
The label is refreshed when the search finished - not after every single
hit was added to search result to avoid extensive redraws.

Signed-off-by: Mathias Rieder <mathias.rieder@gmail.com>

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/ReferenceSearchResultContentProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/ReferenceSearchResultContentProvider.java
@@ -8,11 +8,13 @@ import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.jface.viewers.ITreeContentProvider;
+import org.eclipse.jface.viewers.StructuredSelection;
 import org.eclipse.jface.viewers.TreeViewer;
 import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.search.ui.ISearchResultListener;
 import org.eclipse.search.ui.SearchResultEvent;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.TreeItem;
 import org.eclipse.ui.progress.UIJob;
 import org.eclipse.xtext.resource.IEObjectDescription;
 import org.eclipse.xtext.resource.IReferenceDescription;
@@ -197,8 +199,26 @@ public class ReferenceSearchResultContentProvider implements ITreeContentProvide
 				}
 			}
 			viewer.refresh();
-			viewer.expandToLevel(1);
+			expandToFirstChild();
 			return Status.OK_STATUS;
+		}
+	}
+	
+
+	private void expandToFirstChild() {
+		if (rootNodes != null && !rootNodes.isEmpty()){
+			ReferenceSearchViewTreeNode[] rootNodesArray = Iterables.toArray(
+					rootNodes, ReferenceSearchViewTreeNode.class);
+			//find the top element
+			viewer.getComparator().sort(viewer, rootNodesArray);
+			
+			ReferenceSearchViewTreeNode topElement = rootNodesArray[0];
+			Object[] firstChildren = topElement.getChildren().toArray();
+			
+			if (firstChildren.length >= 0) {
+				viewer.getSorter().sort(viewer, firstChildren);
+				viewer.setSelection(new StructuredSelection(firstChildren[0]), true);
+			}
 		}
 	}
 
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/Messages.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/Messages.java
@@ -19,6 +19,7 @@ public class Messages extends NLS {
 	public static String FindReferencesHandler_3;
 	public static String FindReferencesHandler_labelPrefix;
 	public static String ReferenceQuery_monitor;
+	public static String ReferenceSearchResult_Matches;
 	public static String ReferenceSearchResultContentProvider_label;
 	public static String ReferenceSearchResultLabelProvider_invalid;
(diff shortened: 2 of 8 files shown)


cdietrich:
Nice. I wonder if we should update the match count when matches are removed from the result as it is done for Java.

riederm:
Hey Christian,
of course you're right. I took a closer look at the whole remove-semantics and I was playing around with the code now for several evenings and I'm having a hard time to get this working without breaking a bunch of compatibilities. This has mainly two reasons:

1) the original code uses a ReferenceSearchResult that holds a list of IReferenceDescriptions - thats fine. IReferenceDescriptions are added to the ReferenceSearchResult and the viewer adds corresponding objects to the tree via the ReferenceSearchResultContentProvider receiving events from the search-result. So far so good - removing from the viewer is accomplished via the content-provider's internal model (ReferenceSearchViewTreeNode) - the origina SearchResult model is not touched. This has two implications: 

a) the label that shows the number of matches does not know anything about the internal model managed by the contentprovider, so from the outside the ReferenceSearchResult still contains the 
same number of search result.

b) the SearchResult stored in the history (Show previous Searches) still holds the removed matches.

so I was going to do the removes on the ReferenceSearchResult and propagate the removes again via event into the contentprovider. I also feel that the responsibilities are cleaner that way - it just means some serious changes to the contentprovider - I'm not sure if you want such API breaking changes in this ContentProvider. (e.g. some protected methods make no sense anymore, it would no longer implement the IResourceDescription.Event.Listener since the removes are propagated via the ReferenceSearchResult-

2) The second problem is the more serious problem. There are two ways hot elements are removed from the table - either the user deletes a selected item, or an item is removed because of an index-change-event and we can no longer find that particular reference. The thing is, I cannot make this index-based feature work in the original code. 

The thing is that the ReferenceFinder gives me quite different source-URIs than what the IndexChangeEvent would provide me:

- platform:/resource/org.xtext.example.mydsl/src/org/xtext/example/mydsl/MyDsl.xtext#/0/@rules.0/@type
- platform:/resource/org.xtext.example.mydsl/src/org/xtext/example/mydsl/MyDsl.xtext#/0/@rules.1/@type

 The ReferenceFinder gave me these sourceURIs:
- platform:/resource/org.xtext.example.mydsl/src/org/xtext/example/mydsl/MyDsl.xtext#//@rules.0/@alternatives/@terminal
- platform:/resource/org.xtext.example.mydsl/src/org/xtext/example/mydsl/MyDsl.xtext#//@rules.1/@alternatives/@terminal

so these URIs are quite different. The content-index in the beginning may differ, and I dont really see why the fragment would be so different? Do you have any clue for me? It seams that it has something to do with local references?

cdietrich:
I will see when I find the time to look at this.
Maybe @kthoms or @szarnekow  have an idea too

cdietrich:
hmmm i assume for the change events we need to start a new query.
they just give you which named objects changed but not which reference holder changed.

riederm:
I agree, it looks like this is necessary. This changes the feature quite a bit in a way that it not only removes invalid hits, but it would also add new hits as soon as they're built - so quite some activity in the search-view. 
do we want this?

cdietrich:
hmmm. am not sure how other searches e.g file and java behave.
if i remember correct they dont live update. you have to rerun.
but you can delete find results and the number updates.

@kthoms @szarnekow what do you two think?

szarnekow:
I don't think that we can easily make this a live view of the search results. Having this as a snapshot that updates the counter when you remove entries from the view itself (maybe adding a remark that n items have been removed from the view) sounds reasonable.

riederm:
> I don't think that we can easily make this a live view of the search results. Having this as a snapshot that updates the counter when you remove entries from the view itself (maybe adding a remark that n items have been removed from the view) sounds reasonable.

So do i get this right, that I only handle removes by the user and ignore index-changes at all? There is existing code that listens for index-changes in order to remove invalidated entries, but it does not work due to the URI-differences I explained above. Maybe it worked in earlier versions? i don't know.

cdietrich:
yes. hust handle the deletes by the user would be perfectly fine for me. if a use wants to update the search to model changes, then the search can be rexecuted. 

java search behaves the same. if you want updated results you have to reexecute the search

cdietrich:
i wonder how to continue with this pr

riederm:
my apologies for beeing so silent, i was pulled into work the last couple of weeks. I'll give this commit another try next week.

riederm:
i gave this another visit. the state of the current version is:

- number of matches are shown as requested
- removing matches updates the number of matches
- removed matches are remembered by the SearchResult so they are not added again when the query is re-run form the view's history (I find it a weird behavior, but Java acts the same, so I implemented it accordingly)

- live changes from the index are still not reflected by the search result. I cannot make the existing code work due to the different URIs as explained earlier. (do you want me to remove the old code that does not work?)

cdietrich:
Yes we should keep it minimal and working (as in java)

cdietrich:
@riederm loooks like there are compile errors now against the Oxygen target platform

riederm:
> @riederm loooks like there are compile errors now against the Oxygen target platform

whoops, some java11 methods slipped in

cdietrich:
@kthoms @ArneDeutsch what do you think?

cdietrich:
ping @kthoms @ArneDeutsch

cdietrich:
@kthoms what do you think?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/findrefs/ReferenceSearchResultContentProvider.java
```diff
@@ -197,10 +199,28 @@ public IStatus runInUIThread(IProgressMonitor monitor) {
...
 			viewer.refresh();
-			viewer.expandToLevel(1);
+			expandToFirstChild();
 			return Status.OK_STATUS;
 		}
 	}
+	
+
+	private void expandToFirstChild() {
+		if (rootNodes != null && !rootNodes.isEmpty()){
```
kthoms: personally I prefer fast returns instead of an if-block that spans the whole method body

cdietrich:
thx for the contribution @riederm

PR #1172, opened 2019-09-03
[releng] Set version to 2.20.0

kthoms:


org.eclipse.xtext.activities/pom.xml
```diff
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.eclipse.xtext</groupId>
 		<artifactId>org.eclipse.xtext.tycho.plugin.parent</artifactId>
-		<version>2.19.0-SNAPSHOT</version>
+		<version>2.20.0-SNAPSHOT</version>
```
szarnekow: Unrelated question: Given that we have a relative path to the parent - shouldn't it be possible to maintain the version only there?

cdietrich: afaik the parent version is needed

szarnekow: Hmm that's bad.

Looks like the fancy "new" thing `${revision}` is also not very nice (https://maven.apache.org/maven-ci-friendly.html)

kthoms: That's one of the drawbacks of Maven, unfortunately. I also wish that the parent's version could be obsolete, would make many things easier.
But when you think of deployed POMs it becomes obvious: Suppose you try to retrieve an artifact which inherits the parent's version and uses the `project.parent.version` variable. When there is only a relative path and you download the artifact, it is unclear which artifact should be downloaded as the parent POM.

szarnekow: My assumption was, that the version number could be added / would be added automatically before / during signing / deployment.

Issue #1182, opened 2019-09-20
Content Assist Tests in Eclipse and Xtend fail with Java 13


cdietrich:
we have Content Assist Tests fails in Eclipse and Xtend with Java 13 that i dont know how to properly fix. i have the feeling the content assist test framework is a  bit bogus when it comes to methods vs getters.

@szarnekow it would be nice if you somewhen could find some time to have a look.

e.g

org.eclipse.xtext.xbase.ui.testing.AbstractXbaseContentAssistTest.testNestedTypes_01()
and
org.eclipse.xtext.xbase.ui.testing.AbstractXbaseContentAssistTest.testStaticFeatures_02()

fail

the stuble over the componentType vs getComponentType

cdietrich:
Fixed in 2.20

PR #1184, opened 2019-09-20
[eclipse/xtext-eclipse#1182] fixed some test fails with java 13

cdietrich:
[eclipse/xtext-eclipse#1182] fixed some test fails with java 13

### DIFF
--- a/org.eclipse.xtext.xbase.ui.testing/src/org/eclipse/xtext/xbase/ui/testing/AbstractXbaseContentAssistTest.java
+++ b/org.eclipse.xtext.xbase.ui.testing/src/org/eclipse/xtext/xbase/ui/testing/AbstractXbaseContentAssistTest.java
@@ -14,6 +14,8 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.emf.common.util.URI;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.xbase.ui.testing/src/org/eclipse/xtext/xbase/ui/testing/AbstractXbaseContentAssistTest.java
```diff
@@ -53,6 +55,82 @@
  */
 public abstract class AbstractXbaseContentAssistTest extends Assert implements ResourceLoadHelper, IJavaProjectProvider {
 
+	private static final boolean isJava11OrLater = determineJava11OrLater();
+
```
LorenzoBettini: In general, I think these utility methods might be useful also in other tests, why not make them available somewhere?

org.eclipse.xtext.xbase.ui.testing/src/org/eclipse/xtext/xbase/ui/testing/AbstractXbaseContentAssistTest.java
```diff
@@ -53,6 +55,82 @@
...
+	private static boolean determineJava13OrLater() {
+		String javaVersion = System.getProperty("java.version");
+		try {
+			Pattern p = Pattern.compile("(\\d+)(.)*");
+			Matcher matcher = p.matcher(javaVersion);
+			if (matcher.matches()) {
+				String first = matcher.group(1);
+				int version = Integer.parseInt(first);
+				return version >= 13;
```
LorenzoBettini: If I'm not wrong, these 3 methods are exactly the same but for the constant after `>=`... I guess it would be nice to have only one method taking the Java version as a parameter

org.eclipse.xtext.xbase.ui.testing/src/org/eclipse/xtext/xbase/ui/testing/AbstractXbaseContentAssistTest.java
cdietrich: see https://github.com/eclipse/xtext/issues/1452

org.eclipse.xtext.xbase.ui.testing/src/org/eclipse/xtext/xbase/ui/testing/AbstractXbaseContentAssistTest.java
cdietrich: see https://github.com/eclipse/xtext/issues/1452

Issue #1217, opened 2019-10-14
Domainmodel example: Outline view improvements


miklossy:
Given is the following `dmodel` file:

```
import java.util.List
import java.util.LinkedList

entity A {
	a : String
	b : String[]
	op foo(List<? extends String> p) : String {
		return a
	}
}

entity B extends LinkedList<String> {
	
}

entity C extends LinkedList<String> {
	a : String
}
```

Currently, the outline view of this file looks like:
![screenshot](https://user-images.githubusercontent.com/20393472/66754030-5d0a3a80-ee95-11e9-90f4-89ffd29f8309.png)

I suggest the following outline view improvements:
- do not create any outline view nodes for the `Jvm Parameterized Type Reference` elements
- use `B extends LinkedList<String>` text instead of the `B`
- use styled String after the `:` similar to Xtend

With these modifications the outline view could look like:
![screenshot2](https://user-images.githubusercontent.com/20393472/66754223-cb4efd00-ee95-11e9-9ca4-4bb382c5a87e.png)

miklossy:
Fixed in 2.20.

PR #1218, opened 2019-10-14
[#1217] Domainmodel example: Outline view improvements.

miklossy:
- Modify the implementation of the DomainmodelLabelProvider and the
DomainmodelOutlineTreeProvider to not to create any outline view nodes
for the Jvm Parameterized Type Reference elements, to use B extends
LinkedList<String> text instead of the B label, to use styled String
after the : similar to Xtend.
- Implement corresponding OutlineTest test cases.


org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.ui/src/org/eclipse/xtext/example/domainmodel/ui/labeling/DomainmodelLabelProvider.xtend
```diff
@@ -53,15 +55,27 @@ class DomainmodelLabelProvider extends XbaseLabelProvider {
 		return super.doGetImage(element)
 	}
 
-	def String text(Property property) {
+	def Object text(Entity entity) {
```
cdietrich: why object

miklossy: Because of the StyledText

PR #1239, opened 2019-10-29
[no-ticket] Fixed warnings

szarnekow:
[eclipse/xtext#1518] Fixed warning about missing entries in MANIFEST.MF
[no-ticket] Fixed warning about missing bundles from TP and bogus output directory


szarnekow:
Ping @kthoms - please review

PR #1243, opened 2019-11-05
added option to skip legacy platform uri map calculation

cdietrich:
added option to skip legacy platform uri map calculation


cdietrich:
it gives advanced users the chance to avoid the legacy calls to EcorePlugin.computePlatformURIMap if the are sure its not needed in their usecase / product.

i hesitated to document this but just want to give an option to achieve this without c&p the whole class

szarnekow:
I was thinking about making it a little bit more flexible, e.g. having something like `addEntriesFromPlatformResourceMap(Map<..>)` where a specialised impl can do a little more. But in the end the candidate to override for this usecase would be the other method I suppose

PR #1247, opened 2019-11-06
[eclipse/xtext#1573] Provide proper build.properties

szarnekow:
[eclipse/xtext#1573] Provide proper build.properties

### DIFF
--- a/org.eclipse.xtext.builder/build.properties
+++ b/org.eclipse.xtext.builder/build.properties
@@ -2,7 +2,7 @@ source.. = src/,\
            emf-gen/,\
            xtend-gen/,\
            deprecated/
-output.. = target/classes/
+output.. = bin/
 bin.includes = META-INF/,\
                .,\
                plugin.xml,\
--- a/org.eclipse.xtext.builder/build.properties
+++ b/org.eclipse.xtext.builder/build.properties
@@ -2,6 +2,7 @@ source.. = src/,\
            emf-gen/,\
            xtend-gen/,\
            deprecated/
+output.. = target/classes/
 bin.includes = META-INF/,\
                .,\
                plugin.xml,\
--- a/org.eclipse.xtext.xbase.junit/build.properties
+++ b/org.eclipse.xtext.xbase.junit/build.properties
@@ -1,7 +1,7 @@
 source.. = src/,\
            testdata/,\
            xtend-gen/
-output.. = target/classes/
+output.. = bin/
 bin.includes = META-INF/,\
                .,\
                about.html,\
--- a/org.eclipse.xtext.xbase.junit/build.properties
+++ b/org.eclipse.xtext.xbase.junit/build.properties
@@ -1,6 +1,7 @@
 source.. = src/,\
            testdata/,\
            xtend-gen/
+output.. = target/classes/
 bin.includes = META-INF/,\
                .,\
                about.html,\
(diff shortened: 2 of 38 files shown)


org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/XtextHyperlinkHelper.java
```diff
@@ -76,7 +77,7 @@ protected void createLinksToBase(ITextRegion nameLocation,	AbstractRule rule, IH
 				URI uri = uriConverter.normalize(EcoreUtil.getURI(overwritten));
 
 				XtextHyperlink result = getHyperlinkProvider().get();
-				result.setHyperlinkRegion(toRegion(nameLocation));
+				result.setHyperlinkRegion((IRegion)toRegion(nameLocation));
```
cdietrich: change intended?

szarnekow: Yes. I wanted to get rid of the deprecation warning.

cdietrich: i have the feeling we kept it the old way back then for a reason.

cdietrich: https://github.com/eclipse/xtext-eclipse/pull/1146

szarnekow: I changed it in a different way to get rid of the warning.

org.eclipse.xtext.builder/build.properties
```diff
@@ -2,6 +2,7 @@ source.. = src/,\
            emf-gen/,\
            xtend-gen/,\
            deprecated/
+output.. = target/classes/
```
ArneDeutsch: I think this should be "bin" ... otherwise it will probably clash between maven and eclipse build.

org.eclipse.xtext.xbase.junit/build.properties
```diff
@@ -1,6 +1,7 @@
 source.. = src/,\
            testdata/,\
            xtend-gen/
+output.. = target/classes/
```
ArneDeutsch: Again I would expect "bin" here.

org.eclipse.xtext.builder/build.properties
szarnekow: So we need to change the classpath / project configuration? Right now it compiles to target/classes in eclipse.

cdietrich: do we have any concept for that?
are the tycho recommendations for that?
separate/same output folder for eclipse and build?

ArneDeutsch: To my knowledge it is best to seperate eclipse and maven output folders. Otherwise mamven cleans eclipse class files and IDE is not so happy.

cdietrich: hmm the current config is rather mixed

szarnekow: I'm on it

szarnekow: Done

org.eclipse.xtext.xbase.junit/build.properties
szarnekow: Done

szarnekow:
@ArneDeutsch Can you please take another look at this?

PR #1253, opened 2019-11-09
removed deps to old generator

cdietrich:
removed deps to old generator


org.eclipse.xtext.common.types.eclipse.tests/META-INF/MANIFEST.MF
```diff
@@ -17,7 +17,6 @@ Require-Bundle: org.eclipse.xtext.junit4,
  org.eclipse.jdt.launching;bundle-version="3.9.100",
  org.eclipse.xtext.common.types.ui,
  org.eclipse.emf.mwe2.launch;bundle-version="2.9.1";resolution:=optional;x-installation:=greedy,
```
szarnekow: mwe2 launch - shouldn't this be moved to the additional.bundles section, too?

org.eclipse.xtext.ui.tests/build.properties
```diff
@@ -13,4 +13,6 @@ bin.includes = model/generated/,\
                icons/,\
                testfiles/
 src.includes = about.html
+additional.bundles = org.eclipse.emf.mwe2.launch,\
```
szarnekow: Looks like the handling of the mwe2 dependency is not consistent yet.

org.eclipse.xtext.common.types.eclipse.tests/META-INF/MANIFEST.MF
cdietrich: might be some tests that use some classes transitively at runtime.
need to check that

org.eclipse.xtext.ui.tests/build.properties
cdietrich: fixed

cdietrich:
@szarnekow please reexamine

PR #1254, opened 2019-11-11
make sure xtext.xtext.ui can be activated without xpand/xtend1 present

cdietrich:
make sure xtext.xtext.ui can be activated without xpand/xtend1 present


cdietrich:
@szarnekow what do you think?

Issue #251, opened 2017-05-02
Get rid of EGit workaround in XtextBuilder.MAKE_EGIT_JOB_SYSTEM


cdietrich:
XtextBuilder.MAKE_EGIT_JOB_SYSTEM 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=459525 has a workaround that should be no longer be needed.
we rarely run into this trace

[...]

PR #1255, opened 2019-11-13
[#251] Remove workaround for bug#474003

kthoms:
The EGit IndexDiffUpdateJob interrupted Xtext builds, which made the
workaround necessary. EGit was fixed with Eclipse Mars, which is below
minimum target requirement now. This makes the workaround obsolete.

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
@@ -7,14 +7,12 @@
  *******************************************************************************/
 package org.eclipse.xtext.builder.impl;
 
-import java.lang.reflect.Field;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 
 import org.apache.log4j.Logger;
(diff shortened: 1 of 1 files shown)


cdietrich:
i am ok with this. @szarnekow what do you think.
@kthoms we also should document the minimal egit version in the release notes

kthoms:
I've added this to the release notes:
```
### EGit minimal version

Xtext requires to use EGit 4.1 now. We have removed a workaround that prevented an endless build loop caused by EGit [Bug#474003](https://bugs.eclipse.org/bugs/show_bug.cgi?id=474003). Since we expect that no clients are still using Xtext >= 2.20 together with the ancient EGit version, we felt safe to finally remove this workaround.
```
@szarnekow fine for you now?

szarnekow:
Can we somehow ensure that Xtext does not run with an ancient EGit?

kthoms:
Not really. When you try to install Xtext 2.20 you'll need at least Oxygen.3a. This is constrained. When you install EGit on top of that, you'll use either the SimRel repo or Git release repo and get a newer version. I think when you actively compose a target with Oxygen or newer, Xtext 2.20 or newer and EGit 4.0 or older, then this might be possible. I think this together is really unlikely.

cdietrich:
i dont think we can.

Issue #1259, opened 2019-11-17
Xtext Wizard: Add API to be able to use images in the template descriptions.


miklossy:
The new project wizard API should be extended to be able to register images (that can be used on the new project wizard template descriptions) on the `org.eclipse.ui.forms.widgets.FormText` Widget.

The Xtext Statemachine Example should also be extended by the new project wizard / new file wizard to demonstrate the usage of the API.

miklossy:
Fixed in 2.20

PR #1260, opened 2019-11-17
[#1259] Statemachine example: add New Project Wizard/New File Wizard.

miklossy:
- Extend the
org.eclipse.xtext.ui.wizard.template.AbstractProjectTemplate class by
the List<Pair<String, Image>> getImages() API method to be able to
register images that can be used in the template descriptions.
- Add new project wizard / new file wizard to the Xtext Statemachine
Example to demonstrate the usage of the API.


cdietrich:
cc @ArneDeutsch

PR #1268, opened 2019-11-20
added missing mwe.core lib

cdietrich:
added missing mwe.core lib


org.eclipse.xtext.xtext.ui/META-INF/MANIFEST.MF
```diff
@@ -15,6 +15,7 @@ Require-Bundle: org.eclipse.xtext.ui,
  org.eclipse.xtext.ui.shared,
  org.eclipse.ui.editors;bundle-version="3.11.0",
  org.eclipse.ui;bundle-version="3.109.0",
+ org.eclipse.emf.mwe.core;resolution:=optional;x-installation:=greedy,
```
cdietrich: why does tycho / xtend maven plugin not do this automatically

cdietrich: @szarnekow  I also wonder why here is no explicit dep to Xtext

PR #1277, opened 2019-11-26
[#491694] Statemachine example: bind Sample Ecore Editor.

miklossy:
- Extend the plugin.xml of the Xtext Statemachine example with the
corresponding editorContentTypeBinding to be able to open the
*.statemachine files with the Sample Ecore Editor.

Bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=491694

### DIFF
--- a/org.eclipse.xtext.xtext.ui.examples/projects/fowlerdsl/org.eclipse.xtext.example.fowlerdsl.ui/plugin.xml
+++ b/org.eclipse.xtext.xtext.ui.examples/projects/fowlerdsl/org.eclipse.xtext.example.fowlerdsl.ui/plugin.xml
@@ -11,6 +11,10 @@
 			id="org.eclipse.xtext.example.fowlerdsl.Statemachine"
 			name="Statemachine Editor">
 		</editor>
+		<editorContentTypeBinding
+			contentTypeId="org.eclipse.xtext.example.fowlerdsl.Statemachine.contenttype"
+			editorId="org.eclipse.emf.ecore.presentation.EcoreEditorID">
+		</editorContentTypeBinding>
 	</extension>
 	<extension
 		point="org.eclipse.ui.handlers">


miklossy:
This PR contains the corresponding editor binding of the Xtext Statemachine Example to be able to open *.statemachine files with the Sample Ecore Editor. @kthoms Please review, if you also agree that this is the right way to go, I will provided a follow up PR to modify the generator to generate the binding.

szarnekow:
Please check if the ecore editor does work with languages that have jvm model inferrer before changing the generator.

cdietrich:
AFAIK it usually does not work

miklossy:
@szarnekow Thanks for the hint! @cdietrich is right, the Sample Ecore Model Editor cannot be opened on DSL files having JVM Model Inferrer (e.g. Xtext Domainmodel Example)
screenshot

miklossy:
Since it does not work for all Xtext-based languages, I suggest to add a generator option (that is set to false by default) to generate the corresponding editor binding.

szarnekow:
I have to object this. Assuming it was set to false by default, the plugin.xml will be generated without the extension. Switching the option to true afterwards, will only change the plugin.xml_gen. The user has to adjust the plugin.xml manually. That is the same task that needs to be done, if we do not add such an option to the generator.

miklossy:
I see a difference: if the user checked in the `plugin.xml_gen` file than the generator (when the option is set to true) will produce the same content, without any changes.

szarnekow:
> I see a difference: if the user checked in the `plugin.xml_gen` file than the generator (when the option is set to true) will produce the same content, without any changes.

Sorry, I probably miss your point. Is your line of thought, that we should add and maintain a generator option because some people don't want to diff their plugin.xml with the plugin.xml_gen?

miklossy:
I think that adding a generator option would make it more explicit why the `plugin.xml_gen` file has been changed. But I will be also satisfied with a solution to generate that additional binding always (without any option) as you suggested.

szarnekow:
No, I did not suggest to always generate the extension. I don't want the option and I don't want to generate the extension that is likely to suggest something, that does not work in the general case. Users who want to use the reflective editor and offer it as part of their officially supported editors for the DSL are free to add the extension to their plugin.xml files.

miklossy:
But the users often not even know that adding such a binding is possible. That's why a generator option could help.

szarnekow:
I don't think that uses would check the mwe2 generator options to figure out, what plugin.xml extensions are possible.

cdietrich:
@miklossy i would prefer not to have a generator option for everthing.
but this is a perfect candidate for an (english language) technical blog post

miklossy:
OK, I agree with @cdietrich will close this PR and will create a technical blog about this topic. 
Thank you both for your time for the feedback!

PR #1279, opened 2019-11-27
[#https://github.com/eclipse/xtext-xtend/issues/789] Improve Xbase UI

miklossy:
- Modify the Xbase launching infrastrukture to provide possibilities to
replace selection on an IPackageFragment element using the
JavaElementDelegate.getJavaElementForPackage(IPackageFragment) method.

### DIFF
--- a/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/launching/JavaElementDelegate.java
+++ b/org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/launching/JavaElementDelegate.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2011, 2020 itemis AG (http://www.itemis.eu) and others.
  * All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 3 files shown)


org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/launching/JavaElementDelegate.java
```diff
@@ -224,6 +234,14 @@ protected IJavaElement getJavaElementForResource(IResource resource) {
 		}
 		return null;
 	}
+	
+	/**
+	 * Subclasses may override.
+	 * @since 2.21 
+	 */
+	protected IJavaElement getJavaElementForPackage(IPackageFragment packageFragment) {
+		return null;
```
szarnekow: Why does this return null?

miklossy: I have chosen `null` as the return value to indicate that the Xbase framework does nothing by default and to emphasize that the subclasses should override that method to provide a meaningful return value.

cdietrich:
@miklossy @szarnekow  how to proceed here?

miklossy:
I would like to merge this as soon as possible and collect feedback from the community afterwards.

Issue #1188, opened 2019-09-23
Please add an option to disable build workspace before rename Element


lidochca:
Hello, 

I had a discussion with Christian on xtext-dev mailing list about this topic
https://www.eclipse.org/lists/xtext-dev/msg01335.html 

We found that when we try to user 'Rename element' for our dsl files, it triggers workspace build. Our projects has a custom builder that executes a shell script that is time consuming.
It makes 'Rename Element' impossible to use.
We would need an option in preferences to disable/to avoid building the whole workspace when 
user tries to 'Rename Element'.
Just customizing our builder to avoid it's execution when nothing has changed, is not a solution for us.
We have project with mixed sources CPP/python/other custom languages, whose configuration paths and other files could change outside eclipse, by running an external builder, e.g. a ccmake from command line,  and when we trigger a build from eclipse, it still should happen event if for eclipse nothing has changed.
 
A simple option that would make possible to ovoid building the whole workspace would be great.
Something in SyncUtil.java like 

public void waitForBuild(IProgressMonitor monitor) {
try { 
      if ( workspace_build_enabled_during_rename) {
                workspace.build(IncrementalProjectBuilder.INCREMENTAL_BUILD, monitor);
      }
} catch (CoreException e) {
      throw new OperationCanceledException(e.getMessage());
}
}
Could you please implement it in next xtext releases?

Thank you in advance.
Kind Regards
Lidia

kthoms:
I'm strongly against a visible preference, since this requirement is extremely rare and unintended usage usually leads to an undesired behavior. 

However, preferences don't need to have a visible possibility for configuration in the workspace preferences. There are a couple of examples in Eclipse platform code where preferences are internally visible only. This might require some advanced knowledge of their presence and a custom provisioning of a workspace/project/bundle settings.

> Could you please implement it in next xtext releases?

I doubt that we will work on this soon, but we are open for contributions.

lidochca:
Ok Thank you Karsten,
It is an acceptable solution for us. I can contribute with something like this:

IPreferenceStore prefs = Activator.getDefault().getPreferenceStore();
Boolean isDisabled = prefs.getBoolean(KEY_DISABLE_BUILD_DURING_RENAME);

And we can set the desired value in config.ini

Kind Regards
Lidia

cdietrich:
i am very sceptic about that.
doing it as preference is not an option imho
maybe we can do a hidden secrect undocumented jvm option but i am very sceptic about that.

[...]

PR #1281, opened 2019-12-02
[eclipse#1188] Adding a jvm option that skips build before rename

lidochca:
- In some cases when the user tries to rename an element, a build in
advance is not desirable.
- Added to be able to skip build workspace before rename if jvm option
is used and set to true, like -Dxtext.skipWaitForBuild=true.

Signed-off-by: Lidia Popescu <lidia.popescu@windriver.com>

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/ui/SyncUtil.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/ui/SyncUtil.java
@@ -40,6 +40,7 @@ import com.google.inject.Inject;
  */
 public class SyncUtil {
 	private static final Logger LOG = Logger.getLogger(SyncUtil.class);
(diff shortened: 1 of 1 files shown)


cdietrich:
@szarnekow @kthoms what do you think?
https://github.com/eclipse/xtext-eclipse/issues/1188

szarnekow:
I'm anticipating bug reports that say `rename is broken` and we'll waste a lot of time figuring out that his option may have been set.

cdietrich:
i also still wonder why this cannot be fixed by customizing the places that call the code.
i also wonder if we can add a comment stating that this is a unsupported feature

szarnekow:
@lidochca Thank you for the suggested enhancement. Your feedback is very valuable but as you see from the discussion, we are not sure if this is the best thing to do on the framework level.
Did it work for you to patch this logic in your project specifically?

szarnekow:
I'm inclined to close this without merging. Please let me know if that would be a blocker for you.

szarnekow:
Please reopen if the problem cannot be fixed locally.

PR #1296, opened 2019-12-13
[#1295] Don't validate files from mvn output folder

### LINKED ISSUE #1295: Don't create markers on files in target folder

szarnekow:
Don't validate files from mvn output folder even if the eclipse project has a different output folder configured.

closes #1295


cdietrich:
@szarnekow  i miss a change to the parent pom

szarnekow:
> @szarnekow i miss a change to the parent pom

I was looking for the right poms to be modified when doing this. Which parent do you mean?

szarnekow:
Got it

org.eclipse.xtext.buildship/src/org/eclipse/xtext/buildship/IgnoreBuildDirContribution.java
```diff
@@ -7,35 +7,44 @@
  *******************************************************************************/
 package org.eclipse.xtext.buildship;
 
+import java.io.File;
 import java.util.Collections;
+import java.util.Optional;
 
-import org.eclipse.buildship.core.internal.CorePlugin;
-import org.eclipse.buildship.core.internal.preferences.PersistentModel;
+import org.eclipse.buildship.core.GradleBuild;
```
cdietrich: we should also make sure we use buildship 3.x in the manifest
and also do the same pom thing we do for codemining
(have to find out which release has buildship 3.x)

szarnekow: Done

szarnekow:
I think this is ready to be merged. I'd like to proceed with the preferences and project configs. Any objections?

Issue #1228, opened 2019-10-19
Xtext Wizard Fragment: no file headers are generated


miklossy:
When using the Xtext project wizard / file wizard fragments in the mwe2 workflow, the generated `Messages.java` class does not contain any file headers, even if it is configured previously.

When extending the Xtext FowlerDsl Example by the 
```
projectWizard = {
	generate = true
}
```
configuration, the generated `Messages.java` class does not contain the configured file header
```
/*
 * generated by Xtext
 */
```
![screenshot](https://user-images.githubusercontent.com/20393472/67143212-5995fb00-f269-11e9-84bf-9bcba1ac7898.png)

@ArneDeutsch Could you please take a look? Thanks a lot in advance!

cdietrich:
@ArneDeutsch i have no idea where this is generated. maybe some eclipse magic?

cdietrich:
ahh its the ProjectTemplateProcessor
which is an active annotation processor

ArneDeutsch:
I have an solution but am not sure if we should implement.
The "Messages.java" is not generated by the generator but by an ActiveAnnotation in the generated code. In result it is not possible to read the configured value from the mwe2 file. The usesr could also place the ActiveAnnotation at some other class that is not linked to mwe2 at all (just to make it clear that there is really no connection possible).

A solution/workaround would be to copy the file header comment from the original Xtend file. In our case this would be the desired effect, the generated "Messages.java" would contain the configured file header comment. Implemented this and it works. But there is a drawback. The generated "Messages.java" is potentially generated from two or more classes (FileTemplateProvider and ProjectTemplateProvider in our case). If the user modifies the comment in one of these classes the "Messages.java" would contain this modified comment. And if the user then modifies the other file, the comment in "Messages.java" would toggle. This might lead to unwanted dirty flags in version control. I tend towards accepting the missing header comment and do not implement this. Your opinion?

cdietrich:
for me this would not be a workaround but the correct solution. we might also add a string param with the header to the aa which would be another option

cdietrich:
@szarnekow @miklossy  what do you think

szarnekow:
The `Messages.java` should get the header of the class that contains the processed annotations. So I'm on the same page with Christian and Arne's proposal

ArneDeutsch:
Not sure if you get my comment completely. Just to make this crystal clear:
[...]

PR #1313, opened 2020-01-09
[#1228] Statemachine example: Regenerate the Messages.java class.

miklossy:


cdietrich:
i dont have any changes locally. is the a generate once then update manually resource?

cdietrich:
even if i delete it i dont get any comment

cdietrich:
i also have the feeling that the n:1 generation does not really work with incremental builds

miklossy:
I deleted the `Messages.java` file and executed a `Project -> Clean...` After that the Messages.java file has been generated with the copyright header.
screencast

cdietrich:
for me not. i wonder what the active annotation does and which file it picks to read the header in my and in your case

cdietrich:
hmm now i am getting it.

Issue #1317, opened 2020-01-14
Open Model Element Dialog: label for matches is cut


cdietrich:
Open Model Element Dialog: label for matches is cut

![Bildschirmfoto vom 2020-01-14 16-05-16](https://user-images.githubusercontent.com/238736/72355447-c9404880-36e7-11ea-9d4c-1ecd8a5caa4a.png)

miklossy:
On Windows it looks like fine:
![image](https://user-images.githubusercontent.com/20393472/72420846-b6795280-377f-11ea-9c7e-affb0f57a9bf.png)

cdietrich:
hmm i see the problem on different linux machines. will check on mac in the evening

miklossy:
Fixed in 2.21.

PR #1318, opened 2020-01-15
[#1317] Improve GUI Layout on the Open Model Element Dialog.

miklossy:
- Ensure that the matching elements label is made wide enough to fit the
entire horizontal space.


cdietrich:
:+1: searching label looks good now too

PR #1319, opened 2020-01-16
Add hashCode() and equals() to ReconcilerReplaceRegion

mmews-n4:
see https://github.com/eclipse/xtext-core/pull/1355

Mind that the field `ReconcilerReplaceRegion#events` of type `List<DocumentEvent>` is not part of hashCode and equals methods (yet?). Reason is that I would like to hear your opinion on this first since `org.eclipse.jface.text.DocumentEvent` does not provide `hashCode()` or `equals()` methods.


cdietrich:
thx @mmews-n4

PR #1320, opened 2020-01-16
Removed deprecated API and fixed a warning

davidmedinasigasi:
Signed-off-by: David Medina <david.medina@sigasi.com>


miklossy:
The `org.eclipse.jface.viewers.StructuredViewer.setSorter(ViewerSorter sorter)` method is deprecated, as its javadoc says:
`@deprecated use <code>setComparator()</code> instead.`

cdietrich:
thx @davidmedinasigasi

PR #1321, opened 2020-01-20
Some cosmetic changes to Domainmodel validator

LorenzoBettini:
This PR aims at cleaning up the Domainmodel validator after the porting to Java (from Xtend):

- removed useless parenthesis, curly brackets, parameter types in lambdas
- turned most uses of IterableExtensions (nested calls) to Java streams (fluent style)

I left a use of IterableExtensions but we might change that as well.
Moreover, we can further remove local variable declarations that are then used for streams (but I'd like to have feedback about that first)


LorenzoBettini:
@miklossy I've updated the year, but please check whether it's as expected: listing single years is right?

miklossy:
Instead of `2016, 2019, 2020` I prefer to use `2016, 2020`

cdietrich:
first, last is the recommendation of the foundation for ranges
https://www.eclipse.org/projects/handbook/#ip-copyright-headers

LorenzoBettini:
@miklossy @cdietrich updated copyright and rebased

miklossy:
Thanks @LorenzoBettini ! Please merge the PR.

PR #1335, opened 2020-02-10
Fixed Editor Search in UNDO

cdietrich:
Fixed Editor Search in UNDO

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/impl/EditorDocumentUndoChange.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/impl/EditorDocumentUndoChange.java
@@ -66,7 +66,7 @@ public class EditorDocumentUndoChange extends Change {
 			if (editor != null && Objects.equal(editor.getSite().getId(), editorID))
 				return (ITextEditor) editor;
 			else
-				return (ITextEditor) page.openEditor(editorInput, editorID);
+				return (ITextEditor) page.openEditor(editorInput, editorID, true, IWorkbenchPage.MATCH_INPUT | IWorkbenchPage.MATCH_ID);
 		} catch (Exception exc) {
 			LOG.error("Error restoring editor", exc);
 			return null;


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/impl/EditorDocumentUndoChange.java
```diff
@@ -66,7 +66,7 @@ protected ITextEditor getEditor() {
 			if (editor != null && Objects.equal(editor.getSite().getId(), editorID))
 				return (ITextEditor) editor;
 			else
-				return (ITextEditor) page.openEditor(editorInput, editorID);
+				return (ITextEditor) page.openEditor(editorInput, editorID, true, IWorkbenchPage.MATCH_INPUT | IWorkbenchPage.MATCH_ID);
```
cdietrich: am not sure if activate should be true

szarnekow:
How can I reproduce the bug?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/impl/EditorDocumentUndoChange.java
szarnekow: I think `activate==true` is correct here since we want to provide some context for the user.

cdietrich:
if you have mutile editors then searching needs to be done on input and id to get the text editor here and not the editor on the same/file input only

cdietrich:
we currently call this on

```
	@Override
	public IEditorPart openEditor(IEditorInput input, String editorID) throws PartInitException {
		return openEditor(input, editorID, true, MATCH_INPUT);
	}
```

only

cdietrich:
@rytina could you please create a bug?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/refactoring/impl/EditorDocumentUndoChange.java
szarnekow: > if you have mutile editors then searching needs to be done on input and id to get the text editor here and not the editor on the same/file input only

I see. Makes sense. It's a side effect of not properly using the right document type. Good catch.

rytina:
> 
> 
> @rytina could you please create a bug?

Done https://github.com/eclipse/xtext-eclipse/issues/1340

Issue #1208, opened 2019-10-08
EGit Merge tool doesn't work with XtextEditor


rheydenr:
Steps to reproduce:

- create an Xtext project (with at least one Xtext document) and add it to a Git repository
- on master branch: change the Xtext document and commit it
- switch to another branch (cloned from master) and change the same Xtext document (with another modification which should provoke a merge conflict with master branch's file), commit it
- merge branch master into the current branch so that a merge conflict occurs
- use *Team -> Merge Tool* on that Xtext document
The following exception is thrown:
```
Caused by: org.eclipse.ui.PartInitException: Editor could not be initialized.
	at org.eclipse.ui.texteditor.AbstractTextEditor.internalInit(AbstractTextEditor.java:3188)
	at org.eclipse.ui.texteditor.AbstractTextEditor.init(AbstractTextEditor.java:3197)
	at org.eclipse.xtext.ui.editor.XtextEditor.init(XtextEditor.java:305)
	at org.eclipse.xtext.ui.compare.DefaultMergeViewer.createSourceViewerConfiguration(DefaultMergeViewer.java:190)
	... 62 more
Caused by: java.lang.IllegalArgumentException: document may not be null
	at org.eclipse.xtext.ui.editor.DocumentBasedDirtyResource.connect(DocumentBasedDirtyResource.java:36)
	at org.eclipse.xtext.ui.editor.DirtyStateEditorSupport.initDirtyResource(DirtyStateEditorSupport.java:448)
	at org.eclipse.xtext.ui.editor.DirtyStateEditorSupport.initializeDirtyStateSupport(DirtyStateEditorSupport.java:390)
	at org.eclipse.xtext.ui.editor.XtextEditor.initializeDirtyStateSupport(XtextEditor.java:278)
	at org.eclipse.xtext.ui.editor.XtextEditor.doSetInput(XtextEditor.java:260)
	at org.eclipse.xtext.ui.compare.DefaultMergeEditor.doSetInput(DefaultMergeEditor.java:43)
	at org.eclipse.ui.texteditor.AbstractTextEditor$5.run(AbstractTextEditor.java:3154)
	at org.eclipse.jface.operation.ModalContext.runInCurrentThread(ModalContext.java:437)
	at org.eclipse.jface.operation.ModalContext.run(ModalContext.java:353)
	at org.eclipse.ui.internal.WorkbenchWindow$14.run(WorkbenchWindow.java:2195)
[...]

PR #1344, opened 2020-02-17
[eclipse/xtext-eclipse#1208] try to handle storage with null path

jrosczak:
I did the least inversive change. Worked in my environment, but not sure if we do need the IPathEditorInput check, too.

see eclipse/xtext#2384

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
@@ -10,10 +10,13 @@ package org.eclipse.xtext.ui.editor.model;
 
 import java.util.Collections;
 
+import org.eclipse.core.filesystem.EFS;
+import org.eclipse.core.filesystem.IFileStore;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IStorage;
 import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.emf.common.util.URI;
@@ -22,6 +25,7 @@ import org.eclipse.emf.ecore.resource.ContentHandler;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.ResourceSet;
 import org.eclipse.ui.IEditorInput;
+import org.eclipse.ui.IPathEditorInput;
 import org.eclipse.ui.IStorageEditorInput;
 import org.eclipse.ui.IURIEditorInput;
 import org.eclipse.xtext.resource.IExternalContentSupport;
@@ -75,11 +79,22 @@ public class ResourceForIEditorInputFactory implements IResourceForEditorInputFa
 		try {
 			if (editorInput instanceof IStorageEditorInput) {
 				IStorage storage = ((IStorageEditorInput) editorInput).getStorage();
+				if (storage.getFullPath() == null) {					
+					IFile file = Adapters.adapt(editorInput, IFile.class);
+					if (file != null) 
+						storage = file;
+				}
 				Resource result = createResource(storage);
 				if (result != null)
 					return result;
 			} else if (editorInput instanceof IURIEditorInput) {
 				Resource result = createResource(((IURIEditorInput) editorInput).getURI());
+				if (result != null)
+					return result;
+			} else if (editorInput instanceof IPathEditorInput) {
+				IPath path = ((IPathEditorInput) editorInput).getPath();
+				IFileStore fileStorage = EFS.getLocalFileSystem().getStore(path);
+				Resource result = createResource(fileStorage.toURI());
 				if (result != null)
 					return result;
 			}
@@ -107,8 +122,15 @@ public class ResourceForIEditorInputFactory implements IResourceForEditorInputFa
 	}
 
 	protected Resource createResourceFor(IStorage storage) throws CoreException {
+		String path = null;
+		if (storage.getFullPath() != null)
+			path = storage.getFullPath().toString();
+		if (path == null)
+			path = storage.getName();
+		if (path == null)
+			throw new IllegalArgumentException("Couldn't create EMF Resource for storage " + storage);
 		ResourceSet resourceSet = getResourceSet(storage);
-		URI uri = URI.createPlatformResourceURI(storage.getFullPath().toString(), true);
+		URI uri = URI.createPlatformResourceURI(path, true);
 		configureResourceSet(resourceSet, uri);
 		URI uriForResource = uri; 
 		if (!uri.isPlatform()) {
@@ -119,7 +141,7 @@ public class ResourceForIEditorInputFactory implements IResourceForEditorInputFa
 		resource.setValidationDisabled(isValidationDisabled(uri, storage));
 		return resource;
 	}
-
+	
 	/**
 	 * Checks whether syntax validation should be disabled for {@code uri}.
 	 * This is called when creating a resource for {@code uri}.


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
```diff
@@ -75,6 +75,11 @@ public Resource createResource(IEditorInput editorInput) {
 		try {
 			if (editorInput instanceof IStorageEditorInput) {
 				IStorage storage = ((IStorageEditorInput) editorInput).getStorage();
+				if (storage.getFullPath() == null) {
+					IFile file = editorInput.getAdapter(IFile.class);
```
szarnekow: Should probably be `Adapters.getAdapter(..)` to honor registered adapter factories.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
```diff
@@ -107,6 +112,8 @@ protected Resource createResource(IStorage storage) throws CoreException {
 	}
 
 	protected Resource createResourceFor(IStorage storage) throws CoreException {
+		if (storage.getFullPath() == null)
```
szarnekow: I wouldn't expect a createXYZ method to return null.

szarnekow:
@jrosczak could you please add the support for IPathEditorInput, too?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
```diff
@@ -107,6 +112,8 @@ protected Resource createResource(IStorage storage) throws CoreException {
 	}
 
 	protected Resource createResourceFor(IStorage storage) throws CoreException {
+		if (storage.getFullPath() == null) 
+			return null;
 		ResourceSet resourceSet = getResourceSet(storage);
 		URI uri = URI.createPlatformResourceURI(storage.getFullPath().toString(), true);
```
szarnekow: Please remove the null guard above and to a second attempt in line 118 `URI.createPlatformResourceURI(storage.getFullPath().toString(), true);` with the name of the storage if the path is null. If both are null, I'd say we should throw an exception similar to the one in `createResource(IEditorInput)`.
Out of paranoia we could even try with storage.toString() and attempt to figure if this looks like a file name.

jrosczak:
@szarnekow Sure I can add support for IPathEditorInput, but then I would have to change the signature of the protected method. Would this be okay? I do not want to break something.

szarnekow:
Maybe a new overload with storage and path as arguments will do the trick?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
jrosczak: Yes, I changed it.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
jrosczak: I changed it.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
jrosczak: I check for fullPath and name, but I am not sure if using storage.getName is really a good idea. In this case I loose the project information which for instance is required by xbase dsls to compute the Java classpath.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
```diff
@@ -107,8 +122,15 @@ protected Resource createResource(IStorage storage) throws CoreException {
...
+		String path = null;
+		if (storage.getFullPath() != null)
+			path = storage.getFullPath().toString();
+		if (path == null)
+			path = storage.getName();
+		if (path == null)
+			throw new IllegalArgumentException("Couldn't create EMF Resource for storage " + storage);
 		ResourceSet resourceSet = getResourceSet(storage);
-		URI uri = URI.createPlatformResourceURI(storage.getFullPath().toString(), true);
+		URI uri = URI.createPlatformResourceURI(path, true);
```
szarnekow: If we take the name of the storage, we should probably use a different resource scheme. if there is no member in the workspace that matches the name.
@cdietrich WDYT?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
szarnekow: Thank you.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
cdietrich: i did not find the time to investigate/reproduce the issue yet

jrosczak: @szarnekow I think there will not be a member in the workspace that matches the name, because without the full path we do not know to which project the storage belongs.
So I would opt for not using the name of the storage.

jrosczak:
Is there any change that is gets included in the next Xtext release?

miklossy:
Currently it is scheduled to Xtext 2.22.0 that will be releasen on 02.06.2020.

cdietrich:
the problem is:
i dont have any time to think about the usecase or problem you try to fix or to setup a reproducible example => i cannot judge the risk

@szarnekow what do you think

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
szarnekow: Agreed.

szarnekow:
In fact I'd like to see this land in 2.21

Two more things, though:
1) The TODO in org.eclipse.xtext.ui.editor.model.ResourceForIEditorInputFactory.createResource(ResourceSet, URI) should probably be resolved. As soon as it *is* resolved, we can still use the name of the storage (if available) since we simply need a non-null URI. In fact, it could be arbitrary.

2) In org.eclipse.xtext.ui.editor.model.ResourceForIEditorInputFactory.getResourceSet(IStorage) we try to find a project for a storage. That logic could benefit from the changes in this PR, too.

jrosczak:
@szarnekow I am not sure what you want to achieve with the action items.

1. The TODO in createResource: It can be replaced with ResourceFactory, but why have the ResourceSet as parameter than? Looks strange to me if it is not used.
And in other places where the ResourceFactory is already used, the result is directly cast to XtextResource, would that not need the same safety check as in createResource?

2. What do you want to add here? Due to the change in createResource(IEditorInput) the IStorage is probably already an IFile. If the path of the storage is empty you are out of luck I guess. The name of the storage will not help at all, because it does not contain a project part.

cdietrich:
@szarnekow move to 2.22?

cdietrich:
@szarnekow @jrosczak what is the state of this PR?

szarnekow:
I've applied a similar though not identical fix in a project and it looked good so far. I'll come back to this with concrete suggestions.

szarnekow:
@jrosczak I'd like to hijack your PR and apply a few changes. Are you ok with that?

jrosczak:
@szarnekow Sure, it is all yours.

cdietrich:
move to 2.24

cdietrich:
move to 2.25

cdietrich:
@szarnekow any update here?

szarnekow:
> @szarnekow any update here?

Unfortunately not.

cdietrich:
move to 2.26
any update here @szarnekow ?

szarnekow:
> any update here @szarnekow ?

Life is sad.

cdietrich:

PR #1352, opened 2020-02-24
[eclipse/xtext-umbrella#202] workaround for egit oomph problem

cdietrich:
[eclipse/xtext-umbrella#202] workaround for egit oomph problem


szarnekow:
I'm actually inclined to leave this 'workaround' as the default. Wdyt?

cdietrich:
Would be ok with that

PR #1375, opened 2020-03-22
Migrated the Xtend code from org.eclipse.xtext.builder project to Java 

nbhusare:
- [x] Migrated Xtend code to Java
- [x] Ran the Maven/Gradle build locally
- [x] Removed "org.eclipse.xtend.lib" dependency from the manifest file
- [x] Removed "xtend-gen" folder and updated the .classpath file
- [x] Replaced the references to *.xbase.* with appropriate Java equivalent wherever possible.


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/ParallelFileSystemAccess.java
```diff
@@ -0,0 +1,164 @@
+/**
+ * Copyright (c) 2017, 2020 itemis AG (http://www.itemis.eu) and others.
```
nbhusare: @cdietrich @tivervac Kindly confirm the date. The Copyright is missing in the file history. I checked other files that were introduced in the "2.7" release and put the date.  Please suggest if this needs to be changed.

cdietrich: should be 2014

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/JdtQueuedBuildData.java
```diff
@@ -0,0 +1,138 @@
...
+			return true;
+		}
+		return false;
+	}
+
+	@Override
+	public boolean needsRebuild(IProject project, Collection<IResourceDescription.Delta> deltas) {
+		JavaBuilderState oldState = javaBuildState.get(project.getName());
+		JavaBuilderState newState = JavaBuilderState.getLastBuiltState(project);
+		return doNeedRebuild(newState, (delta) -> {
```
nbhusare: @cdietrich @tivervac @szarnekow The earlier code sets a `null` value in the second param if the if-condition on line 82 is  not satisfied. I believe that was not needed. Please review and confirm.

org.eclipse.xtext.builder/.classpath
```diff
@@ -8,7 +8,6 @@
 			<attribute name="ignore_optional_problems" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry kind="src" path="xtend-gen"/>
```
szarnekow: Yeehaa! :)

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/ParallelFileSystemAccess.java
nbhusare: Updated the year to 2014.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/JdtQueuedBuildData.java
szarnekow: The old code passes null as the callable into doNeedRebuild. Since this code here is very brittle and relies a lot on the internals of the Java builder, I'd prefer to keep it exactly as it was.

nbhusare: @szarnekow I have made the changes. Please check.

cdietrich:
thx @nbhusare

PR #1378, opened 2020-03-23
restored old api of SingleHoverShowingHyperlinkPresenter

cdietrich:
restored old api of SingleHoverShowingHyperlinkPresenter

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hyperlinking/SingleHoverShowingHyperlinkPresenter.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hyperlinking/SingleHoverShowingHyperlinkPresenter.java
@@ -13,7 +13,6 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
 
 import org.apache.log4j.Logger;
 import org.eclipse.jface.text.AbstractInformationControlManager;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/hyperlinking/SingleHoverShowingHyperlinkPresenter.java
```diff
@@ -55,37 +56,38 @@ public SingleHoverShowingHyperlinkPresenter(MultipleHyperlinkPresenter delegate)
...
 	/**
 	 * @since 2.9
 	 */
-	protected List<IHyperlink> makeNullsafe(List<IHyperlink> hyperlinks) {
-		if (hyperlinks.stream().anyMatch(hyperLink -> (hyperLink == null || hyperLink.getHyperlinkRegion() == null))) {
+	protected IHyperlink[] makeNullsafe(IHyperlink[] hyperlinks) {
+		if (Arrays.stream(hyperlinks).anyMatch(hyperLink -> (hyperLink == null || hyperLink.getHyperlinkRegion() == null))) {
 			List<IHyperlink> nullSafeList = new ArrayList<>();
-			hyperlinks.forEach(hyperLink -> {
+			Arrays.stream(hyperlinks).forEach(hyperLink -> {
```
szarnekow: I'm still not a big fan of streams where the forEach modified some captured instance, e.g. the nullSafeList.
A old-school for loop may have been the better choice.
Since it looks correct though, I'll approve

miklossy:
I also encountered the problem that the hyperlinking (e.g. in Xtend files) was not working. After updating to the Xtext-latest the problem is gone. Thanks for the fix @cdietrich !

Issue #1383, opened 2020-03-30
XtextBuilder should not call pollQueuedBuildData on clean


cdietrich:
XtextBuilder calls pollQueuedBuildData which may lead to a needRebuild call on the builder.
(for unrelated deltas)
this then leads to a immediate full build of the project.
(which undermines the clean)

we should remove the pollQueuedBuildData again. it will be called on following builds on the incremental build anyway.

Also org.eclipse.xtext.builder.impl.ClosedProjectsQueue.exhaust()
is protected and thus impossible to easly be called from XtextBuilder subclasses

cdietrich:
Fixed in 2.22

PR #1384, opened 2020-03-30
No poll at clean + api for ClosedProjectsQueue

cdietrich:
[#1383] No poll at clean + api for ClosedProjectsQueue


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/ClosedProjectsQueue.java
```diff
@@ -130,7 +130,7 @@ public int getAttempt() {
 	 * @param toBeBuilt
 	 *            their contents.
 	 */
-	protected void enqueue(Set<String> projectNames, ToBeBuilt toBeBuilt) {
+	public void enqueue(Set<String> projectNames, ToBeBuilt toBeBuilt) {
```
szarnekow: Do we want to add `@since` tags now?

PR #1387, opened 2020-03-31
Migrating a subset of Xtend code from org.eclipse.xtext.xbase.ui project to Java

nbhusare:
- [x] Migrated subset of Xtend code to Java
- [x] Ran the Maven/Gradle build locally


org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,162 @@
...
+								classpathEntries.add(classpathEntry);
+							}
+						}
+						
+						for (IClasspathEntry sourceFolder : classpathEntries) {
+							if (sourceFolder.getOutputLocation() != null && sourceFolder.getOutputLocation().isPrefixOf(resource.getFullPath())) {
+								IPath relative = resource.getFullPath().removeFirstSegments(sourceFolder.getOutputLocation().segmentCount());
+								Optional<IPackageFragmentRoot> source = FluentIterable.from(project.findPackageFragmentRoots(sourceFolder))
+										.first();
+								if (source.isPresent()) {
```
nbhusare: Added a safety-check which was not present in the original code. The code `project.findPackageFragmentRoots(sourceFolder)` may return an empty array.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,162 @@
...
+									}
+								}
+							}
+						}
+
+					}
+				}
+			}
+		} catch (JavaModelException e) {
+			LOG.error("Error finding the original source for the editor input", e);
```
nbhusare: Logging an error message. The chose to do it after looking at other places in the source code where `JavaModelException` is caught. Please suggest if you would like me to replace with a `RuntimeException`.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,162 @@
...
+			if (original != input) {
+				return original;
+			}
+
+			IEclipseTrace trace = traceInformation.getTraceToSource(resource);
+			if (trace == null) {
+				return input;
+			}
+
+			Optional<? extends ILocationInEclipseResource> sourceInformation = Iterables.tryFind(trace.getAllAssociatedLocations(),
```
nbhusare: Simplified the code. The original code made use of a `while-loop` to find the first element for which the “languageInfo.equals(candidate.getLanguage())” is true. I do the same using the new code

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,162 @@
...
+		return input;
+	}
+
+	/**
+	 * @return the {@code IPackageFragmentRoot} that contains the provided {@code element}
+	 * @throws NullPointerException
+	 *             if the provided {@code element} is null
+	 */
+	public IPackageFragmentRoot getPackageFragmentRoot(IJavaElement element) {
+		Preconditions.checkNotNull(element, "Couldn't find source package fragment root.");
```
nbhusare: Added a safety-check here.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
```diff
@@ -0,0 +1,197 @@
...
+
+import com.google.common.base.Optional;
+import com.google.common.collect.FluentIterable;
+import com.google.inject.Inject;
+
+/**
+ * @author Anton Kosyakov - Initial contribution and API
+ * @since 2.4
+ */
+public class XbaseCopyQualifiedNameService extends DefaultCopyQualifiedNameService {
```
nbhusare: `org.eclipse.xtext.ui.editor.copyqualifiedname.DefaultCopyQualifiedNameService` is deprecated. We'll have to revisit this code and replace the deprecated API with the correct ones. Corresponding change needs to be made in the `XbaseUiModule#bindCopyQualifiedNameService()` method.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
```diff
@@ -0,0 +1,197 @@
...
+		if (actualType != null) {
+			resolvedFeatures = overrideHelper.getResolvedFeatures(actualType);
+		}
+
+		List<IResolvedConstructor> declaredConstructors = null;
+		if (resolvedFeatures != null) {
+			declaredConstructors = resolvedFeatures.getDeclaredConstructors();
+		}
+
+		Optional<IResolvedConstructor> resolvedConstructor = FluentIterable.from(declaredConstructors)
```
nbhusare: `FluentIterable` - A bit of inspiration from @szarnekow code.

nbhusare:
@cdietrich @tivervac @szarnekow Kindly review.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseContentProposalPriorities.java
```diff
@@ -0,0 +1,52 @@
...
+public class XbaseContentProposalPriorities extends ContentProposalPriorities {
+
+	@Override
+	public void adjustCrossReferencePriority(ICompletionProposal proposal, String prefix) {
+		if (proposal instanceof ConfigurableCompletionProposal) {
+			ConfigurableCompletionProposal configurableProposal = (ConfigurableCompletionProposal) proposal;
+			Object desc = configurableProposal.getAdditionalData(XbaseProposalProvider.DESCRIPTION_KEY);
+			if (desc instanceof SimpleIdentifiableElementDescription) {
+				if (!("this".equals(configurableProposal.getReplacementString())
+						&& "super".equals(configurableProposal.getReplacementString()))) {
```
tivervac: Either add the `!` before both equals, or do `&&` -> `||`

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
tivervac: Going to leave this one for @szarnekow

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
tivervac: Seems fine to me especially with the extra log

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,162 @@
...
+		} catch (JavaModelException e) {
+			LOG.error("Error finding the original source for the editor input", e);
+		}
+
+		return input;
+	}
+
+	public IEditorInput findOriginalSource(IEditorInput input) {
+		IFile resource = ResourceUtil.getFile(input);
+		if (resource != null) {
```
tivervac: Might be better to just write `if (resource == null) return input;`

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
tivervac: LGTM

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
```diff
@@ -0,0 +1,197 @@
...
+		} else if (constructor != null && constructorCall != null) {
+			return _getQualifiedName(constructor, constructorCall);
+		} else if (constructor != null && constructorCall == null) {
+			return _getQualifiedName(constructor, (Void) null);
+		} else if (constructor == null && constructorCall != null) {
+			return _getQualifiedName((Void) null, constructorCall);
+		} else if (constructor == null && constructorCall == null) {
+			return _getQualifiedName((Void) null, (Void) null);
+		} else {
+			throw new IllegalArgumentException("Unhandled parameter types: " + Arrays.asList(constructor, constructorCall).toString());
```
tivervac: You can remove this, it's impossible to end up here, I don't if you leave it though

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
```diff
@@ -0,0 +1,197 @@
...
+			resolvedFeatures = overrideHelper.getResolvedFeatures(actualType);
+		}
+
+		List<IResolvedConstructor> declaredConstructors = null;
+		if (resolvedFeatures != null) {
+			declaredConstructors = resolvedFeatures.getDeclaredConstructors();
+		}
+
+		Optional<IResolvedConstructor> resolvedConstructor = FluentIterable.from(declaredConstructors)
+				.filter(constr -> constr.getDeclaration().equals(consturctor)) //
```
tivervac: constructor*

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
tivervac: Does this work if declaredConstructors is null?

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
nbhusare: Let me change it to `declaredConstructor` to be consistent with the list `declaredConstructors`.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
nbhusare: `declaredConstructors` can be empty, but not null. PS - https://bit.ly/3dIG53z

Generally speaking, `FluentIterable.from` will throw a NPE if the passed param. is null.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseContentProposalPriorities.java
```diff
@@ -0,0 +1,52 @@
...
+			ConfigurableCompletionProposal configurableProposal = (ConfigurableCompletionProposal) proposal;
+			Object desc = configurableProposal.getAdditionalData(XbaseProposalProvider.DESCRIPTION_KEY);
+			if (desc instanceof SimpleIdentifiableElementDescription) {
+				if (!"this".equals(configurableProposal.getReplacementString())
+						&& !"super".equals(configurableProposal.getReplacementString())) {
+					adjustPriority(proposal, prefix, 570);
+					return;
+				}
+			} else if (desc instanceof StaticFeatureDescriptionWithTypeLiteralReceiver) {
+				adjustPriority(proposal, prefix, 560);
```
szarnekow: This will fall through to the super.adjust.. call. Not sure if that was the intention in the xtend code.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
szarnekow: Please only log on error level if the exception is not `e.isDoesNotExist()`

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	/**
+	 * @return the {@code IPackageFragmentRoot} that contains the provided {@code element}
+	 * @throws NullPointerException
+	 *             if the provided {@code element} is null
+	 */
+	public IPackageFragmentRoot getPackageFragmentRoot(IJavaElement element) {
+		Preconditions.checkNotNull(element, "Couldn't find source package fragment root.");
+		return element instanceof IPackageFragmentRoot ? (IPackageFragmentRoot) element : getPackageFragmentRoot(element.getParent());
```
szarnekow: We probably want to use `org.eclipse.jdt.core.IJavaElement.getAncestor(int)` here

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,165 @@
...
+						for (IClasspathEntry classpathEntry : project.getRawClasspath()) {
+							if (classpathEntry.getEntryKind() == IClasspathEntry.CPE_SOURCE) {
+								classpathEntries.add(classpathEntry);
+							}
+						}
+						
+						for (IClasspathEntry sourceFolder : classpathEntries) {
+							if (sourceFolder.getOutputLocation() != null && sourceFolder.getOutputLocation().isPrefixOf(resource.getFullPath())) {
+								IPath relative = resource.getFullPath().removeFirstSegments(sourceFolder.getOutputLocation().segmentCount());
+								Optional<IPackageFragmentRoot> source = FluentIterable.from(project.findPackageFragmentRoots(sourceFolder))
```
szarnekow: Iff we change this code, we could loop over all matching packageFragmentRoots until you find one that fulfills `newFile.exists()` below.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,165 @@
...
+								IPath fullPath = source.getCorrespondingResource().getProjectRelativePath().append(relative);
+								IFile newFile = resource.getProject().getFile(fullPath);
+								if (newFile.exists()) {
+									return new FileEditorInput(newFile);
+								}
+							}
+						}
+
+						// Check if it's sitting in one of the output folders set on the source folders
+						List<IClasspathEntry> classpathEntries = new ArrayList<>();
```
szarnekow: Populating a list to loop over the list on step later appears to be unnecessary. Maybe extract a few smaller methods to improve the readability and avoid the intermediate list here?

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,165 @@
...
+			if (resource != null) {
+				// if the given resource is already pointing to a language file
+				if (fileExtensionProvider.isValid(resource.getFullPath().getFileExtension())) {
+					IJavaProject project = JavaCore.create(resource.getProject());
+					if (project.exists()) {
+						// Check if it's sitting in the output dir, look for original one
+						if (project.getOutputLocation().isPrefixOf(resource.getFullPath())) {
+							IPath relative = resource.getFullPath().removeFirstSegments(project.getOutputLocation().segmentCount());
+
+							List<IPackageFragmentRoot> packageFragmentRoots = new ArrayList<>();
```
szarnekow: Same comment here as a few lines below: We create unnecessary lists and end up with a very long method. Maybe it's better to refactor this by extracting a few smaller methods.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,165 @@
...
+		if (original != input) {
+			return original;
+		}
+
+		IEclipseTrace trace = traceInformation.getTraceToSource(resource);
+		if (trace == null) {
+			return input;
+		}
+
+		Optional<? extends ILocationInEclipseResource> sourceInformation = Iterables.tryFind(trace.getAllAssociatedLocations(),
```
szarnekow: I'd prefer a more straight forward 
```
for(PlatformResource candidate: trace.getAllAssociatedLocations()) {
  if (languageInfo.equals(candidate.getLanguage()) {
    IStorage storage = candidate.getPlatformResource();
    if (storage != null) {
      return ..
    }
  }
}
return input;
```

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
szarnekow: Please take a look at the conversion that was applied to the org.eclipse.xtext.naming.DefaultCopyQualifiedNameService - there we use a PolymorphicDispatcher. 
I'm afraid we have to do the same here in order to remain backwards compatible.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
szarnekow: declaredConstructors will still be empty if resolvedFeatures is null.

These null checks look a little too xtendy to me. Many of the intermediate local vars will never be null if the actual receiver was present.
Please reverse the logic and exit early but otherwise keep a normal flow.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
```diff
@@ -0,0 +1,195 @@
...
+
+		List<IResolvedConstructor> declaredConstructors = null;
+		if (resolvedFeatures != null) {
+			declaredConstructors = resolvedFeatures.getDeclaredConstructors();
+		}
+
+		Optional<IResolvedConstructor> resolvedConstrutor = FluentIterable.from(declaredConstructors)
+				.filter(constr -> constr.getDeclaration().equals(constructor)) //
+				.first();
+		return resolvedConstrutor.isPresent() ? resolvedConstrutor.get() : null;
```
szarnekow: Typo in variable name

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
szarnekow: Also my preference would always be something like
```
for(IResolvedConstructor candidate: declaredConstructors) {
  if (constructor.equals(candidate.getDeclaration()) {
    return candidate;
 }
}
return null

```

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
```diff
@@ -0,0 +1,195 @@
...
+			return _resolveExecutable((JvmConstructor) consturctor, (XConstructorCall) expression, resolvedTypes);
+		} else if (consturctor instanceof JvmOperation && expression instanceof XAbstractFeatureCall) {
+			return _resolveExecutable((JvmOperation) consturctor, (XAbstractFeatureCall) expression, resolvedTypes);
+		} else {
+			throw new IllegalArgumentException(
+					"Unhandled parameter types: " + Arrays.asList(consturctor, expression, resolvedTypes).toString());
+		}
+	}
+
+	protected IResolvedExecutable _resolveExecutable(JvmConstructor consturctor, XAbstractFeatureCall featureCall,
```
szarnekow: Typo in parameter name

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
```diff
@@ -0,0 +1,195 @@
...
+		if (actualType != null) {
+			resolvedFeatures = overrideHelper.getResolvedFeatures(actualType);
+		}
+
+		List<IResolvedOperation> allOperations = null;
+		if (resolvedFeatures != null) {
+			allOperations = resolvedFeatures.getAllOperations();
+		}
+
+		Optional<IResolvedOperation> resolvedOperation = FluentIterable.from(allOperations)
```
szarnekow: Not sure how often we do similar things in the code base, but I can imagine this logic to be moved to `ResolvedFeatures`, e.g. `findResolvedOperation(JvmOperation declaration)`

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/util/InsertionOffsets.java
```diff
@@ -0,0 +1,41 @@
...
+		return NodeModelUtils.findActualNodeFor(element).getOffset();
+	}
+
+	public int after(EObject element) {
+		return NodeModelUtils.findActualNodeFor(element).getEndOffset();
+	}
+
+	public int inEmpty(EObject element) {
+		ICompositeNode compositeNode = NodeModelUtils.findActualNodeFor(element);
+		Optional<ILeafNode> openingBraceNode = Iterables.tryFind(compositeNode.getLeafNodes(), (node) -> "{".equals(node.getText()));
```
szarnekow: Not a big fan of these usages of optionals. Checking for `isPresent` is usually not good style. You probably want to do something like `Iterables.tryFind(..).transform(node->getOffset() + 1).or(compositeNode::getEndOffset)`

szarnekow: Or preferably a plain for loop.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
tivervac: It is going to be null if `resolvedFeatures` is null

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
nbhusare: @szarnekow What should I do if the condition `e.isDoesNotExist()` is not met?

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
nbhusare: You are right, I missed that.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseContentProposalPriorities.java
nbhusare: PS - https://bit.ly/3bJ43Kg

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
nbhusare: Very good suggestion. Tx.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
nbhusare: The above code goes through all-the-candidates until it finds one for which `storage != null`. If you look at the original code (https://bit.ly/2wXAeXN), it gets the **first** candidate that matches the condition `languageInfo.equals(candidate.getLanguage())`, and uses it for further processing. 

@szarnekow Is it possible for more than one `candidate` to match the condition `languageInfo.equals(candidate.getLanguage())` ?

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
nbhusare: The methods `ResolvedFeatures#getAllOperations()` and `ResolvedFeatures#getDeclaredConstructors()` are used in very few places, not for the same purpose as we have here. Nevertheless, I like you idea of adding methods `#findResolvedOperation(JvmOperation declaration)` (and  `#findDeclaredConstructor(JvmConstructor constructor)`) to the `ResolvedFeatures` class.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
nbhusare: I'll file a separate jira for this.

nbhusare:
@szarnekow @tivervac Tx for the review. I have addressed all your review comments. Kindly have a look, and resolve the conversation if you are ok with the changes.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,150 @@
...
+										}
+									}
+								}
+							}
+						}
+					}
+				}
+			}
+		} catch (JavaModelException e) {
+			if (e.isDoesNotExist()) {
```
szarnekow, line 102: It's the opposite of what I tried to suggest: Only log, ig it is not `e.isDoesNotExist()`

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	/**
+	 * @return the {@code IPackageFragmentRoot} that contains the provided {@code element}
+	 * @throws NullPointerException
+	 *             if the provided {@code element} is null
+	 */
+	public IPackageFragmentRoot getPackageFragmentRoot(IJavaElement element) {
+		Preconditions.checkNotNull(element, "Couldn't find source package fragment root.");
+		return element instanceof IPackageFragmentRoot //
```
szarnekow, line 145: The semantics of getAncestor do already imply the check of the element itself. Can be implemented as `return (IPackageFragmentRoot) element.getAncestor()`

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
```diff
@@ -0,0 +1,172 @@
...
+					"Unhandled parameter types: " + Arrays.asList(consturctor, expression, resolvedTypes).toString());
+		}
+	}
+
+	protected IResolvedExecutable _resolveExecutable(JvmConstructor constructor, XAbstractFeatureCall featureCall,
+			IResolvedTypes resolvedTypes) {
+		XExpression actualReceiver = featureCall.getActualReceiver();
+		if (actualReceiver != null) {
+			LightweightTypeReference actualType = resolvedTypes.getActualType(actualReceiver);
+			ResolvedFeatures resolvedFeatures = overrideHelper.getResolvedFeatures(actualType);
```
szarnekow, line 133: Code is duplicated in the method below.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/copyqualifiedname/XbaseCopyQualifiedNameService.java
szarnekow: https://github.com/eclipse/xtext-extras/issues/581

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/contentassist/XbaseContentProposalPriorities.java
szarnekow: I think this was a bug in the old impl.

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	/**
+	 * @return the original source for an editor input that points to an Xtext resource copied to the output folder, the given input
+	 *         otherwise
+	 */
+	public IEditorInput findOriginalSourceForOuputFolderCopy(IEditorInput input) {
+		try {
+			IFile resource = ResourceUtil.getFile(input);
+			if (resource != null) {
```
tivervac, line 60: I would early return here as well, also this doesn't need to be in try catch, try to keep it as small as possible

org.eclipse.xtext.xbase.ui/src/org/eclipse/xtext/xbase/ui/editor/XbaseEditorInputRedirector.java
```diff
@@ -0,0 +1,150 @@
...
+				// if the given resource is already pointing to a language file
```
tivervac, line 61: s/if/If/

nbhusare:
@tivervac @szarnekow Addressed review comments. Please check.

cdietrich:
many thanks @nbhusare

Issue #1391, opened 2020-04-05
Xtext Editor: 'Create rule' quickfix adds Linux line separator even on Windows.


miklossy:
When applying the `Create Rule ` quickfix on Windows, the newly created rule contains Linux line separators on Windows. Reproducible with the Xtext MyDsl example:
![screencast](https://user-images.githubusercontent.com/20393472/78471993-c4fb5880-7735-11ea-943c-8085cdfe1daa.gif)

miklossy:
Fixed in 2.22

szarnekow:
It’s unfortunately not correct to use the system line delimiter.

miklossy:
Thanks for the feedback @szarnekow ! I created the corresponding PR to use the `document.getLineDelimiter` instead of the `System.lineseparator`.

miklossy:
Fixed in 2.22

PR #1393, opened 2020-04-05
[#1391] Use document.getLineDelimiter instead of System.lineSeparator.

miklossy:


miklossy:
Thanks for the infos @szarnekow!

Issue #1399, opened 2020-04-12
Xtext Outline View - Return type styling


miklossy:
Currently, the styling of the rule's return types is not very aesthetic (at least on Windows)
![screenshot](https://user-images.githubusercontent.com/20393472/79064355-79522d00-7ca8-11ea-80bf-2b02c3b1f524.png)

@cdietrich Could you upload a screenshot how it currently looks like on Mac?

I would suggest to use the same styling for the rule's return types as JDT does for the method's return type
![screenshot2](https://user-images.githubusercontent.com/20393472/79064376-b3233380-7ca8-11ea-9869-6c3f9f8d2f33.png)

cdietrich:
linux

![Bildschirmfoto_2020-04-12_10-44-19](https://user-images.githubusercontent.com/238736/79064610-a43d8080-7caa-11ea-9d9c-6b0623f0cf11.png)

cdietrich:
mac

![Bildschirmfoto 2020-04-12 um 10 46 06](https://user-images.githubusercontent.com/238736/79064641-dea71d80-7caa-11ea-8b3a-f0f8f006362c.png)

miklossy:
Fixed in 2.22

PR #1400, opened 2020-04-12
[#1399] Xtext Outline View - Return type styling improvements.

miklossy:
- Modify the XtextOutlineTreeProvider to use the same styling for the
rule's return types as JDT does for the method's return type.

### DIFF
--- a/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/outline/XtextOutlineTreeProvider.java
+++ b/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/outline/XtextOutlineTreeProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2010, 2020 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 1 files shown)


szarnekow (changes requested):
For the sake of consistency we should remove all usages of createXtextStyleAdapter

miklossy:
Since the createXtextStyleAdapter call is present on different places in the [XtextOutlineTreeProvider](https://github.com/eclipse/xtext-eclipse/blob/ab20b6bc984f974700bdfe5bcf2bf33521f2467a/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/outline/XtextOutlineTreeProvider.java#L64-L70) I would leave it for now as it is and refactor it later on.

szarnekow:
I don't agree. If I change my coloring settings to italic for data type rules, the outline will look weird for that case, too.

miklossy:
@szarnekow You are right! I updated the PR as requested.

miklossy:
There is a test failure, the `org.eclipse.xtext.common.types.ui.notification.TypeResourceUnloaderTest.testCloseEditorAndDiscardWorkingCopy` test case is failing on Jenkins, but I think it is not related to this change. @cdietrich Is this test flaky?

cdietrich:
yes please create and issue for that (TypeResourceUnloaderTest)

miklossy:
Thanks for restarting the CI Job @cdietrich , it is green again!

miklossy:
@szarnekow Please review again!

szarnekow (approved):
Nice.
While testing this I found the icon in the outline for return types to be weird. Can you please update that, too?

miklossy:
> While testing this I found the icon in the outline for return types to be weird. Can you please update that, too?

@szarnekow Hmm... I did not get your point. Can you please upload a screenshot that I can better imagine?

szarnekow:
Filed https://github.com/eclipse/xtext/issues/2383

PR #1424, opened 2020-05-03
[https://github.com/eclipse/xtext/issues/1631] Add deprecation.

miklossy:
- Deprecate org.eclipse.xtext.junit4.util.ResourceLoadHelper in favor of
org.eclipse.xtext.ui.testing.util.ResourceLoadHelper

### DIFF
--- a/org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/util/ResourceLoadHelper.java
+++ b/org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/util/ResourceLoadHelper.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2011 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2011, 2020 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -16,7 +16,9 @@ import org.eclipse.xtext.resource.XtextResource;
  * Abstraction for other test utilities that want to load a resource for a
  * textual model.
  * @author Sebastian Zarnekow - Initial contribution and API
+ * @deprecated Use org.eclipse.xtext.ui.testing.util.ResourceLoadHelper instead
  */
+@Deprecated
 public interface ResourceLoadHelper {
 
 	XtextResource getResourceFor(InputStream stream);


org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/util/ResourceLoadHelper.java
```diff
@@ -16,7 +16,9 @@
  * Abstraction for other test utilities that want to load a resource for a
  * textual model.
  * @author Sebastian Zarnekow - Initial contribution and API
+ * @deprecated Use org.eclipse.xtext.ui.testing.util.ResourceLoadHelper instead
```
cdietrich, line 19: am not sure if we want this.
there are gazillion of our tests that use this one.
@szarnekow  what do you think.

miklossy: I found only a couple of places where this interface is used:
image

cdietrich: these classes have gazillions of subclasses

org.eclipse.xtext.junit4/src/org/eclipse/xtext/junit4/util/ResourceLoadHelper.java
```diff
@@ -16,7 +16,9 @@
...
  */
+@Deprecated
```
szarnekow, line 21: I don't think deprecating this class is providing added value.

miklossy: Since the entire `org.eclipse.xtext.junit4` project is deprecated (has been superseded by the `org.eclipse.xtext.testing` project), I think it is worth marking all elements (classes/interfaces) in that project as deprecated (where there is a corresponding alternative to use) to see where in the Xtext code-base deprecated elements are used and where there is a need for restructuring to eliminate the usage of deprecated elements from the Xtext code-base.

szarnekow: I can live with this, if (and only if) we eliminate most of the usages of the now-deprectated code from the Xtext codebase. In this case, this is trivial to show that the "new" solution is working at least as good as the old one, but generally speaking this is to ensure that the new code fulfils the same requirements as the old code.
The drawback of that is, that the now-deprecated code is then subject to be forgotten since it would no longer be tested. Interesting twist, I suppose.

If you want to work on the deprecation of the old junit4 bundles, I recommend to do it completely or not at all. E.g. add a package-info to all the packages in the deprecated bundle, and mark the entire package as deprecated with a hint that points to its new equivalent. There is no reason to make ResourceLoadHelper deprecated but keep AbstractXtextTests as non-deprecated base type. Before that, we'd need to figure a migration path for the tests, though. If AbstractXtextTests stays as is, the entire deprecation effort is as half-hearted as it is right now. And if it is deprecated, we'll have a lot more warnings in the code base.

miklossy: I am fine with that, I created an [issue](https://github.com/eclipse/xtext-eclipse/issues/1430) for that work. I am closing this PR without merging it.

PR #1427, opened 2020-05-04
Fix potential null pointer access in the AdvancedNewProjectPage class.

miklossy:


miklossy:
The e parameter can never be null, since the caller creates a new `SelectionEvent` object:
```java
package org.eclipse.swt.widgets;

public class TypedListener implements Listener {
	...	
		case SWT.Selection: {
			...
			SelectionEvent event = new SelectionEvent (e);
			((SelectionListener) eventListener).widgetSelected (event);
			...
		}
	...
}
```

org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/AdvancedNewProjectPage.java
```diff
@@ -290,6 +290,8 @@ public void checkWidgets(SelectionEvent e) {
 		Object source = null;
 		if (e != null) {
```
tivervac, line 291: Reverse logic and keep source in one assignment

miklossy: Fixed.

PR #1432, opened 2020-05-05
[eclipse/xtext#1679]ported test code to java

cdietrich:
[eclipse/xtext#1679]ported test code to java


org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/AbstractFSSynchronizationTest.java
```diff
@@ -0,0 +1,427 @@
...
+			build();
+			Assert.assertEquals("object Foo", readFile(output.getFile(new Path("Foo.txt"))));
+			File javaIoFile = output.getFile(new Path("Foo.txt")).getLocation().toFile();
+			javaIoFile.delete();
+			javaIoFile.getParentFile().delete();
+			file.touch(monitor());
+			build();
+			Assert.assertEquals("object Foo", readFile(output.getFile(new Path("Foo.txt"))));
+		} catch (CoreException e) {
+			throw Exceptions.sneakyThrow(e);
```
szarnekow, line 109: In tests it would be safe to throw new RuntimeException or Assert.fail.
But it's only test code, so I'm fine with this.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/AbstractFSSynchronizationTest.java
```diff
@@ -0,0 +1,427 @@
...
+		disableAutobuild(() -> {
+			cleanBuild();
+			Assert.assertEquals(expectedSize, ouputDirectory.list().length);
+			Assert.assertTrue(ArrayExtensions.contains(ouputDirectory.list(), "Lalala.txt"));
+		});
+	}
+
+	protected void testCleanUpDerivedResourcesWithCreateBetween(IContainer output) {
+		File outputDirectory = output.getLocation().toFile();
+		final int initialSize;
```
szarnekow, line 240: Some unnecessary final locals in this class, but again not crucial.

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/AbstractFSSynchronizationTest.java
```diff
@@ -0,0 +1,427 @@
...
+			Assert.assertTrue(javaIoFile.createNewFile());
+			return javaIoFile;
+		} catch (IOException e) {
+			throw Exceptions.sneakyThrow(e);
+		}
+	}
+
+	protected void setContent(File javaIoFile, String content) {
+		try {
+			FileOutputStream fos = new FileOutputStream(javaIoFile);
```
szarnekow, line 414: try-with-resource

org.eclipse.xtext.builder.tests/src/org/eclipse/xtext/builder/impl/AbstractFSSynchronizationTest.java
cdietrich: Local variable initialSize defined in an enclosing scope must be final or effectively final

PR #1451, opened 2020-05-13
Streamline BuildScheduler_ManyJobsName message text

kthoms:


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/messages.properties
```diff
@@ -1,7 +1,7 @@
 BuildScheduler_JobName=Updating projects
 BuildScheduler_SingleJobName=Updating project ''{0}''
 BuildScheduler_TwoJobsName=Updating projects ''{0}'' and ''{1}'' 
-BuildScheduler_ManyJobsName=Updating projects ''{0}'' and {1} others 
+BuildScheduler_ManyJobsName=Updating projects ''{0}'', ''{1}'' and others
```
cdietrich, line 4: message would be
Updating projects hugo, 2 and others

kthoms:
I stumbled over this code and misread the message. Sebastian is right, it is just to make projects singular. Updated.

Issue #1361, opened 2020-03-03
Keybinding conflicts occurred in Template Preference Page


cdietrich:
1. start runtime eclipse
2. open preferences and go to e.g. Xtend templates
3. close Preference Dialog
4. open the Preference dialog again

this will result in

```
Binding(F2,
	ParameterizedCommand(Command(org.eclipse.ui.edit.rename,Rename,
		Rename the selected item,
		Category(org.eclipse.ui.category.file,File,null,true),
		org.eclipse.ui.internal.WorkbenchHandlerServiceHandler@65b1dc67,
		,,true),null),
	org.eclipse.ui.defaultAcceleratorConfiguration,
	org.eclipse.ui.contexts.window,,,system)
Binding(F2,
	ParameterizedCommand(Command(org.eclipse.ui.edit.text.showInformation,Show Tooltip Description,
		Displays information for the current caret location in a focused hover,
		Category(org.eclipse.ui.category.textEditor,Text Editing,Text Editing Commands,true),
		org.eclipse.ui.internal.WorkbenchHandlerServiceHandler@c129b51,
		,,true),null),
	org.eclipse.ui.defaultAcceleratorConfiguration,
	org.eclipse.xtext.ui.embeddedTextEditorScope,,,system)
```

on the first call into
`org.eclipse.xtext.ui.editor.embedded.EmbeddedEditorActions.initialize()`

the active contexts are
`[org.eclipse.ui.contexts.dialogAndWindow, org.eclipse.ui.contexts.dialog]`
on the second

[...]

PR #1454, opened 2020-05-17
[#1361] Add missing "@since 2.22" tag.

miklossy:


cdietrich:
am not sure if i consider this api
@szarnekow what do you think?

miklossy:
The API-Diff report recognizes this as added API.

cdietrich:
ping @szarnekow

szarnekow:
Even though it's not API, a since tag documents when something was introduced. For this particular case, I'm fine with leaving this out.

PR #1455, opened 2020-05-19
Regenerate Xtext Test Languages, Examples, PureXbase.

miklossy:
- Regenerate plugin.xml_gen.
- Merge plugin.xml_gen into plugin.xml.


miklossy:
> no change in domain model?

Thanks for the hint! I was also wondering about that, but actually the generated plugin.xml_gen of the domain model example is not checked in the repository, that's why I missed that one. I will provide a follow up PR with that change.

PR #1457, opened 2020-05-20
[eclipse/xtext-core#1499] Stricter JavaDoc validation on org.eclipse.xtext.ui

szarnekow:
see eclipse/xtext-core#1499


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/model/ResourceForIEditorInputFactory.java
```diff
@@ -133,7 +133,6 @@ protected boolean isValidationDisabled(java.net.URI uri) {
 	}
 
 	/**
-	 * @param uri
```
cdietrich, line 136: ?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/XtextUIMessages.java
```diff
@@ -39,7 +39,6 @@ private XtextUIMessages() {
 	}
 
 	/**
-	 * @param key
```
cdietrich, line 42: so missing params are not complained?

szarnekow: So far only *present* comments are validated. Enabling this for missing params would be ... quite some effort to be fixed.

I'd like to approach this step-by-step.
Unfortunately there is no such thing as validating only the touched files.

org.eclipse.xtext.ui/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -67,7 +67,7 @@ org.eclipse.jdt.core.compiler.problem.missingHashCodeMethod=warning
 org.eclipse.jdt.core.compiler.problem.missingJavadocComments=ignore
 org.eclipse.jdt.core.compiler.problem.missingJavadocCommentsOverriding=disabled
 org.eclipse.jdt.core.compiler.problem.missingJavadocCommentsVisibility=public
-org.eclipse.jdt.core.compiler.problem.missingJavadocTagDescription=return_tag
+org.eclipse.jdt.core.compiler.problem.missingJavadocTagDescription=all_standard_tags
 org.eclipse.jdt.core.compiler.problem.missingJavadocTags=ignore
```
szarnekow, line 71: @cdietrich here is the setting missingJavadocTags=ignore but if present, validate them

PR #1465, opened 2020-05-28
[https://github.com/eclipse/xtext-core/issues/1506] Add reproducible

miklossy:
test case to the Xtext SimpleArithmetics Example (currently ignored).


miklossy:
@cdietrich @hannesN : I uploaded a test case to the Xtext SimpleArithmetics Example that reproduce the problem. I hope this will make debugging/fixing the problem more easy.

cdietrich:
therefore i created the mimimal grammar already @miklossy

cdietrich:
as we have a simple helo world grammar to reproduce we dont need something checked in

miklossy:
If @hannesN shares the same opinion, I will close this PR without merging it.

hannesN:
I think a unit test reproducing the problem should be added despite the grammar in the issue.

szarnekow:
Of course we'll have tests for that, but not as part of the example project.

cdietrich:
yes but not in a random example project that is never executed in ci.

miklossy:
If there will be some automated unit tests somewhere else in the Xtext code base than I am fine with closing this PR without merging it.

PR #1480, opened 2020-06-23
[bugzilla 564578] Fix wrong colors on colors preference page

heeckhau:
This fixes an issue on the syntax coloring preference page:
when one color is changed, all other colors were shown as black

The preference store of field editors should not be replaced when
when a color is changed. The preference store is already a chained
preference store that contains the internal store.

This fixes bugzilla 564578

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/fields/AbstractDetailsPart.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/fields/AbstractDetailsPart.java
@@ -104,8 +104,11 @@ public abstract class AbstractDetailsPart extends FieldEditorPreferencePage {
 		if (event.getSource() instanceof FieldEditor) {
 			FieldEditor fe = (FieldEditor) event.getSource();
 			internalStore.setValue(fe.getPreferenceName(), fe.getPreferenceStore().getDefaultString(fe.getPreferenceName()));
+			// store changed value in internalStore
 			fe.setPreferenceStore(internalStore);
 			fe.store();
+			// reset preference store of fieldEditor to chained store, so it also works for other token styles.
+			fe.setPreferenceStore(getPreferenceStore());
 		}
 		super.propertyChange(event);
 	}


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/fields/AbstractDetailsPart.java
```diff
@@ -104,7 +104,6 @@ public final void propertyChange(PropertyChangeEvent event) {
 		if (event.getSource() instanceof FieldEditor) {
 			FieldEditor fe = (FieldEditor) event.getSource();
 			internalStore.setValue(fe.getPreferenceName(), fe.getPreferenceStore().getDefaultString(fe.getPreferenceName()));
-			fe.setPreferenceStore(internalStore);
```
cdietrich, line 107: wonder why this was here in the first place

cdietrich: this changes seems to break the editor changing on preference change

heeckhau: I was to quick. I'll have another look.

cdietrich: looks better now. lets see what the build tells

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/fields/AbstractDetailsPart.java
```diff
@@ -104,8 +104,11 @@ public final void propertyChange(PropertyChangeEvent event) {
 		if (event.getSource() instanceof FieldEditor) {
 			FieldEditor fe = (FieldEditor) event.getSource();
 			internalStore.setValue(fe.getPreferenceName(), fe.getPreferenceStore().getDefaultString(fe.getPreferenceName()));
+			// store changed value in internalStore
 			fe.setPreferenceStore(internalStore);
 			fe.store();
+			// reset preference store of fieldEditor to chained store, so it also works for other token styles.
+			fe.setPreferenceStore(getPreferenceStore());
```
tivervac, line 111: Shouldn't this be done in a finally block? I assume store can throw either Core or IOExceptions

heeckhau: Maybe, but this is not done for similar code in this class.

I tried to keep the fix as small as possible

szarnekow:
Thank you, @heeckhau !

szarnekow:
cf https://bugs.eclipse.org/bugs/show_bug.cgi?id=564578

PR #1481, opened 2020-06-28
[eclipse/xtext#1812] Use centos-7 pod template & config. JDK

kthoms:
The centos-7 pod template provides a configuration suitable for most
builds. This makes the configuration of the custom container obsolete.

Additionally JDK versions are now selectable by parameter.

The nightly build uses 'latest' target and JDK 11.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,7 @@ pipeline {
   }
 
   parameters {
-    choice(name: 'target_platform', choices: ['oxygen', 'photon', 'r201809', 'r201812', 'r201903', 'r201906', 'r201909', 'r201912', 'r202003', 'r202006', 'latest' ], description: 'Which Target Platform should be used?')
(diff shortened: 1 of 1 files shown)


cdietrich:
if i build with java 11 from upstream (start in lib with trigger downstream), then eclipse 4.7 is picked, which makes no sense

kthoms:
So upstream (xtext-extras) should call downstream with a different TARGET_PLATFORM?
How to decide? If Java 11 is selected (and trigger downstream), then use latest target platform? Or a specific one? Or should the target platform be already choosable in xtext-lib (and the other upstreams) and propagate TARGET_PLATFORM to downstream?
Shall the build be aborted with combination Java11/Eclipse 4.7 by a check early?

cdietrich:
i think if there is no explicit target and java 11 then the target should become latest

kthoms:
@cdietrich @Hejado I have integrated now detecting whether the build was triggered by upstream with JDK11. The target platform will be set to 'latest'.
The other PRs are updated, too, and everything is building.
Please have a look.

Hejado (approved):

Issue #1447, opened 2020-05-11
Provide ui test infrastructure for testing multi quickfixes.


miklossy:
It would be useful to extend the Xtext ui test infrastructure to enable the comfortable testing of multi quickfix capabilities.

**Hint:** The `org.eclipse.xtext.ui.tests.editor.quickfix.AbstractMultiQuickfixTest` already provides such basic functionality. 

- [x] Move `AbstractMultiQuickfixTest` class to the `org.eclipse.xtext.ui.testing` package.
- [x] Define test cases based on that base class for the Xtend languages. https://github.com/eclipse/xtext-xtend/issues/1061
- [x] Define test cases based on that base class for one of the Xtext examples.
- [x] Verify that the testing API is properly designed. On demand refactor the abstract base class and the implemented tests cases to eliminate duplication and to improve readability. 
- [x] Add documentation to the [UI testing section](https://www.eclipse.org/Xtext/documentation/310_eclipse_support.html#ui-testing) on the Xtext website.

miklossy:
@drkstr101 : Thanks for offering you support on maintaining documentation, creating code samples, test cases, etc. It would be awesome if you could take care of this issue.

drkstr101:
My pleasure! I will take a look and see what I can do.

drkstr101:
Just a quick update...

I've freed up some time this week al will begin working on this now. I was unable to do the full build, but I believe it is just test failures, so I will move past it for now and attempt to make my changes by building the components individually. I will post another update once I've made some meaningful progress.

drkstr101:
Alright, so I've done the first and easiest of the 3 tasks I've understood in the description.

1. This class should be moved to the org.eclipse.xtext.ui.testing package

Which leaves;

2.  The code should be cleaned up 

I will hold off on this for a bit until I have a better understanding of what needs to be cleaned up, which leaves;

3. further test cases based on that base class should be added to the Xtext/Xtend code base to verify that the API is properly designed


Since this class is responsible for testing in itself, I assume this means I should build and package a release with the relocated class, and attempt to use it to implement my own tests using that feature. This will get me over the first hump of building the whole system and bundling/installing it into a fresh workspace for testing.

Good first issue indeed!  I will report back again once I have something more useful to report.

drkstr101:
[...]

PR #1486, opened 2020-07-05
[#1447] Add AbstractMultiQuickfixTest.testMultiQuickfix API method.

miklossy:
- Add a testMultiQuickfix() API method to the AbstractMultiQuickfixTest
class to provide an easy and compact way to test multi-quickfixes.

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractMultiQuickfixTest.java
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractMultiQuickfixTest.java
@@ -65,6 +65,29 @@ public abstract class AbstractMultiQuickfixTest extends AbstractEditorTest {
 		IResourcesSetupUtil.setAutobuild(WAS_AUTOBUILD);
 	}
 
+	/**
+	 * Test that the multi-quickfixes work as expected.
+	 * 
+	 * @param initialText
+	 *            the initial DSL text.
+	 * @param initialTextWithMarkers
+	 *            the initial DSL text enhanced by the recognized marker information.
+	 * @param resultTextWithMarkers
+	 *            the result DSL text enhanced by the recognized marker information after applying all the quickfixes.
+	 */
+	public void testMultiQuickfix(CharSequence initialText, CharSequence initialTextWithMarkers, CharSequence resultTextWithMarkers) {
+		// Given
+		IFile file = dslFile(initialText);
+		IMarker[] markers = getMarkers(file);
+		assertContentsAndMarkers(file, markers, initialTextWithMarkers);
+		
+		// When
+		applyQuickfixOnMultipleMarkers(markers);
+		
+		// Then
+		assertContentsAndMarkers(file, resultTextWithMarkers);
+	}
+
 	protected IFile dslFile(CharSequence content) {
 		return dslFile(getProjectName(), getFileName(), getFileExtension(), content);
 	}


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractMultiQuickfixTest.java
```diff
@@ -65,6 +65,29 @@ public static void afterClass() throws Exception {
...
+	/**
+	 * Test that the multi-quickfixes work as expected.
+	 * 
+	 * @param initialText
+	 *            the initial DSL text.
+	 * @param initialTextWithMarkers
+	 *            the initial DSL text enhanced by the recognized marker information.
+	 * @param resultTextWithMarkers
+	 *            the result DSL text enhanced by the recognized marker information after applying all the quickfixes.
+	 */
```
cdietrich, line 77: misses `@since` annotation

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractMultiQuickfixTest.java
```diff
@@ -65,6 +65,29 @@ public static void afterClass() throws Exception {
 		IResourcesSetupUtil.setAutobuild(WAS_AUTOBUILD);
 	}
 
+	/**
+	 * Test that the multi-quickfixes work as expected.
+	 * 
+	 * @param initialText
```
cdietrich, line 71: wonder why there need to be two initalText params

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractMultiQuickfixTest.java
miklossy: The entire [AbstractMultiQuickfixTest](https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractMultiQuickfixTest.java#L45)class is marked with the `@since 2.23` , so there is no need to add the since tag to the method.

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractMultiQuickfixTest.java
miklossy: The first parameter is for the DSL text, while the second one is for the DSL text enhanced by the recognized marker information. Such a call invocation could look like for example:

```xtend
	@Test def test001() {
		'''
			package foo
			class Foo {
				def m() {
					val a = new Object
					val b = new Object
					
					if (a == null) {}
					if (b != null) {}
				}
			}
		'''.testMultiQuickfix('''
			package foo
			class Foo {
				def m() {
					val a = new Object
					val b = new Object
					
					if (a <0<==>0> null) {}
					if (b <1<!=>1> null) {}
				}
			}
			-------------------------------
			0: message=The operator '==' should be replaced by '===' when null is one of the arguments.
			1: message=The operator '!=' should be replaced by '!==' when null is one of the arguments.
		''', '''
			package foo
			class Foo {
				def m() {
					val a = new Object
					val b = new Object
					
					if (a === null) {}
					if (b !== null) {}
				}
			}
			--------------------------
			(no markers found)
		''')
	}
```

cdietrich: but how is the marker information extracted from then? isnt it sufficient to have the second parameter only?

miklossy: If you have only the second parameter, than the additional marker information has to be removed from it to get the initial DSL text. That is really difficult in case of complex dsls, I think.

cdietrich: i still dont like this. at other places we remove markers too

drkstr101: Is the `--------------------------` separator a static constant, or variable? I am getting...

    expected:<...--------------------[]
    0: message=The oper...> but was:<...--------------------[---------------------------------------]
    0: message=The oper...>

...when [I attempted to consume this API](https://github.com/drkstr101/xtext-xtend/blob/43594bae47cde2a13fc4a3f8a289b26fbb796b20/org.eclipse.xtend.ide.tests/longrunning/src/org/eclipse/xtend/ide/tests/quickfix/EqualsWithNullMultiQuickfixTest.xtend#L66).

miklossy: It is dynamically calculated and have the same length as the longest line in the dsl.

drkstr101: What are everyone's thoughts on an API that works more like...?

```
val initialText = '''
    package foo
    class Foo {
        def m(Object a, Object b, Object c) {
            if(a == null || b != null || c === null) 0 else 1
        }
    }
'''
val expectedText = '''
    package foo
    class Foo {
        def m(Object a, Object b, Object c) {
            if(a == null || b != null || c === null) 0 else 1
        }
    }
'''
val markerText = '''
    0: message=The operator '==' should be replaced by '===' when null is one of the arguments.
    1: message=The operator '!=' should be replaced by '!==' when null is one of the arguments.
'''


testMultiQuickfix(initialText, expectedText, markerText)
```

Perhaps making the `markerText` value optional.

miklossy: Good idea! I was also thinking about providing more overloaded versions of the `testMultiQuickfix` method.
@cdietrich @szarnekow @kthoms @ArneDeutsch What do you think?

Issue #1494, opened 2020-07-28
Race Condition on loading index from multiple non workspace locking threads


cdietrich:
there is a Race Condition on loading index from multiple non workspace locking threads
```
CustomEMFBasedPersister.load() line: 87	
AbstractBuilderState$1.run(IProgressMonitor) line: 78	
Workspace.run(ICoreRunnable, ISchedulingRule, int, IProgressMonitor) line: 2292	
Workspace.run(IWorkspaceRunnable, ISchedulingRule, int, IProgressMonitor) line: 2317	
CustomClusteringBuilderState(AbstractBuilderState).load() line: 92	
CustomClusteringBuilderState(AbstractBuilderState).ensureLoaded() line: 107	
CustomClusteringBuilderState(AbstractBuilderState).isEmpty() line: 254	
SomeOtherUtil.lambda$0(IProgressMonitor) line: 33	
1321697288.run(IProgressMonitor) line: not available	
Job$2.run(IProgressMonitor) line: 185	
Worker.run() line: 63	
```

```
CustomBasedPersister.load() line: 87	
AbstractBuilderState$1.run(IProgressMonitor) line: 79	
Workspace.run(ICoreRunnable, ISchedulingRule, int, IProgressMonitor) line: 2292	
Workspace.run(IWorkspaceRunnable, ISchedulingRule, int, IProgressMonitor) line: 2317	
CustomClusteringBuilderState(AbstractBuilderState).load() line: 92	
CustomClusteringBuilderState(AbstractBuilderState).ensureLoaded() line: 107	
CustomClusteringBuilderState(AbstractBuilderState).isEmpty() line: 254	
DirtyStateAwareResourceDescriptions.isEmpty() line: 143	
SomeUtil.initialize() line: 319	
ExecutionHierarchyView.createViewer(Composite) line: 467	
ExecutionHierarchyView.createPartControl(Composite) line: 303	
CompatibilityView(CompatibilityPart).createPartControl(IWorkbenchPart, Composite) line: 154	
```

as you can see `(AbstractBuilderState).load() line: 92` both threads are in the `NO` case and call `workspace.run` with no scheduling rule. i assume we have to pass `ResourcePlugin.getWorkspace().getRoot()` there.

what do you thing @szarnekow? cc @iloveeclipse

szarnekow:
[...]

PR #1495, opened 2020-07-28
[#1494] fixed race condition on loading the Xtext index from multiple threads

cdietrich:
[#1494] fixed race condition on loading the Xtext index from multiple threads

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/AbstractBuilderState.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/AbstractBuilderState.java
@@ -21,6 +21,8 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.SubMonitor;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/builderState/AbstractBuilderState.java
```diff
@@ -68,17 +70,26 @@
...
 			IWorkspaceRunnable runnable = new IWorkspaceRunnable() {
 				@Override
 				public void run(IProgressMonitor monitor) throws CoreException {
 					if (!isLoaded) {
-						resourceDescriptionData = new ResourceDescriptionsData(persister.load());
-						if(storage2UriMapper instanceof IStorage2UriMapperExtension)
-							((IStorage2UriMapperExtension) storage2UriMapper).initializeCache();
-						isLoaded = true;
+						try {
+							lock.acquire();
```
szarnekow, line 83: If I'm not mistake, the lock should be aquired before the try block. Otherwise we'll try to release something in the finally block, if the acquisition failed from some reason.

szarnekow: Hmm weird, the documentation for ILock has an example that states otherwise...

Issue #1497, opened 2020-07-31
Provide ui test infrastructure for testing the code mining capabilities.


miklossy:
It would be useful to extend the Xtext ui test infrastructure to enable the comfortable testing of code mining capabilities.

miklossy:
Thanks @cdietrich for the rich reviews!

Fixed in 2.23

PR #1498, opened 2020-07-31
[#1497] Provide the AbstractCodeMiningTest abstract base class.

miklossy:
AbstractCodeMiningTest abstract class to provide a base infrastructure
for testing the code mining capabilities.
- Extend the Xtext Examples by CodeMiningTest test cases to
demonstrate the usage of the AbstractCodeMiningTest class.

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractCodeMiningTest.java
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractCodeMiningTest.java
@@ -1,0 +1,187 @@
+/*******************************************************************************
+ * Copyright (c) 2020 itemis AG (http://www.itemis.eu) and others.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ * 
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.xtext.ui.testing;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.jface.text.codemining.ICodeMining;
+import org.eclipse.jface.text.codemining.ICodeMiningProvider;
+import org.eclipse.jface.text.codemining.LineHeaderCodeMining;
+import org.eclipse.jface.text.source.ISourceViewer;
+import org.eclipse.xtext.resource.FileExtensionProvider;
+import org.eclipse.xtext.ui.editor.XtextEditor;
+
+import com.google.common.annotations.Beta;
+import com.google.inject.Inject;
+
+/**
+ * @author miklossy - Initial contribution and API
+ * @since 2.23
+ */
+@Beta
+public class AbstractCodeMiningTest extends AbstractEditorTest {
+
+	@Inject
+	protected FileExtensionProvider fileExtensionProvider;
+
+	@Inject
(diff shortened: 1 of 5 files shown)


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractCodeMiningTest.java
```diff
@@ -0,0 +1,123 @@
...
+		List<? extends ICodeMining> codeMinings;
+		try {
+			codeMinings = codeMiningProvider.provideCodeMinings(viewer, new NullProgressMonitor()).get();
+		} catch (InterruptedException | ExecutionException e) {
+			throw new RuntimeException(e);
+		}
+
+		int currentOffset = 0;
+
+		for (ICodeMining codeMining : codeMinings) {
```
cdietrich, line 82: does this work with multiple codeminings? does it work with multiple codeminings at the same place? does it work if the codeminings are not sorted by offset?

miklossy: Multiple codeminings are already supported, see org.eclipse.xtext.example.arithmetics.ui.tests.codemining.CodeMiningTest test class in the PR:
screenshot

I think, the other corner cases you mentioned are not supported, but the user has the possibility to override the `protected String insertCodeMinings(XtextEditor editor)` method to provide a custom implementation for such cases.

cdietrich: i dont think we should let that up to the user ....

cdietrich: the usual algorithm is to sort codeminings by offset backwards (apply them from bottom to top)

cdietrich: e.g. adding 

```
		for (Property f :  EcoreUtil2.eAllOfType(resource.getContents().get(0), Property.class)) {
			int endOffset = NodeModelUtils.findActualNodeFor(f).getEndOffset();
			acceptor.accept(createNewLineContentCodeMining(endOffset, " Breaking Tamas Test"));
		}
```

to domain model code mining provider easily breaks your code

miklossy:
I updated the PR by sorting the code mining according to their offset.

@cdietrich: Could you please review again?

miklossy:
I updated the PR to also work for multiple codeminings at the same offset.

@cdietrich: Could you please review again?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractCodeMiningTest.java
```diff
@@ -0,0 +1,182 @@
...
+			if (codeMining instanceof LineHeaderCodeMining) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	protected int getLastCodeMiningOnOffset(List<? extends ICodeMining> codeMinings, int offset) {
+		int i = 0;
+		for(;i<codeMinings.size();i++) {
```
cdietrich, line 157: formatting

miklossy: done

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractCodeMiningTest.java
miklossy: done

PR #1509, opened 2020-08-10
[xtext-eclipse#1412]: Xtext outline icons are confusing

drkstr101:
Task-Url: https://github.com/eclipse/xtext/issues/2383
Signed-off-by: Aaron R Miller <aaron.miller@waweb.io>

	Swapped out icon asset in FilterTerminalRulesContribution


cdietrich:
@szarnekow what do you think?

szarnekow (commented):
Thank you for the pull request and sorry for being a little late with the review.
In the outline, we use the three horizontal lines to denote rules. The porposed icon with the `[T]` looks arguably better suited for terminals than the icon we currently use.
Unfortunately I feel that the added icon for the filter does not really improve the consistency.
Either we also change the icon for the terminal rules in the outline to use the same `[T]`or we try to find a different icon for the "show / hide return type" action.

Another thought:
We could use the `[T]` as an adornment for terminal rules but keep the "main" rule icon. That would be a nice improvement I guess and it would allow to use the `[T]` for the filter action.

What do you think?

drkstr101:
No worries, I got wrapped up in some things myself. 

I can use whichever icon asset you like. Where you hoping for me to create one similar in design? I've neer done this before, but I don't think it would be to hard select a font and export the text `[T]` has a gif.

cdietrich:
ping @szarnekow

drkstr101:
I will just go ahead and create the icon asset then post it up here for review.

drkstr101:
@cdietrich @miklossy @szarnekow 

I was finally able to throw something together.

[TerminalsIcon]

This SVG can be imported into https://vectr.com/ for editing. I can rasterize it once approved.

Cheers!

cdietrich:
@szarnekow what do you think?

szarnekow:
I think it looks good but I'm not sure yet where this is heading to. Are we going to use adornment? Will this be the icon for the outline tree? What is your plan?

drkstr101:
I think the issue is that there is a single icon asset used to represent two separate actions, so one or the other needs to be swapped with something else; hence, an original icon asset was created. I don't really have any strong opinion one way or the other on it. I am just tossing something up on the board to move it forward.

szarnekow (approved):

cdietrich:
@szarnekow i dont think there was anything commited yet.
so which new icon should we commit?

szarnekow:
I like the proposed SVG. Please go ahead with that. No adornments for now.

cdietrich:
@drkstr101 any update here?

cdietrich:
moved out of 2.24

cdietrich:
move to 2.26

cdietrich:

PR #1513, opened 2020-08-14
added hint about Java 11 to new Xtext project wizard.

cdietrich:
added hint about Java 11 to new Xtext project wizard.

### DIFF
--- a/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/WizardNewXtextProjectCreationPage.java
+++ b/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/WizardNewXtextProjectCreationPage.java
@@ -196,7 +196,7 @@ public class WizardNewXtextProjectCreationPage extends WizardNewProjectCreationP
 		}
 		return candidate;
 	}
-
+	
 	@Override
 	protected boolean validatePage() {
 		if (!super.validatePage())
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/WizardNewXtextProjectCreationPage.java
```diff
@@ -314,4 +318,12 @@ public JavaVersion getJavaVersion() {
 		JavaVersion version = JavaVersion.fromBree(selected);
 		return version;
 	}
+	
+	@Override
+	public void setVisible(boolean visible) {
+		super.setVisible(visible);
+		if (visible) {
+			validatePage();
```
kthoms, line 326: Why do this is setVisible? Looks like an unexpected side effect here. Is there another place better suited to call it?
I guess the intention is here to validate when the page becomes active.

cdietrich: I don’t know the correct place to do that on there is nobody available to tell which one is better

kthoms: Adding `validatePage()` to `createControl` seems to do the same job. It is just the initial validation that you want to trigger, all further validation is already handled by editing the page.

cdietrich: thisis about page flipping

cdietrich:
obsolete for now

PR #1515, opened 2020-08-14
[J11] Update domain model to Java 11

cdietrich:
[J11] Update domain model to Java 11


kthoms:
What if the Domainmodel is installed on a machine with Java 8 only, or on a lower target?

cdietrich:
The domain model example also won’t work on oxygen. Using 2020-06 in the target would be bogus too.
All the platform approach of moving „somehow“ to java 11 is so bad

ArneDeutsch (approved):
LGTM

cdietrich:
obsolete for now

PR #1519, opened 2020-08-24
process ui events after test to avoid memory leak in tests

cdietrich:
process ui events after test to avoid memory leak in tests

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.java
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractHighlightingTest.java
@@ -12,7 +12,6 @@ import org.eclipse.core.resources.IFile;
 import org.eclipse.swt.custom.StyleRange;
 import org.eclipse.swt.custom.StyledText;
 import org.eclipse.swt.graphics.Color;
-import org.eclipse.swt.widgets.Display;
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractAutoEditTest.java
```diff
@@ -50,6 +51,15 @@ public void tearDown() throws Exception {
 		files.clear();
 		closeEditors();
 		super.tearDown();
+		waitForEventProcessing();
+	}
+	
+	/**
+	 * @since 2.23
+	 */
+	protected void waitForEventProcessing() {
```
szarnekow, line 60: Should this be called from `closeEditors`?

cdietrich: good question. or from the baseclass

szarnekow: I've a preference towards pushing `org.eclipse.xtext.ui.testing.AbstractHighlightingTest.waitForEventProcessing()` into the `AbstractWorkbenchTest` and calling it from its tearDown.

szarnekow (approved):

Issue #1463, opened 2020-05-26
annotations turn grey or go missing when executing "validate" from context menu


asheiduk:
The error/warning annotations in an open editor go missing or are greyed when explizit validation is triggered.

Steps to reproduce:
 - Use the "Hello World" project and enable the check.
 - Start guest eclipse.
 - Type a file like this:
```
        Hello Xtext!
        Hello foo!
```
- Save
- Notice that there is one warning annotation both in the left vertical ruler and in the right overview ruler.
- Open the context menu of the **editor**, select "Validate" (the one with the hover "Trigger expensive validation", not the other one) 
- Notice that the vertical ruler now contains only a grey warning triangle and the overview ruler is empty. The "Problems" view still contains the warning.

Further observation: It seems only "dirty" editors show this bug:

- Close the editor window and reopen it. The annotations are displayed.
- Select "Validate" again from the editor's context menu
- Now both annotations are still displayed.
- Type something and save
- Select "Validation"
- Notice that the annotations are missing again.

We see this behaviour in Xtext 2.14 (which my project uses right now.) but als in 2.21.

While debugging we came to `org.eclipse.xtext.ui.editor.validation.MarkerIssueProcessor` which seems to remove the markers/annotations and a reference to #462 which worked in the same area.

miklossy:
Thanks for your bug report @asheiduk ! I could also reproduce the problem with Xtext 2.22.0.v20200525-0122
![screencast](https://user-images.githubusercontent.com/20393472/82904688-19b39680-9f63-11ea-8e41-d22dbb224e8d.gif)

szarnekow:
Please note that "expensive" validations and the manual validate action are not meant to be used along with the XtextBuilder in the same project. If you intend to do that, the marker management is up to the language implementor.

asheiduk:
@szarnekow I must admitt that I don't understand what you are trying to tell me.

The hover says "expensive Validation" but it executes **all** checks and therefore
handles **all** errors.

Both the menu action and the build command are configured by default.
So I guess everyone assumes that this **should** work together out of the box. 

I have no idea what **project specific** code would be required to align
them both. I guess that code would be better handled by Xtext directly.

cdietrich:
ping @szarnekow

cdietrich:
we are affected  by a similar issue.
debugging
{{org.eclipse.ui.texteditor.ResourceMarkerAnnotationModel.batchedUpdate(IMarkerDelta[])}}
[...]

PR #1523, opened 2020-08-28
[#1463] fixed handling of fast check issues in validate action

cdietrich:
[#1463] fixed handling of fast check issues in validate action

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/handler/ValidateActionHandler.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/handler/ValidateActionHandler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2008, 2020 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/handler/ValidateActionHandler.java
```diff
@@ -47,17 +53,36 @@
...
-					CheckMode.ALL);
-			validationJob.schedule();
+			if(resource != null) {
+				IValidationIssueProcessor issueProcessor = new MarkerIssueProcessor(resource, xtextEditor.getInternalSourceViewer().getAnnotationModel(), markerCreator, markerTypeProvider);
+				IValidationIssueProcessor issueProcessorFastIssues = new AnnotationIssueProcessor(xtextDocument, xtextEditor.getInternalSourceViewer().getAnnotationModel(), issueResolutionProvider);
+				ValidationJob validationJob = new ValidationJob(resourceValidator, xtextDocument, new IValidationIssueProcessor() {
+					
+					@Override
+					public void processIssues(List<Issue> issues, IProgressMonitor monitor) {
+						List<Issue> fast = new ArrayList<Issue>();
```
szarnekow, line 65: Theoretically we'd need to do a SubMonitor.convert(monitor) with two ticks such that we can pass the monitor consistently into the child processors. Not sure if it matters a lot, though.

cdietrich: you mean 4 ticks tplit 3/1?

szarnekow: I'd be ok with a 1:1 split since I assume that both list will almost always have the same number or at least almost the same number of entries (most validations are FAST).

szarnekow: If we want to go all-in, that we use the combined sizes of both lists as the number of ticks.

cdietrich: updated

szarnekow (approved):
Thank you for the fix of the long-standing issue.

PR #1531, opened 2020-09-11
[eclipse/xtext#1843] Avoid NPE in XtextBuildConsole

frankbenoit:


szarnekow (approved):

kthoms (approved):
makes sense to check that

PR #1543, opened 2020-09-21
Avoid the quickfix to create a terminal fragment where the fragment usage would be illegal

nbhusare:
Removed invalid terminal-fragment quick-fix proposal and the corresponding unit-test


szarnekow (approved):

cdietrich:
thx @nbhusare

PR #1544, opened 2020-09-21
[eclipse/xtext#1851] Add more Junit5 bundles to TP

szarnekow:
see eclipse/xtext#1851

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -785,6 +785,9 @@
             name="org.junit.platform.runner"
             versionRange="[1.1.0,1.7.0)"/>
         <requirement
+            name="org.junit.platform.suite.api"
+            versionRange="[1.1.0,1.7.0)"/>
+        <requirement
             name="org.opentest4j"
             versionRange="[1.0.0,1.3.0)"/>
         <requirement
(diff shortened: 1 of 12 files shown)


szarnekow:
@cdietrich what do you think?

cdietrich:
do we need it in oomph too?

cdietrich:
+ can you do same for xtend?

szarnekow:
> * can you do same for xtend?

The Xtend target platform does not list anything about Junit5. Should we start doing that?

cdietrich:
have no idea why @kthoms any idea?

szarnekow:
> do we need it in oomph too?

Done

szarnekow:
> have no idea why @kthoms any idea?

Technically we don't need it from Xtend since there shouldn't be real dependency. We do have `org.eclipse.xtend.core.validation.XtendValidator.junitAnnotations` as plain strings though.

cdietrich:
Bildschirmfoto vom 2020-09-22 13-30-50

szarnekow:
True, but we don't run through the code paths for Junit5, do we?

cdietrich:
yes.

cdietrich:
but do we in eclipse?

szarnekow:
Hmm fair point. We actually don't.
Why is it in the TP?

cdietrich:
do we have hybrid base classes with `@beforeeach`  somewhere?

szarnekow:
Indeed we do. I was looking for `@Test` usages ... but the BeforeEach lead to the right track.

kthoms (approved):
I'm fine with those additions.

cdietrich:
so shoudl we simply add it to xtend too to be on the safe side?

szarnekow:
> so shoudl we simply add it to xtend too to be on the safe side?

Yes. It'll also help to avoid exactly the confusion that we currently have.

cdietrich (approved):

PR #1549, opened 2020-10-01
true and false keywords are highlighted as such when used as ValidId

nbhusare:
Enhanced the SemanticHighlightingCalculator to handle the highlighting of keywords used as a feature in the assignment


org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/syntaxcoloring/SemanticHighlightingCalculator.java
```diff
@@ -104,4 +105,13 @@ public INode getFirstFeatureNode(EObject semantic, EStructuralFeature feature) {
 			return nodes.get(0);
 		return null;
 	}
+	
+	@Override
+	protected boolean highlightElement(EObject object, IHighlightedPositionAcceptor acceptor, CancelIndicator cancelIndicator) {
```
szarnekow, line 110: Could you shift this into an else block for the condition in line 92? e.g. `if (SPECIAL...) .. else { default }` but *only* if the text is `true` or `false` to reduce the number of emitted highlighting tokens?

cdietrich:
please also check for test fails

org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/editor/syntaxcoloring/SemanticHighlightingCalculator.java
nbhusare: Done.

nbhusare:
@cdietrich @szarnekow Tx for the review.

PR #1557, opened 2020-10-14
more explicit exception treatment in example tests

cdietrich:
more explicit exception treatment in example tests


szarnekow (approved):
Appreciated!

Issue #1566, opened 2020-10-23
WorkspaceFileGenerator uses wrong encoding for StringInputStream


hannesN:
We use a NewFileWizard which is supposed to create a file containing an UTF-8 String.
However, the FileGenerator is creating a StringInputStream with the default charset, on windows "windows-1252" therefore creating invalid input.

I suggest using the charset specified in the IFile to create. 
e.g.
```
is = new StringInputStream(fileEntry.getValue().toString(), file.getCharset());
file.create(is, true, subMonitor);
``` 

I tested it with my code and it works for me.

cdietrich:
can you please provide a PR

cdietrich:
cc @ArneDeutsch

hannesN:
I created a PR.

cdietrich:
fixed in 2.24

PR #1568, opened 2020-10-23
[#1566] set encoding to StringInputStream

hannesN:
Added the encoding to the StingInputStream

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/wizard/template/WorkspaceFileGenerator.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/wizard/template/WorkspaceFileGenerator.java
@@ -8,6 +8,7 @@
  *******************************************************************************/
 package org.eclipse.xtext.ui.wizard.template;
 
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Map;
 import java.util.SortedMap;
@@ -18,9 +19,12 @@ import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.ui.actions.WorkspaceModifyOperation;
+import org.eclipse.xtext.ui.internal.Activator;
 import org.eclipse.xtext.util.StringInputStream;
 
 import com.google.common.annotations.Beta;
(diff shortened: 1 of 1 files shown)


cdietrich:
@hannesN can you please sign off the commit and also sign the eca if not done yet

cdietrich:
LGTM @ArneDeutsch can you please also have a look

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/wizard/template/WorkspaceFileGenerator.java
```diff
@@ -49,12 +54,15 @@ protected void execute(IProgressMonitor monitor) throws CoreException, Invocatio
...
 			for (Map.Entry<String, CharSequence> fileEntry : files.entrySet()) {
 				IFile file = workspace.getRoot().getFile(new Path(fileEntry.getKey()));
-				file.create(new StringInputStream(fileEntry.getValue().toString()), true, subMonitor);
+				file.create(new StringInputStream(fileEntry.getValue().toString(), file.getCharset()), true, subMonitor);
 				if (firstFile == null) {
 					firstFile = file;
 				}
 				subMonitor.worked(1);
 			}
+		} catch (UnsupportedEncodingException e) {
```
cdietrich, line 63: as the api is still beta, should we change it to throws Exception @ArneDeutsch ?
then we would not need these

hannesN: I wouldn't do that. As a user I can't set the encoding so what should I do if the encoding is not supported. Throwing a Core exception should be enough.

cdietrich: well. this is about geneneral exception handling.
throws CoreException, InvocationTargetException, InterruptedException {
=> replace this with "throws Exception" as we dont care about the other exceptons neither

hannesN: No because then please only throw RuntimeExceptions :-)
Seriously, I would encapsulate the incovationTargetException and InterruptedException as well in an core Exception and only throw this, because this is handled by the eclipse frame work (with an error dialog if I remember correctly).
Enforcing a lot of catches just blows the code using the method and I mostly just throw a runtimeException in this cases.

hannesN: Another thought: Why would I need 3 different exception, which tell me: "the generation failed!" I have to deal with that, I don't want to catch different types of reasons. If I relly need to know them, e.getReason() is the method to go. But usually I need to do some clean-up or other stuff.
More important than different exception is a good javadoc, which is still missing here.

cdietrich: in this case we have explicit handing for some at caller place

szarnekow: The `#execute` method is inherited from `WorkspaceModifyOperation` - we cannot add a checked exception to the signature.
I think wrapping it in a CoreException is the right move + some JavaDoc on the failure scenarios.

szarnekow (approved):

cdietrich (approved):

cdietrich:
@ArneDeutsch i am gonna merge this now. please do have a look anyway

cdietrich:
thx @hannesN

PR #1571, opened 2020-10-26
Bug 568152 - Use FrameworkUtil#getBundle instead of Platform#getBundle

kthoms:
Regenerated test languages.


kthoms:
Some languages are missing. I will reopen the PR once this is done.

Issue #1589, opened 2020-11-17
FixedScopedPreferenceStore: Fire property change event only on value changes


kthoms:
I just run into a loop while analyzing the change for eclipse/xtext-xtend#1078 (dark theme). There I observed that property change events were issued while the property value did not change.

The IPreferenceChangeListener should only fire a PropertyChangeEvent when the value is changing.

<img width="569" alt="Bildschirmfoto 2020-11-17 um 06 03 51" src="https://user-images.githubusercontent.com/265597/99349076-5bcaee80-289b-11eb-8807-e3f66428db97.png">

cdietrich:
@kthoms should we move to 2.25?

cdietrich:
move to 2.25

cdietrich:
move to 2.26

PR #1590, opened 2020-11-17
[#1589] Fire PropertyChangeEvent only when value is changing

kthoms:

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/FixedScopedPreferenceStore.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/FixedScopedPreferenceStore.java
@@ -380,7 +380,7 @@ public class FixedScopedPreferenceStore extends EventManager implements IPersist
 		// important: create intermediate array to protect against listeners
 		// being added/removed during the notification
 		final Object[] list = getListeners();
-		if (list.length == 0) {
+		if (list.length == 0 || Objects.equal(oldValue, newValue)) {
 			return;
 		}
 		final PropertyChangeEvent event = new PropertyChangeEvent(this, name,


cdietrich:
please rebase

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/FixedScopedPreferenceStore.java
```diff
@@ -380,7 +380,7 @@ public void firePropertyChangeEvent(String name, Object oldValue,
 		// important: create intermediate array to protect against listeners
 		// being added/removed during the notification
 		final Object[] list = getListeners();
-		if (list.length == 0) {
+		if (list.length == 0 || Objects.equal(oldValue, newValue)) {
```
cdietrich, line 383: this differs from other pref stores impls. eg.
ScopedPreferenceStore where this seems to be c&ped from

szarnekow: Also my impression. If in doubt, better change the listener impl that does too many things when the event is indicating a no-op.

Please note that platform classes have logic like this:
```
if (changed > 0)
			masterPreferenceStore.firePropertyChangeEvent(getPreferencePrefix(), null, null);
```
This event would be swallowed with the proposed change.

szarnekow (changes requested):
Please investigate further why you see an endless loop.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/preferences/FixedScopedPreferenceStore.java
kthoms: Thanks for the pointers. I'll have a second look.

cdietrich:
@kthoms ok with move to 2.25?

cdietrich:
move to 2.25

cdietrich:
cleared milestone. feel free to readd if you are working on this again @kthoms

kthoms:
Let's close this for now. I am not able to reproduce this ATM.

PR #1662, opened 2021-03-30
make Storage2UriMapperJavaImpl better customizable

cdietrich:
make Storage2UriMapperJavaImpl better customizable


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
```diff
@@ -471,7 +471,7 @@ public void setWorkspace(IWorkspace workspace) {
 	/**
 	 * @since 2.4
 	 */
-	private void updateCache(IJavaProject project) {
+	protected void updateCache(IJavaProject project) {
```
szarnekow, line 474: While we're at it: Should we make the (injected) fields protected, too (at least the ones that don't have an accessible getter)?

szarnekow (approved):

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
cdietrich: what about the other methods?

szarnekow (approved):

PR #1666, opened 2021-04-09
expose org.eclipse.xtext.ui.wizard.template as API

cdietrich:
expose org.eclipse.xtext.ui.wizard.template as API
Fixes eclipse/xtext#1962


szarnekow (approved):
Thank you @cdietrich

PR #1670, opened 2021-04-26
[eclipse/xtext#1952] converted example test code to java

cdietrich:
[eclipse/xtext#1952] converted example test code to java


org.eclipse.xtext.xtext.ui.examples/projects/fowlerdsl/org.eclipse.xtext.example.fowlerdsl.tests/src/org/eclipse/xtext/example/fowlerdsl/tests/StatemachineFormatterTest.java
```diff
@@ -0,0 +1,317 @@
...
+/**
+ * @author miklossy - Initial contribution and API
+ */
+@RunWith(XtextRunner.class)
+@InjectWith(StatemachineInjectorProvider.class)
+public class StatemachineFormatterTest {
+	@Inject
+	private FormatterTestHelper formatterTestHelper;
+	
+	private static final String NL = System.lineSeparator();
```
cdietrich, line 29: i am not sure if i like this.
we could also
- call toUnitLineSeparator on all cases
- uses internal MapBasedPreference to configure newLine
- have a FormatterTestHelper that does does it internal

@szarnekow @miklossy what do you think?

szarnekow: I prefer option (3)

szarnekow: Another one would be to use format strings with `%n` but it's kind of nasty since it still needs string concat.

cdietrich: created https://github.com/eclipse/xtext-core/issues/1703 for the map based preference values api problem

cdietrich: maybe something like

```
public class FormatterTestHelper {

	public static class Platform extends FormatterTestHelper {
		@Override
		public void assertFormatted(FormatterTestRequest req) {
			req.setToBeFormatted(Strings.toPlatformLineSeparator(req.getToBeFormatted()));
			CharSequence expectation = req.getExpectation();
			if (expectation != null) {
				req.setExpectation(Strings.toPlatformLineSeparator(expectation));
			}
			super.assertFormatted(req);
		}
	}
```

szarnekow: Makes sense to me. Is the rewrite of the request itself necessary?

cdietrich: the problem is that the original api is very bad to intercept as it is
public void assertFormatted(Procedure1<FormatterTestRequest> init) {
maybe you have a better idea?

LorenzoBettini: IMHO: it is crucial that such tests are easy to write and maintain. If you use the JUnit view failure of an assertEquals with strings, you can copy the actual value (with newlines) and paste it in a Java editor and Eclipse will add the corresponding string concatenation and the `\n` character at the end. If I don't use Xtend I keep the concatenated string generated by Eclipse in the editor and have a utility function to deal with possible different line endings in different operating systems. Using a solution with explicit NL makes these tests hard to write and hard to maintain...

cdietrich: yes. an approach using the preferences (when made api) could be

```
	@Inject
	private FormatterTestHelper formatterTestHelper;
	
	public void assertFormatted(Procedure1<FormatterTestRequest> init) {
		formatterTestHelper.assertFormatted(new Procedure1<FormatterTestRequest>() {

			@Override
			public void apply(FormatterTestRequest p) {
				init.apply(p);
				p.preferences(m -> m.put(FormatterPreferenceKeys.lineSeparator, "\n"));
				
			}
		});
	}
	```
	
	then  it should work with "\n" as copy&pastable too

cdietrich: this change also would be test local

szarnekow: Looks way better to me.

cdietrich:
please re-review

szarnekow (approved):

PR #1671, opened 2021-05-03
Make sure the auto build is triggered after a force kill

tivervac:
When an Xtext application is force killed, the Xtext will not have been
written (as it's written on shutdown). The applications thus forgets
the last built state during the next startup. The idea is that
the auto-builder will then pick up that the projects need to be rebuild.

However, for the latter we need to first touch the relevant projects.

PR #1678, opened 2021-05-28
[#1671]always run in job to avoid deadlock and check accessibility in job

cdietrich:
[#1671]always run in job to avoid deadlock and check accessibility in job


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuilderStateDiscarder.java
```diff
@@ -68,7 +68,9 @@ protected void touchProject(IProject project) {
 		} else {
 			Job touchJob = Job.create("Touch project " + project.getName(), progressMonitor -> {
 				try {
-					project.touch(progressMonitor);
+					if (project.isAccessible()) {
+						project.touch(progressMonitor);
```
tivervac, line 72: Looking at the call hierarchy, no one is making sure that the projects that are passed in are accessible. Thus it's probably a good idea to also wrap either the entirety of the function in a `project.isAccessible()` or to wrap line 64 with `project.isAccessible()`

cdietrich:
@tivervac @szarnekow please review. 
or should i undo the revious commit for now?

szarnekow (approved):

PR #1680, opened 2021-06-01
[eclipse/xtext-xtend#1198] adapt message for KeepLocalHistory with hint on new global flag

cdietrich:
[eclipse/xtext-xtend#1198] adapt message for KeepLocalHistory with hint on new global flag

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/messages.properties
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/messages.properties
@@ -17,6 +17,6 @@ OutputConfigurationPage_Directory=Directory:
 OutputConfigurationPage_IgnoreSourceFolder=Ignore?
 OutputConfigurationPage_OutputDirectory=Output directory
 OutputConfigurationPage_OverrideExistingResources=Overwrite existing files
-OutputConfigurationPage_KeepLocalHistory=Keep local history for generated files
+OutputConfigurationPage_KeepLocalHistory=Keep local history for generated files. Please note: From Eclipse 2021-06\nthis preference is overruled by the global "History for derived files" preference.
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/messages.properties
```diff
@@ -17,6 +17,6 @@ OutputConfigurationPage_Directory=Directory:
 OutputConfigurationPage_IgnoreSourceFolder=Ignore?
 OutputConfigurationPage_OutputDirectory=Output directory
 OutputConfigurationPage_OverrideExistingResources=Overwrite existing files
-OutputConfigurationPage_KeepLocalHistory=Keep local history for generated files
+OutputConfigurationPage_KeepLocalHistory=Keep local history for generated files. Please note: From Eclipse 2021-06\nthis preference is overruled by the global "History for derived files" preference.
```
cdietrich, line 20: @miklossy @kthoms @ArneDeutsch do you know if there is a easy way to find out the platform version eclipse is running on? this way i could introduce a second message and a toggle in the BuilderConfigurationBlock

cdietrich:
see also https://github.com/eclipse/xtext-xtend/pull/1199

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/preferences/messages.properties
ArneDeutsch: Think the platform version is difficult, since users could configure eclipse themself, even without specific product. Probably detect the version of one or a set of bundles to make the decision!? `Platform.getBundle("org.eclipse.core.runtime").getHeaders().get( "Bundle-Version"); `or something along the lines ....

ArneDeutsch (approved):

PR #1687, opened 2021-06-11
Lb task 1686 simplify domainmodel

### LINKED ISSUE #1686: Simplify Domainmodel example target platform
The `.target` file of the Domainmodel example is exceedingly and uselessly complex: it uses `slicer` instead of `planner`, forcing to list all requirements, without any benefit. I guess that happened when @kthoms introduced several target platforms using the TPD DSL (something that was then removed). The TP should use planner and list many less features. I'll provide a PR

LorenzoBettini:
Closes #1686 

I've tested the Maven build locally and it works also with tests


LorenzoBettini:
Note that it might conflict with https://github.com/eclipse/xtext-eclipse/pull/1684

cdietrich:
Can you also pr this here so that we know it works 
https://github.com/itemis/xtext-reference-projects/tree/master/domainmodel/2.26.0

LorenzoBettini:
> Can you also pr this here so that we know it works
> https://github.com/itemis/xtext-reference-projects/tree/master/domainmodel/2.26.0

Done: https://github.com/itemis/xtext-reference-projects/pull/181

org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/tp/domainmodel.target
```diff
@@ -1,37 +1,28 @@
...
     </location>
-    <location includeMode="slicer" includeAllPlatforms="true" includeSource="true" includeConfigurePhase="false" type="InstallableUnit">
-      <unit id="org.objectweb.asm" version="9.1.0.v20210209-1849"/>
-      <unit id="org.objectweb.asm.tree" version="9.1.0.v20210209-1849"/>
-      <unit id="com.google.inject" version="3.0.0.v201605172100"/>
-      <unit id="org.apache.commons.cli" version="1.2.0.v201404270220"/>
-      <unit id="com.google.guava" version="30.1.0.v20210127-2300"/>
-      <unit id="io.github.classgraph" version="4.8.35.v20190528-1517"/>
+    <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+      <!--
```
cdietrich, line 21: an not sure (have no idea) how properly maintained our manifests are
and what would be fixed if we have
version x in manifest (no range)
and orbit has x and x+1

cdietrich: if so should we also change the wizard?

LorenzoBettini: @cdietrich what do you mean by "change the wizard"? Do you mean the one that generates a new Xtext project? That was my next proposal ;)
If you have version x without no upper limit then it means that when the bundle will be installed it will use any latest version, so having a fixed version in the TP would probably be even worse, because you built it with a known version but it will be installed with a probably more up to date version.
Indeed, for objectweb.asm we have very strict version ranges in the MANIFEST, IIRC.

cdietrich: https://github.com/eclipse/xtext-core/blob/857b6e2efbc350a7cda1b0ff22a72c300d1d9dd5/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TargetPlatformProject.xtend

unfortunately i really have no idea on target platforms and thus cannot judge if this is a good idea

LorenzoBettini: If you use "0.0.0" in the TP then the latest version will be used that respects the version ranges in features and bundles.
In any case, that's only for compiling. As I said above, if a feature or bundle is known NOT to work with a specific newer version, then a version range must be set explicitly in the feature or bundle: having a fixed version in the TP will only make sure you compile the code with that fixed version, but then users will still be able to install your feature or bundle with a newer (not working version) if you haven't specified the range correctly.

cdietrich: i really would like a second or 3rd opinion here @kthoms @miklossy @ArneDeutsch

LorenzoBettini: No problem, let's wait :)
For sure, "slicer" should be avoided, but for the other parts let's see

LorenzoBettini:
> can you please rebase

@cdietrich 
Done, and I've run the Maven build locally, making sure it actually takes Google Guice 5.

LorenzoBettini:
Has this been merged somewhere else?

cdietrich:
no

LorenzoBettini:
and so? should I rebase it one more time?

cdietrich:
i still have doubts about non ranges in our plugins and new stuff show up in orbit and co. but i am not a tycho expert.

LorenzoBettini:
feel free to close it

org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/tp/domainmodel.target
szarnekow: Looks like there won't be a second opinion coming soonish. 

I can see why this makes sense for an example project but for "production" mode I'd rather pin the dependencies. Which raises the question if we want the examples to be different from what I personally would consider reliable practise for production environments.

LorenzoBettini: As you prefer. For sure, pinning dependency versions help make builds reproducible.
In any case, "slicer" makes no sense ;)

szarnekow:
I'm inclined to close this as there doesn't seem to be an agreement on how to do this.

@LorenzoBettini can you help me understand the difference between slicer and planner?

LorenzoBettini:
@szarnekow planner means that transitive dependencies are automatically downloaded, while slicer leaves transitive dependencies to your responsibility. Unless you specify a target platform for RAP, the only other reason to use slicer I can think of is masochism ;)

szarnekow:
Does this include optional transitive dependencies, too?

LorenzoBettini:
> Does this include optional transitive dependencies, too?

As far as I know, yes.
Note also that all the .target files used for building Xtext use "planner" (see, e.g., https://github.com/eclipse/xtext-eclipse/blob/master/releng/org.eclipse.xtext.target/org.eclipse.xtext.target-latest.target).

If I understand correctly, the "slicer" in the Domainmodel example comes from using the target platform DSL to create it (maybe @kthoms did that some time ago?)

Another piece of information: PDE has always had a (big) limitation: if you want to have a target platform for all platforms (e.g., to create a product or application for Linux, Windows and macOS), you cannot use "planner" and must revert to "slicer" (that also might the be the cause of "slicer" in this example: it has been created with the Target editor by specifying "use all platforms"). If you don't need to build a product, there's no need to have includeAllPlatforms="true" (which, as I said, would force you to revert to "slicer"). Even if you need to build a product, nowadays, you don't use PDE wizards to export the products: you use Tycho, which handles the "all platforms" for you automatically and ignores the includeAllPlatforms="false" (and indeed, Tycho lets you specify a reduced set of possible platforms in its plugin configuration). So, there's really no point in using "slicer" (as I said, the only case is a target platform for RAP applications because you must have full control of what gets into the target platform: you can't have UI bundles inserted automatically in the target platform: you only need the RAP.UI bundles).

cdietrich:
i am not sure if the domain model examples is to test slicer  and wizard uses planner 
https://github.com/eclipse/xtext-core/blob/e93760c27557a94cf8e6f0647a36dcc2654cff7e/org.eclipse.xtext.xtext.wizard/src/org/eclipse/xtext/xtext/wizard/TargetPlatformProject.xtend#L58

mickaelistria:
I still believe "slicer" is to be preferred whenever possible; it performs better and introduce less magic (aka "room for bugs")

LorenzoBettini:
@cdietrich @szarnekow it's your call now. As I said, I'm 95% sure the use of slicer was not intentional

szarnekow:
> I still believe "slicer" is to be preferred whenever possible; it performs better and introduce less magic (aka "room for bugs")

Looking at other projects, e.g. lsp4e and wildwebdeveloper, planner seems favorable, though. No?

org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/tp/domainmodel.target
```diff
@@ -1,38 +1,28 @@
...
-      <unit id="org.objectweb.asm.tree" version="9.2.0.v20210813-1119"/>
-      <unit id="com.google.inject" version="5.0.1.v20210324-2015"/>
-      <unit id="org.aopalliance" version="1.0.0.v201105210816"/>
-      <unit id="org.apache.commons.cli" version="1.2.0.v201404270220"/>
-      <unit id="com.google.guava" version="30.1.0.v20210127-2300"/>
-      <unit id="io.github.classgraph" version="4.8.35.v20190528-1517"/>
+    <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+      <!--
+        Just mention the p2 repo so that additional requirements with ranges specified in
+        features and bundles are taken from Orbit (e.g., objectweb.asm)
```
szarnekow, line 23: Do we want to stick to using the orbit update site?

LorenzoBettini:
> Thank you for the input and the suggestions @LorenzoBettini @mickaelistria
> 
> Since the domain-model example's target platform file is the only one in the entire Xtext codebase that uses slicer, I'm in favor of simplifying it to planner. @cdietrich would you be fine with it? @LorenzoBettini in case Christian agrees, could you please update the merge request?

Yes, for sure. I can also try to further simplify it (e.g., IIRC, PDE implies JDT)

org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel.releng/tp/domainmodel.target
LorenzoBettini: I guess it depends here: if we update to the next Eclipse version and the Orbit contents of the next Eclipse version are fine, then we should be safe.

cdietrich (approved):

LorenzoBettini:
@cdietrich I see you approved.
Shall we merge?

cdietrich:
yes

Issue #1701, opened 2021-07-19
ClusteringBuilderState.queueAffectedResources does not gives details in progress view


cdietrich:
ClusteringBuilderState.queueAffectedResources does not gives details in progress view

<img width="568" alt="Bildschirmfoto 2021-06-01 um 14 20 44" src="https://user-images.githubusercontent.com/238736/126158145-5987518c-5e29-47b3-8650-688aa72080ee.png">

cdietrich:
after change

![Bildschirmfoto 2021-07-19 um 14 19 36](https://user-images.githubusercontent.com/238736/126158772-ee60b447-1245-4614-bd0a-9b068aeffc5c.png)

cdietrich:
done with a first version. feels free to PR a different message if you dont like it.

PR #1702, opened 2021-07-19
[#1701] better progress reporting for queueAffectedResources

cdietrich:
[#1701] better progress reporting for queueAffectedResources

Bildschirmfoto 2021-07-19 um 14 19 36


miklossy (approved):

cdietrich:
@szarnekow  what do you think about the change? do you have a better proposal for the message?

szarnekow:
`Checking if abc.ext is affected`? `Checking if abc.ext must be compiled`? 
Should we have a getter for the message that can be overridden with little effort? E.g. a product based on Xtext may use a different terminology in the UI for the build.

cdietrich:
we dont have getters for any of the messages so far.
replaced evaluating with checking

cdietrich:
please feels free to change if you see a better message here

PR #1706, opened 2021-07-21
dont duplicate deps. that are already part of add. bundles in build.properties

cdietrich:
dont duplicate deps. that are already part of add. bundles in build.properties

### DIFF
--- a/org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel/pom.xml
+++ b/org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel/pom.xml
@@ -42,26 +42,6 @@
 				</configuration>
 				<dependencies>
 					<dependency>
-						<groupId>org.eclipse.emf</groupId>
-						<artifactId>org.eclipse.emf.mwe2.launch</artifactId>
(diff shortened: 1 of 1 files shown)


cdietrich:
for testing: https://github.com/itemis/xtext-reference-projects/tree/cd_addBundlesOnlyTest

org.eclipse.xtext.xtext.ui.examples/projects/domainmodel/org.eclipse.xtext.example.domainmodel/pom.xml
```diff
@@ -41,26 +41,6 @@
...
-						<groupId>org.eclipse.xtext</groupId>
-						<artifactId>org.eclipse.xtext.xtext.generator</artifactId>
-						<version>${xtextVersion}</version>
-					</dependency>
-					<dependency>
-						<groupId>org.eclipse.xtext</groupId>
-						<artifactId>org.eclipse.xtext.xbase</artifactId>
-						<version>${xtextVersion}</version>
-					</dependency>
 					<dependency>
```
LorenzoBettini, line 44: Please let's solve the comments on https://github.com/eclipse/xtext-core/pull/1723 first, OK?
By the way, I had proposed a simplification of the domainmodel TP as well some time ago, maybe we should consider that as well when experimenting with this?

PR #1709, opened 2021-07-29
name parallel loader pool

cdietrich:
name parallel loader pool

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
@@ -43,6 +43,7 @@ import org.eclipse.xtext.util.Tuples;
 import com.google.common.annotations.Beta;
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 /**
  * Loads resources in one more separate threads.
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
```diff
@@ -157,7 +158,7 @@ protected ResourceSet initialValue() {
 					return resourceSet;
 				}
 			};
-			this.executor = Executors.newFixedThreadPool(nThreads);
+			this.executor = Executors.newFixedThreadPool(nThreads, new ThreadFactoryBuilder().setNameFormat("ParallelLoad-%d").build());
```
cdietrich, line 161: am not sure if we can easily find out if this is cross linking or indexing loader here.

szarnekow: Slightly unrelated: Should we reuse the executor for all subsequent builds? Should this be made available from the outside?

szarnekow (approved):

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
szarnekow: > find out if this linking or indexing

I don't think that's possible. We'd need an additional key that is passed from the outside.

cdietrich: > Should this be made available from the outside?
thats something to think about. but this will lead to even more permantent threads.

cdietrich: (delegatingcontentassistcontextfactory already eats up numberoflangsx3

PR #1716, opened 2021-08-10
open up ParallelLoadOperation

cdietrich:
open up ParallelLoadOperation

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
@@ -162,6 +162,34 @@ public class ParallelResourceLoader extends AbstractResourceLoader {
 			this.waitTime = getTimeout();
 		}
 
+		/**
+		 * @since 2.26
+		 */
+		protected int getToProcess() {
+			return toProcess;
+		}
+
+		/**
+		 * @since 2.26
+		 */
+		protected void setToProcess(int toProcess) {
+			this.toProcess = toProcess;
+		}
+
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/resourceloader/ParallelResourceLoader.java
```diff
@@ -162,6 +162,34 @@ protected ResourceSet initialValue() {
 			this.waitTime = getTimeout();
 		}
 
+		/**
+		 * @since 2.26
+		 */
+		protected int getToProcess() {
+			return toProcess;
```
szarnekow, line 169: Either these methods are marked as final or we use them internal to the current class, too. Otherwise overrides wouldn't work.

cdietrich: they are intentionally final. but i am not sure if calling them all over internal has neg. perf. impact.

cdietrich: https://github.com/eclipse/xtext-eclipse/pull/1718/files

am not sure if the class can be sensibly subclassed as besides calling the getters
- other stuff is private and final
- init is done in constructor

szarnekow: I think you convinced me. Let's add the getters but keep the internal impl. Interested parties could very well override the "real" method and custom stuff there.

szarnekow (commented):
LGTM

Issue #1710, opened 2021-08-03
Building twice because of new "touch on recovery"


Bananeweizen:
#1671 has led to the situation that projects are built twice in certain situations:
* run eclipse with an empty workspace
* import an Xtext project
* watch the Xtext console, it is reported as being built twice
(This is on Xtext 2.25 with the BuildStateDiscarder changes being ported back. But I'm quite sure the behaviour is identical on current Xtext).

The technical background is as follows:
* Multiple operations (create project, open project, refresh workspace) of the SmartImportWizard trigger the needsBuild flag of the AutoBuild job, therefore a (full) build of the newly imported project is started.
* The opening of the project schedules the Xtext recovery build, since this is the first Xtext project in the workspace.
* The BuildStateDiscarder decides to discard the build state (because of the recovery build flag) and to schedule a job for touching the project.
* The full build from the AutoBuild job finishes. Only now the scheduled job for touching the project is really started, because it uses the project as lock, and therefore cannot run concurrently to the AutoBuild triggered build. At this point of time the touch is superfluous, but it has been scheduled long ago.

I believe the exceptions and deadlocks reported in the early versions of #1671 (when the touch was executed immediately instead of via job) were also caused by the touch happening when the auto build already built the same project.

The most simple fix would be to check again _inside_ the touch job run() method whether touching is actually still needed. However, I did not find any good method to check whether a project has been built (via Xtext BuildManagerAccess). Any advice how that could be checked, or do you imagine another fix eventually?

cdietrich:
cc @szarnekow @tivervac

cdietrich:
maybe also andreys changes mentioned in https://bugs.eclipse.org/bugs/show_bug.cgi?id=549704 are related

cdietrich:
with Xtext 2.26 snapshot and eclipse 2021-09 i see

```
[Worker-6: Building] Building demo
[Worker-6: Building] Built demo in 2 ms
[Worker-2: Building] Building demo
[Worker-2: Building] indexing platform:/resource/demo/src/mydsl/MrsGrantsSecretCompartments.statemachine
[Worker-2: Building] Built demo in 152 ms
[Worker-2: Building] Building demo
[Worker-2: Building] Built demo in 0 ms
```

@Bananeweizen @rytina 
i am not sure if the reproduction steps are sufficient or if andreys fix actually fixed the problem

rytina:
[...]

PR #1727, opened 2021-08-21
[#1710] Don't touch the project again if it was already built

szarnekow:
The window of opportunity for the autobuild is the time between
forgetting the last built state and the touch of the project in its own
job. Check if there is a known built state before touching the project
again.

Signed-off-by: Sebastian Zarnekow <sebastian.zarnekow@gmail.com>

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuilderStateDiscarder.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuilderStateDiscarder.java
@@ -10,12 +10,8 @@ package org.eclipse.xtext.builder.impl;
 
 import java.util.Map;
 
-import org.apache.log4j.Logger;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
-import org.eclipse.core.runtime.CoreException;
-import org.eclipse.core.runtime.Status;
-import org.eclipse.core.runtime.jobs.Job;
 
 import com.google.common.annotations.Beta;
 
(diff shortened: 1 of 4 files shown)


szarnekow:
cc @Bananeweizen Can you please give this a try?

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuilderStateDiscarder.java
```diff
@@ -34,12 +37,34 @@
 public class BuilderStateDiscarder {
 
 	private static final Logger logger = Logger.getLogger(BuilderStateDiscarder.class);
+	private static final MethodHandle InternalBuilder_getLastBuiltTree;
 
+	static {
+		try {
+			@SuppressWarnings("restriction")
+			MethodHandles.Lookup lookup = MethodHandles.privateLookupIn(
```
cdietrich, line 45: i assume if we keep java 8 we need to do similar old school reflection as in JavaBuilderState / BuildManagerAccess

cdietrich:
mmm on the first try it does not work

szarnekow:
Testing this is indeed challenging. I’ll revisit the fix.

rytina:
>do you have normal project or java projects?
@cdietrich @szarnekow we have normal projects. Our Xtext grammar does not use Xbase.

szarnekow:
> > do you have normal project or java projects?
> 
> @cdietrich @szarnekow we have normal projects. Our Xtext grammar does not use Xbase.

Thank you for the quick response. The approach taken in the current state of the merge request is not sufficient yet for either type of projects. There are more moving parts involved and an unpleasant mix of different locks at play that need to be straighten. I’ll keep you posted.

cdietrich:
Bug486584Test fails even if I replace the jar with a jar containing a buildertestlanguage file `namespace viech { object Viech }`
and the test file to `namespace bar { object B references viech.Viech }`

szarnekow:
Let's see what Jenkins thinks about this approach.

szarnekow:
@Bananeweizen @rytina  @cdietrich I think this is good for a review and some battle testing.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuilderStateDiscarder.java
```diff
@@ -33,50 +29,42 @@
...
-
 	/**
 	 * Returns true, if the given builderArguments indicate that we should discard the built state
 	 * for the given projects.
 	 */
 	public boolean forgetLastBuildState(Iterable<IProject> toUpdate, Map<String, String> builderArguments) {
+		/*
+		 * Implementation note:
+		 * This is an unsafe operation. There no guarantee that any lock is being acquired by the current thread
+		 * while performing this operation. It is  very well possible that this is running concurrently with a build.
```
tivervac, line 40: There's a double space after `It is`

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuilderStateDiscarder.java
```diff
@@ -33,50 +29,42 @@
...
+		 * 
+		 * When we come here, we might, or might not
+		 * - hold the AbstractBuilderState.loadLock
+		 * - hold the org.eclipse.core.internal.resources.WorkManager.lock
+		 * - come from a build and see a situation with org.eclipse.core.internal.resources.Workspace.treeLocked being the case
```
tivervac, line 45: There's two spaces at the end here

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -158,20 +159,141 @@ public BuilderPreferences () {
...
+		}
+	}
+	
+	/**
+	 * True if the next regular build (not a clean build) that will be executed, will be treated as a full build request.
+	 * 
+	 * @since 2.26
+	 */
+	private boolean nextBuildIsFullBuild = false;
+	/**
```
tivervac, line 196: Should be a empty newline before this line

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -198,16 +320,18 @@ public boolean isCanceled() {
...
 			} else {
 				IResourceDelta delta = getDelta(getProject());
 				if (delta == null || isOpened(delta)) {
-					fullBuild(progress.split(1), IBuildFlag.RECOVERY_BUILD.isSet(args));
+					fullBuild(progress.split(1), !wasFullBuildRequested && IBuildFlag.RECOVERY_BUILD.isSet(args));
 				} else {
 					incrementalBuild(delta, progress.split(1));
 				}
 			}
+			// If no exceptions have been thrown, fix the information about the requested full builds
```
tivervac, line 333: Just trying to clarify the scenario.

IIUC, we would call `requestFullBuild`. This will trigger an autobuild. The autobuild will then somewhere along the line call this method.

In this case we have `nextBuildIsFullBuild` be true here and `allowRequestFullBuild` false. If an exception did now occur, we couldn't request a new full build, the next build will be a full build, but there currently (possibly) is no more build.

In short, if an exception occurred, it's possible that we have an unbuilt project that we do not retry to build.

cdietrich:
does not work for mydsl general projects for me

```
[Worker-0: Building] Building demo2
[Worker-0: Building] Built demo2 in 10 ms
[Worker-7: Building] interrupted
[Worker-7: Building] Build interrupted.
[Worker-7: Building] Built demo2 in 1045 ms

```

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
szarnekow: The AutoBuildJob will be rescheduled in case of cancellation.

cdietrich:
same on open a closed mydsl general project after restart
see a interruped build only and not a follow up build

cdietrich:
looks like the problem i saw happens on all opens.
TODO: check if works in real project
TODO: check if old behaviour can easily be restored by projects when needed (subclassing)

rytina:
@szarnekow I copied the changed files over to our target to test the bug locally and I can confirm it's fixed. The builds only happens once and the recovery build is also working perfectly. Thank you! :-)

rytina (approved):

rytina:
@szarnekow could you please merge your PR and provide us an update site, so that I can test your change with all other changes that will come with Xtext 2.26?

PR #1759, opened 2021-11-05
Update the "connected" property on disconnection

rubenporras:
Update the "connected" property on disconnection

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/XtextResourceMarkerAnnotationModel.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/XtextResourceMarkerAnnotationModel.java
@@ -57,6 +57,12 @@ public class XtextResourceMarkerAnnotationModel extends ResourceMarkerAnnotation
 	}
 
 	@Override
+	protected void disconnected() {
+		super.disconnected();
+		this.connected = false;
+	}
+
+	@Override
 	protected void connected() {
 		super.connected();
 		this.connected = true;


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/XtextResourceMarkerAnnotationModel.java
```diff
@@ -56,6 +56,12 @@ public void fireQueuedEvents() {
 		fireModelChanged();
 	}
 
+	@Override
+	protected void disconnected() {
```
cdietrich, line 60: @szarnekow  do you remember why we dont do this in the first place?

szarnekow: No, I don't remember this.
The change appears to be reasonable.
@rubenporras Can you please explain how you found this? Did you observe a misbehavior after a disconnected?

rubenporras: @szarnekow , this is part of our codebase extending/fixing Xtext: https://github.com/dsldevkit/dsl-devkit/blob/master/com.avaloq.tools.ddk.xtext.ui/src/com/avaloq/tools/ddk/xtext/ui/quickfix/FixedXtextResourceMarkerAnnotationModel.java

The code is quite old, @rmitin,  who contributed it, is no longer working on the DDK, so I do not know if he still knows what was the misbehavior or if he has time to look at it.  I expect that there was for sure a misbehavior observed, otherwise, the Xtext UI class would not have been "patched".

The fix seems logical and it is what I would expect, so I decided to submit it upstream. Sadly I cannot give more information.

miklossy (approved):

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/XtextResourceMarkerAnnotationModel.java
rubenporras: Just in case you did not notice, I do not have right to merge the change, so one with the rights should do merge the change. Thanks

miklossy: Thanks for the PR @rubenporras ! @cdietrich will take care of the merge.

szarnekow (approved):
Thank you for the fix and the background info.

szarnekow:
@rubenporras Can you please sign-off the commit?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/XtextResourceMarkerAnnotationModel.java
rubenporras: done

rmitin: Nice to see the fixes being moved up to Xtext! Thanks, Ruben!

I don't have the access to DDK's git history anymore, but from what I remember, the job that computes markers was occasionally failing with exceptions.

The reason why we noticed it and not other users of Xtext could probably be that our editors support lazy load of documents, which is not a standard feature of an Xtext editor. In order to avoid the UI freezing, if we detect that we need to fetch the document content over the network, the editor first gets a fake readonly document until the real one is fully loaded. This means that we often actually disconnect and re-connect the document right after the editor was opened. So errors related to this become more visible.

PR #1775, opened 2022-01-19
[Bug 574908] Improve performance of Storage2UriMapperJavaImpl in modular project

szarnekow:
Since it is extremely unlikely that the JRE contains sources for which
an Xtext builder is supposed to be executed, the JdtTobeBuiltComputer
was always filtering the JRE from the build processing. The
Storage2UriMapperJavaImpl can do the same. Both share the same
implementation to make that decision.

Also replace pessimistic copies of maps by a synchronized collection in
the PackageFragmentRootData to avoid unnecessary copies of data.

In an unscientific coarse measurement, the time to init the cache for
500 projects was reduced from 15seconds to 40ms.

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/IStorage2UriMapperJdtExtensions.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/IStorage2UriMapperJdtExtensions.java
@@ -10,10 +10,12 @@ package org.eclipse.xtext.ui.resource;
 
 import java.util.Map;
 
+import org.apache.log4j.Logger;
 import org.eclipse.core.resources.IStorage;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.jdt.core.IPackageFragmentRoot;
 import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.launching.JavaRuntime;
 import org.eclipse.xtext.util.Pair;
 
 /**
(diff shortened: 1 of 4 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/IStorage2UriMapperJdtExtensions.java
```diff
@@ -26,10 +28,28 @@
 	/**
 	 * @return all managed uris and their corresponding storages contained in the given {@link IPackageFragmentRoot}
 	 */
-	public Map<URI, IStorage> getAllEntries(IPackageFragmentRoot root);
+	Map<URI, IStorage> getAllEntries(IPackageFragmentRoot root);
```
cdietrich, line 31: change intentional?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
```diff
@@ -314,43 +308,43 @@ protected Object getLastModified(IResource resource) throws CoreException {
...
+		if (shouldHandle(root)) {
+			try {
+				final SourceAttachmentPackageFragmentRootWalker<Void> walker = new SourceAttachmentPackageFragmentRootWalker<Void>() {
 
-				@Override
-				protected URI getURI(IJarEntryResource jarEntry,
-						org.eclipse.xtext.ui.resource.PackageFragmentRootWalker.TraversalState state) {
-					if (!uriValidator.isPossiblyManaged(jarEntry))
-						return null;
-					final URI uri = locator.getURI(root, jarEntry, state);
```
cdietrich, line 333: why is this no longer needed?

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/IStorage2UriMapperJdtExtensions.java
szarnekow: public is the default in interfaces. Yes, intentational.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
szarnekow: I think the diff tricked you. See https://github.com/eclipse/xtext-eclipse/pull/1775/files?diff=split&w=1 with ignored whitespace changes.

szarnekow: Or I misunderstood the question.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
```diff
@@ -501,37 +496,37 @@ protected void clearCache(IJavaProject project, Set<PackageFragmentRootData> toB
 			}
 			// create a copy of the known associated roots to avoid concurrent modification
```
szarnekow, line 497: Comment is outdated

cdietrich (approved):

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
szarnekow: Removed

PR #1778, opened 2022-01-24
[eclipse/xtext#2028] replace log4j Require-Bundle with Import-Package

cdietrich:
[eclipse/xtext#2028] replace log4j Require-Bundle with Import-Package


org.eclipse.xtext.junit4/META-INF/MANIFEST.MF
```diff
@@ -23,7 +23,8 @@ Export-Package: org.eclipse.xtext.junit4,
  org.eclipse.xtext.junit4.ui.util;x-friends:="org.eclipse.xtext.common.types.eclipse.tests",
  org.eclipse.xtext.junit4.util,
  org.eclipse.xtext.junit4.validation
-Import-Package: org.junit;version="4.12.0",
+Import-Package: org.apache.log4j;version="1.2.15";resolution:=optional,
```
szarnekow, line 26: I don't really think the package import is optional, even though the bundle req was optional.

szarnekow: I'm wrong. If the LoggingTester isn't used, it's fine. optional is good.

cdietrich: ok will reintro the optional

PR #1780, opened 2022-01-26
[eclipse/xtext#2028] widen version range for log4j fragment

cdietrich:

### DIFF
--- a/org.eclipse.xtext.logging/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.logging/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Bundle-Name: %pluginName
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.logging/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Localization: fragment
 Bundle-Name: %pluginName
-Fragment-Host: org.apache.log4j;bundle-version="[1.2.15,1.2.16)"
```
cdietrich, line 5: why was this so restrictive in the first place?

szarnekow: IIRC log4j didn't really adhere to semantic versioning - new classes were added even in service releases. It should have been widened to 1.2.17 though in the meantime. I think the proposed change is good.

szarnekow (approved):

PR #1808, opened 2022-03-10
waitForBuild throws with reference to original exception

LorenzoBettini:
Currently, waitForBuild catches a CoreException and throws an
OperationCanceledException with just the message of the original
CoreException. Indeed, OperationCanceledException does not have a
constructor for passing also an exception. Without the original
exception, though, it is hard to understand what went wrong during the
test, since we lost the original exception's stacktracke

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/IResourcesSetupUtil.java
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/IResourcesSetupUtil.java
@@ -370,7 +370,9 @@ public class IResourcesSetupUtil {
 		try {
 			ResourcesPlugin.getWorkspace().build(IncrementalProjectBuilder.INCREMENTAL_BUILD, monitor);
 		} catch (CoreException e) {
-			throw new OperationCanceledException(e.getMessage());
+			OperationCanceledException operationCanceledException = new OperationCanceledException(e.getMessage());
+			operationCanceledException.addSuppressed(e);
+			throw operationCanceledException;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/IResourcesSetupUtil.java
```diff
@@ -370,7 +370,7 @@ public static void waitForBuild(IProgressMonitor monitor) {
 		try {
 			ResourcesPlugin.getWorkspace().build(IncrementalProjectBuilder.INCREMENTAL_BUILD, monitor);
 		} catch (CoreException e) {
-			throw new OperationCanceledException(e.getMessage());
+			throw new RuntimeException(e.getMessage(), e);
```
cdietrich, line 373: can you also fix this in 
org.eclipse.xtext.junit4.ui.util.IResourcesSetupUtil.waitForBuild(IProgressMonitor)

i have no idea why there is OperationCanceledException in the first place. maybe @szarnekow remembers

szarnekow: Sadness: in `#setAutobuild` above we use sneakThrow. 
Though I prefer some consistency, I don't see a good reason why this should be OperationCanceledException except for the fact that some of the APIs in Xtext will gracefully rethrow OCE but might catch other plain RuntimeException.

OperationCanceledException#addSuppressed(e) is probably not much different from the new runtime exception. What do you think?

LorenzoBettini: @szarnekow do you mean something like this?

```java
OperationCanceledException operationCanceledException = new OperationCanceledException(e.getMessage());
operationCanceledException.addSuppressed(e);
throw operationCanceledException;
```

if the result is the same (the stacktrace of the original exception is shown) I'd be fine

szarnekow: Yes, it'd be printed with the stacktrace.

LorenzoBettini: @szarnekow OK, I've pushed-forced with your suggestion
@cdietrich I've also updated the same method in org.eclipse.xtext.junit4.ui.util.IResourcesSetupUtil

szarnekow (approved):

cdietrich (approved):

PR #1809, opened 2022-03-11
better version of needsRebuild feature

cdietrich:
better version of needsRebuild feature

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuildContext.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuildContext.java
@@ -57,7 +57,7 @@ public class BuildContext implements IBuildContext {
 	
 	@Override
 	public void needRebuild() {
-		builder.needRebuild();
+		builder.triggerRequestProjectRebuild();
 	}
 	
 	@Override
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2009, 2022 itemis AG (http://www.itemis.eu) and others.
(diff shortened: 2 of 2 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -755,5 +761,38 @@ public ISchedulingRule getRule(int kind, Map<String, String> args) {
...
+				needRebuild();
+			}
+		} else {
+			needRebuild();
+		}
+		
+	}
+	
+	private static boolean isCoreResourceGreaterOrEqual(Version version) {
+		Version installed = ResourcesPlugin.getPlugin().getBundle().getVersion();
```
szarnekow, line 786: I think Version implements Comparable

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
```diff
@@ -755,5 +761,38 @@ public ISchedulingRule getRule(int kind, Map<String, String> args) {
 			default: throw new IllegalArgumentException();
 		}
 	}
+	
+	/**
+	 * @since 2.27
+	 */
+	public void triggerRequestProjectRebuild() {
```
szarnekow, line 768: It might make sense to accept an IResource or a URI as an argument to be able to request rebuilt for the project into which the file was written / from which the file was removed.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/BuildContext.java
```diff
@@ -57,7 +57,7 @@ public boolean isSourceLevelURI(URI uri) {
 	
 	@Override
 	public void needRebuild() {
```
szarnekow, line 59: Maybe overload with an IProject... or a List<IProject> as an argument.

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/XtextBuilder.java
cdietrich: the problem is that the postprocessor api is really bad for it. 
also generating to different projects does not work ootb. i would skip this for now

Issue #1634, opened 2021-02-12
Investigate atomic marker creation


cdietrich:
Check if we can use and profit from atomic marker creation in validation and generation
e.g.

https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/176167/4/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java

ArneDeutsch:
I've done s syntetic test about how much time we could save. Obviously we can save time only in case we have many markers. If we have only few markers the creation process will be of no importance. Seeing many errors is not so common, but probably there are projects with many warnings and there it may be of importance indeed.

For my measurements I've created an Xtend file with 10.000 Errors (simple method calls to not existent method, 10.000 times). I then modified the code and saved. I instrumented the code to meaure the time the complete `ClusteringBuilderState.doUpdate` method has run and the time the `DefaultResourceUIValidatorExtension.createMarkers` method has run. I tried 5 times. Then I modified `MarkerCreator.createMarker` to use the new API and recorded the times again. Here is the raw data:

```
10000: 372ms
Complete: 2221ms
10000: 370ms
Complete: 1904ms
10000: 348ms
Complete: 2092ms
10000: 333ms
Complete: 1785ms
10000: 374ms
Complete: 2162ms

~360ms per run
~2030ms per run


10000: 214ms
Complete: 1643ms
10000: 195ms
Complete: 1743ms
10000: 188ms
Complete: 1523ms
10000: 190ms
Complete: 1856ms
10000: 184ms
Complete: 1504ms

~195ms per run
~1655ms per run
```

We can see, that we can expect to save around 30% - 50% of the time needed to create the markers resulting in a total save of 5%-10% (at least in this scenario). Of course this saving is only possible in case there are really that many markers.

ArneDeutsch:
In case someone likes to run his own experiments ... here is a quick and dirty modification of MarkerCreator.createMarker that is using the new API:

```
	public void createMarker(Issue issue, IResource resource, String markerType) throws CoreException {
		Map<String, Object> attributes = new HashMap<>();
		
		String lineNR = "";
		if (issue.getLineNumber() != null) {
			lineNR = "line: " + issue.getLineNumber() + " ";
		}
		attributes.put(IMarker.LOCATION, lineNR + resource.getFullPath().toString());
		attributes.put(Issue.CODE_KEY, issue.getCode());		
[...]

PR #1810, opened 2022-03-14
Set marker attributes in bulk

rubenporras:


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/validation/MarkerCreator.java
```diff
@@ -36,28 +39,32 @@ public void createMarker(Issue issue, IResource resource, String markerType) thr
...
+			attributes.put(Issue.URI_KEY, issue.getUriToProblem().toString());
 		if(issue.getData() != null && issue.getData().length > 0) {
-			marker.setAttribute(Issue.DATA_KEY, Strings.pack(issue.getData()));
+			attributes.put(Issue.DATA_KEY, Strings.pack(issue.getData()));
 		}
 		if (resolutionProvider != null && resolutionProvider.hasResolutionFor(issue.getCode())) {
-			marker.setAttribute(FIXABLE_KEY, true);
+			attributes.put(FIXABLE_KEY, true);
 		}
+		marker.setAttributes(attributes);
```
cdietrich, line 67: am not sure if we should postpose this until we have a newer platform.
otherwise we need to use reflection here

rubenporras: to me it does not look like it is worth (at least not from my point of view), but if you like I can spend a little more time and use reflection. For me, I would close the PR and one can retrieve it later if wanted.

As information, on newer platforms, there is, by the way, even a better API that allows to create the marker with its attributes in one workspace operation, which would be even better. The problem is that I think one use it within this classes because they offer a protected method to "set the attributes" and not "to create the maker with its attributes".

szarnekow: I think `#setAttributes(..)` should already be available in the oldest supported platform, isn't it?

rubenporras: I have squashed the commit, just in case, and fixed the commit to pass the eclipse FDN. 
It compiles properly, doesn't that mean that `#setAttributes(..)` is available in the platform you support?

cdietrich: https://github.com/eclipse/xtext-eclipse/issues/1634 says no. do you use a different api?

szarnekow: I think #1634 was about `#createMarker(.., attributesMap)` whereas we are looking at `setAttributes(attributesMap)`

cdietrich: but is there any benefit to change here?

szarnekow (approved):

rubenporras:
Thanks.

As usual you need to merge it, I do not have the rights for that.

cdietrich:
@rubenporras we ususally do rebases on branches, not merges

cdietrich:
@rubenporras we ususally do rebases on branches, not merges

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/tasks/TaskMarkerCreator.java
```diff
@@ -26,13 +29,16 @@ public void createMarker(Task task, IResource resource, String markerType) throw
...
 	protected void setMarkerAttributes(Task task, IResource resource, IMarker marker) throws CoreException {
-		marker.setAttribute(IMarker.LOCATION, "line " + task.getLineNumber());
-		marker.setAttribute(IMarker.PRIORITY, getPriority(task.getTag().getPriority()));
-		marker.setAttribute(IMarker.MESSAGE, task.getFullText());
-		marker.setAttribute(IMarker.LINE_NUMBER, task.getLineNumber());
-		marker.setAttribute(IMarker.CHAR_START, task.getOffset());
-		marker.setAttribute(IMarker.CHAR_END, task.getOffset() + task.getTotalLength());
-		marker.setAttribute(IMarker.USER_EDITABLE, false);
+		// Do this in one single setAttributes() call, as each set of an attribute is a workspace operation
+		Map<String, Object> attributes = new HashMap<>(8);
```
cdietrich, line 33: why 8

cdietrich: next 2^x?

rubenporras: Yes, next  2^x

rubenporras:
@cdietrich, I did rebase my branch. Is it right now?

cdietrich (approved):

cdietrich:
yes, thanks

cdietrich:
thx @rubenporras

Issue #1820, opened 2022-03-21
Support more fine granular "needsRebuild" on generation to different projects


cdietrich:
Support more fine granular "needsRebuild" on generation to different projects
see https://bugs.eclipse.org/bugs/show_bug.cgi?id=579082

cdietrich:
@szarnekow do you see any way to achieve this without breaking api
if we replace / add to 

IBuildContext.needsRebuild a IBuildContext.needsRebuild(IProject)
and people subclass that, it wont be called anymore

szarnekow:
I'd probably go for the invenient but so far reliable approach that checks the implementation class. If the type of the BuildContext is the Xtext default type, we don't call needsRebuild. If it is not and the actual type has an overridden needsRebuild, we call it from needsRebuild(IProject)

cdietrich:
yes but i have no clue how jörg manages to write to other projects without subclassing.

szarnekow:
I think that's supported by if one uses custom generator outlets, e.g. if `org.eclipse.xtext.generator.OutputConfiguration.getOutputDirectory()` returns something `../../projectName/...`

cdietrich:
yes but this will only work if projects are contained in the ws and not imported only

szarnekow:
I don't think so - the path resolution is relative the ws root, isn't it?

cdietrich:
yes. thus i really wonder how jörg does it

cdietrich:
@ Jörg the change that we now do request rebuild on affected projects

iloveeclipse:
@cdietrich : I think Jörg is not subscribed to this ticket :-)

cdietrich:
yes i dont know if he has an github account at all

szarnekow:
Pinging @jukzi: Can you please take a look at this conversation, Jörg?

jukzi:
thanks, i can probably test tomorrow.

jukzi:
We have serious issues with the current xtext builds:

At a first glance after clean build many project with Java Problem:
"The project cannot be built until its prerequisite XXX is built. Cleaning and building all projects is recommended"

I debugged and found why: we have set manual build order in preferences. The build order was wrong. Former Xtext did cure that wrong wrong build order by requesting rebuild.
Well, after fixing the build order now i know there is no cycle in our projects: we would not need any rebuilds of projects that already have been build.
[...]

PR #1830, opened 2022-04-07
[eclipse/xtext-eclipse#1820] jdtqueuedbuilddataexperiments

cdietrich:
[eclipse/xtext-eclipse#1820] jdtqueuedbuilddataexperiments


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/impl/javasupport/JdtQueuedBuildData.java
```diff
@@ -107,9 +108,11 @@ protected boolean doNeedRebuild(JavaBuilderState javaBuilderState, Procedure1<?
 				if (processor != null) {
 					processor.apply(ucDelta);
 				}
+			} else if (ucDelta.getProject().equals(project)) {
+				result = true;
```
szarnekow, line 112: Should this also do a `ucDeltas.remove()`?

szarnekow: No, it shouldn't.

cdietrich: no, cause it is not confirmed yet here

szarnekow (approved):

PR #1869, opened 2022-06-23
adapted to needRebuild changes in platform

cdietrich:
adapted to needRebuild changes in platform


cdietrich:
am not sure if we want to change it. as the class is deprecated am not sure who is using it.
what do you think @szarnekow

szarnekow:
> am not sure if we want to change it. as the class is deprecated am not sure who is using it. what do you think @szarnekow

I think the change here is reasonable and good, even though it's a deprecated class.
If extra care is supposed to be taken, extract a protected method so clients can override it according to their needs, but I doubt this is warranted.

Issue #1853, opened 2022-05-18
Races in Storage2UriMapperJavaImpl


cdietrich:
we observe the following races in `Storage2UriMapperJavaImpl`

initialize race

```
Storage2UriMapperJavaImpl.clearCache(IJavaProject, Set<String>)
Storage2UriMapperJavaImpl.updateCache(IJavaProject)
Storage2UriMapperJavaImpl$3.run(IProgressMonitor)
Storage2UriMapperJavaImpl.doInitializeCache()	
Storage2UriMapperJavaImpl.initializeCache(boolean)
Storage2UriMapperJavaImplelementChanged(ElementChangedEvent)	
```

```
Storage2UriMapperJavaImpl.clearCache(IJavaProject, Set<String>)
Storage2UriMapperJavaImpl.updateCache(IJavaProject)
Storage2UriMapperJavaImpl$3.run(IProgressMonitor)
Workspace.run(ICoreRunnable, ISchedulingRule, int, IProgressMonitor)
Workspace.run(IWorkspaceRunnable, ISchedulingRule, int, IProgressMonitor)
Storage2UriMapperJavaImpl.doInitializeCache()	
Storage2UriMapperJavaImpl$2.run()	
Thread.run()	
```

read on non initialized

```
Storage2UriMapperJavaImpl.getCachedData(Storage2UriMapperJavaImplCopy.java)
	at .Storage2UriMapperJavaImp.getData(Storage2UriMapperJavaImplCopy.java)
	at Storage2UriMapperJavaImpl.getAllEntries(Storage2UriMapperJavaImplCopy.java)
	at org.eclipse.xtext.builder.impl.javasupport.JdtToBeBuiltComputer.updateProject(JdtToBeBuiltComputer.java:123)
```
might be called before
```
Storage2UriMapperJavaImpl.getCachedData(Storage2UriMapperJavaImplCopy.java)
	at Storage2UriMapperJavaImpl.updateCache(Storage2UriMapperJavaImplCopy.java)
	at Storage2UriMapperJavaImpl.elementChanged(Storage2UriMapperJavaImplCopy.java)
	at org.eclipse.jdt.internal.core.DeltaProcessor$3.run(DeltaProcessor.java:1755)
```

in both cases the clearCache will remove stuff that is in toBeKept with the same path/uriPrefix but a different PackageFragmentRootData.

this will lead to follow up errors as there is no IContainer for the Jar, and thus visibleContainer calls etc wont work

[...]

PR #1873, opened 2022-07-05
[#1853] another attempt to fix Storage2UriMapperJavaImpl race

cdietrich:
[#1853] another attempt to fix Storage2UriMapperJavaImpl race

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
@@ -368,7 +368,7 @@ public class Storage2UriMapperJavaImpl implements IStorage2UriMapperJdtExtension
 		// Pessimistic copy of the cachedPackageFragmentRootData
 		Collection<PackageFragmentRootData> packageFragmentRootDatas;
 		synchronized(cachedPackageFragmentRootData) {
-			packageFragmentRootDatas = cachedPackageFragmentRootData.values();
+			packageFragmentRootDatas = newArrayList(cachedPackageFragmentRootData.values());
 		}
 		
 		/* 


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/resource/Storage2UriMapperJavaImpl.java
```diff
@@ -368,7 +368,7 @@ public Iterable<Pair<IStorage, IProject>> getStorages(/* @NonNull */ URI uri) {
 		// Pessimistic copy of the cachedPackageFragmentRootData
 		Collection<PackageFragmentRootData> packageFragmentRootDatas;
 		synchronized(cachedPackageFragmentRootData) {
-			packageFragmentRootDatas = cachedPackageFragmentRootData.values();
+			packageFragmentRootDatas = newArrayList(cachedPackageFragmentRootData.values());
```
cdietrich, line 371: @szarnekow do you know why this was not there in the first place

cdietrich: https://github.com/eclipse/xtext-eclipse/issues/1853#issuecomment-1174731216

szarnekow: A copy of the list on every read is probably not what we want to do.
Maybe use a synchronized block on the list itself after it was read from the map?

cdietrich: hmmmm ok. will try it out as workaround in our codebase

cdietrich: we still need to check if the problem happened again

cdietrich:
as i wont find time to further work on this => close

PR #1878, opened 2022-07-12
Make loading of Xtext index better hookable

cdietrich:
Make loading of Xtext index better hookable


cdietrich:
@szarnekow why do you think? bad idea?

szarnekow:
> @szarnekow why do you think? bad idea?

Hmm I understand that the persister itself is not easily customizable?

cdietrich:
we want to basically hook into the loading so that we can transform the index content into a map in the builderstatate subclass
with the load itself we have to basically copy&paste all the isloaded stuff etc

szarnekow:
Makes sense.

szarnekow (approved):

PR #1888, opened 2022-08-09
[eclipse/xtext#2087] dont add junit container anymore

cdietrich:
[eclipse/xtext#2087] dont add junit container anymore
[This basically undoes](https://github.com/eclipse/xtext-core/issues/666#issuecomment-389567981)
https://github.com/eclipse/xtext-core/issues/666#issuecomment-389567981

am not sure if we want to go that route or live with the container and the other deps for now
@szarnekow @kthoms what do you think

### DIFF
--- a/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
+++ b/org.eclipse.xtext.xtext.ui/src/org/eclipse/xtext/xtext/ui/wizard/project/XtextProjectCreator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2009 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2009, 2022 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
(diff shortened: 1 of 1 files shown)


cdietrich:
at least non of our integ tests seem to fail

szarnekow:
I'm leaning towards Junit5 works, if Maven or PDE is chosen. The Junit5 container stuff on the CP don't really make much sense.

LorenzoBettini:
Sorry if I'm late to this, but IIRC the JUnit container is needed to successfully run tests from Eclipse (especially when there are JUnit 5 and JUnit 4 tests)

cdietrich:
i tested it ond had no problems. its too late anyway :(

LorenzoBettini:
From Eclipse? Do you have both JUnit 4 and JUnit 5? I'm not saying it's not working, I'm just saying that a few months ago it wasn't... I'll have to check again.

cdietrich:
yes. i had no issue. can someone else please check

LorenzoBettini:
In a project with both JUnit 4 and JUnit 5 tests, without the JUnit 5 container in the classpath, only the JUnit 5 tests are executed.

cdietrich:
I don’t think this is a usecase for the wizard
And I still would not have a solution for the tycho problem

LorenzoBettini:
I just wanted to report the use-case when that's required. By the way, updating the Maven project effectively removes the JUnit 5 container, so that's probably a problem in JDT itself...

Issue #1890, opened 2022-08-10
Have better explanation in logging


frankbenoit:
https://github.com/eclipse/xtext-eclipse/blob/92c5dc5cbace37540d513e33dd3efccf946716c6/org.eclipse.xtext.ui.shared/src/org/eclipse/xtext/ui/shared/internal/Activator.java#L87

Seeing this message let me search for a Guice problem. Changing the log level of "com.google.inject" without results.
Later I noticed the logger name and found the source and then understood the reason.

Would be nice if the logging message explains it directly.
Proposal:

    xtext.ui.shared: Multiple contributions to extension point 'org.eclipse.xtext.ui.shared.overridingGuiceModule'. Will use them in unspecified order.

Or even better if it would list the plugins from the contributions in the message. That would even more simplify the analysis.

cdietrich:
Can you please open pr

PR #1891, opened 2022-08-11
[#1890] enhance warning log for multiple overrides in ui.shared

frankbenoit:


org.eclipse.xtext.ui.shared/src/org/eclipse/xtext/ui/shared/internal/Activator.java
```diff
@@ -76,21 +76,25 @@ protected void initializeInjector(BundleContext context) throws CoreException {
 		Module module = new SharedModule(context);
 		if (extensions.length != 0) {
 			int numberOfMixedInModules = 0;
+			StringBuilder sb = new StringBuilder();
+			sb.append("xtext.ui.shared: Multiple contributions to extension point 'org.eclipse.xtext.ui.shared.overridingGuiceModule'."
```
szarnekow, line 80: Fine by me. 
A few nits:
- The sb can be initialized with that string literal directly.
- Having a prefix "xtext.ui.shared" is usually not necessary. This will be implicit from the loggers name.
- Please don't hardcode the extension point name but use the same logic that does the lookup or even use `point.getUniqueIdentifier()`

org.eclipse.xtext.ui.shared/src/org/eclipse/xtext/ui/shared/internal/Activator.java
```diff
@@ -76,21 +76,25 @@ protected void initializeInjector(BundleContext context) throws CoreException {
...
-							log.warn("Multiple overriding guice modules. Will use them in unspecified order.");
-						}
+						sb.append("\n\tcontributor: ").append(e.getContributor().getName());
 					} catch (CoreException e1) {
 						log.error(e1.getMessage(), e1);
 					}
 				}
 			}
+			if (numberOfMixedInModules >= 2) {
+				log.warn(sb.toString());
```
szarnekow, line 96: It might be ok to have multiple mixins. Shall we be able to suppress this warning somehow?

cdietrich: but the behaviour will be unspecified if order is random. we dont see which module overrides which

szarnekow: True.
Assuming there are two bindings for the same key in the contributed modules: Likely one of the languages won't work anymore. So a warning is not sufficient in these cases.
If the bindings are not contradicting each other, the warning won't help either.

What if:
Obtain all modules at once and call `Modules.override(module).with(collectionOfAllModules);`
In case of contradictory bindings, this will blow up in the users face rather than silently with a warning.

Yes, this *might* break existing working setups but it would be a saner default, won't it?

cdietrich: indeed

szarnekow:
@frankbenoit The updated warnings message looks good. Would you be willing to work on the discussed changes to the way override modules are consumed?

szarnekow (approved):

cdietrich:
thx @frankbenoit

org.eclipse.xtext.ui.shared/src/org/eclipse/xtext/ui/shared/internal/Activator.java
frankbenoit: In development I would prefer the fail-fast.
On the other hand, this situation might be relevant at the end user installing a bundle combination not expected.
Having it breaking is perhaps worse than hope for the best? E.g. conflict is in a minor important tweeking. But the remaining essential stuff is OK.
Then the user would be better with the current behavior.

PR #1893, opened 2022-08-14
Update to M2E 2.0

HannesWell:
This PR aims to update `org.eclipse.xtext.m2e` Plugin to the M2E 2.0.1 release that is contributed to Eclipse 2022-09 M3 SimRel.
M2E 2.0 is a major release that has breaking changes in its API, that are incompatible with previous 1.x releases.

I updated the code and adjusted required bundle version ranges accordingly. Furthermore I update the target file to include the latest m2e release 2.0.1. That release is not yet in the 2022-09 release repo but will be with M3, so this change in advance should prevent trouble with the xText contribution.
I hope that is all correct, but please let me know if anything is wrong or missing.

Besides that I updated the setup and removed M2E-connectors incompatible with M2E 2.0.
The connectors for tycho-plugins are not required because M2E now has the interesting parts of the Tycho-M2E connector build in and for all other parts PDE does in the IDE what Tycho does during build. Since Tycho 2.7.4 this is reflected by M2E-lifecycle-mapping-metadata embedded into Tycho's plugins that say that their execution can be ignored. For the parts of xText that use older Tycho warnings, those missing lifecyclemappings are now only just warnings so you can ignore them. If you want you can change the severity to `INFO`.

### DIFF
--- a/org.eclipse.xtext.m2e/META-INF/MANIFEST.MF
+++ b/org.eclipse.xtext.m2e/META-INF/MANIFEST.MF
@@ -5,17 +5,13 @@ Bundle-SymbolicName: org.eclipse.xtext.m2e;singleton:=true
 Bundle-Version: 2.28.0.qualifier
 Bundle-Localization: plugin
 Bundle-Vendor: %providerName
-Require-Bundle: org.eclipse.m2e.core;bundle-version="1.8.0";resolution:=optional,
- org.eclipse.core.resources;bundle-version="3.12.0",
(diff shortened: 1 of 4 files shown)


cdietrich:
manual test looks good. will check if i need todo something about j17 tp once M3 is out.

cdietrich:
thx @HannesWell

HannesWell:
Thank you and you are very welcome.

HannesWell:
I just verified with the latest nightly build that now `o.e.xtext.m2e` works well in conjunction with M2E 2.0.

PR #1894, opened 2022-08-15
m2e: use reflection also for old method

cdietrich:
m2e: use reflection also for old method


HannesWell (approved):
Looks good thanks.
Just the same remark as in https://github.com/eclipse/xtext-xtend/pull/1372#pullrequestreview-1072991849.

PR #1918, opened 2022-10-29
add apache.commons.logging needed by batik needed by platform

cdietrich:
add apache.commons.logging needed by batik needed by platform


LorenzoBettini:
This would not be necessary if we used planner instead of slicer, as I proposed in another PR. @szarnekow

cdietrich:
Yes. But it is still unclear how to detect unwanted 3rd parts deps with that

LorenzoBettini:
@cdietrich @szarnekow it's still unclear to me why do you want to detect unwanted 3rd party deps... in any case, if you depend on something X that depends on something else Y you have no choice but depend on Y as well besides X. Moreover, I still don't understand why you want Domainmodel example only to have such a complicated TP using "slicer" (which, again, was only an unwanted effect of the wrong usage of the TP DSL...)

PR #1923, opened 2022-11-09
[eclipse/xtext#2110] dont run text modifys on bootstrap

cdietrich:
[eclipse/xtext#2110] dont run text modifys on bootstrap


LorenzoBettini (approved):
Now those tasks are executed when projects are imported!

PR #1924, opened 2022-11-10
2110: reorder "Extras" before "Xbase/Xtend Libraries"

LorenzoBettini:
Also put "Maven" as the last one

https://github.com/eclipse/xtext/issues/2110

### DIFF
--- a/releng/org.eclipse.xtext.contributor/Xtext.setup
+++ b/releng/org.eclipse.xtext.contributor/Xtext.setup
@@ -554,6 +554,76 @@
         name="master"/>
     <description>Provide the core grammar infrastructure, core framework and Language Server support</description>
   </project>
+  <project name="extras"
+      label="Extras">
+    <setupTask
+        xsi:type="git:GitCloneTask"
+        id="git.clone.xtext.extras"
+        remoteURI="eclipse/xtext-extras"
+        pushURI=""
+        checkoutBranch="${xtext.git.branch}">
+      <annotation
+          source="http://www.eclipse.org/oomph/setup/InducedChoices">
+        <detail
+            key="inherit">
+          <value>github.remoteURIs</value>
+        </detail>
+        <detail
+            key="label">
+          <value>Github Repository</value>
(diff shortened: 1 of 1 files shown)


cdietrich:
@szarnekow  can you test our this file too?

releng/org.eclipse.xtext.contributor/Xtext.setup
```diff
@@ -554,12 +554,13 @@
         name="master"/>
     <description>Provide the core grammar infrastructure, core framework and Language Server support</description>
   </project>
-  <project name="lib"
-      label="Xbase/Xtend Libraries">
+  <project name="extras"
```
cdietrich, line 557: have zero clue why this should play any role order wise

LorenzoBettini: As I said in https://github.com/eclipse/xtext/issues/2110#issuecomment-1310028045 ;

I guess the Xtend compiler in the installed Eclipse kicks in fast and regenerates the xtend-gen contents of xtext-lib projects with errors (since it cannot find required stuff from Xbase?), making the projects of xtext-lib not compilable, breaking the workflow.

This happens if you start with an empty .gradle directory.

cdietrich: no there is no dependency

cdietrich: i really would like to have a complete log of the failing case

LorenzoBettini: Maybe @szarnekow can confirm that: I guess that when the installed Xtend compiler (in the provisioned Eclipse) automatically compiles the Xtend files in the package `org.eclipse.xtend.lib.annotations` it needs Xbase dependencies in the running environment and these are not yet there if "Extras" projects are not compiled first.
If the .gradle directory already contains Xtext and Xbase bundles (from previous projects builds) it runs fine.

PR #1940, opened 2022-11-19
minimize use of deprecated codetemplates code

cdietrich:
minimize use of deprecated codetemplates code


cdietrich:
@szarnekow what do you think?

szarnekow (approved):
Thank for cleaning up the technical debt

PR #1944, opened 2022-11-23
Enhance Xtext Junit Deprecations

cdietrich:


org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractXtextTests.java
```diff
@@ -0,0 +1,433 @@
...
+import com.google.inject.Provider;
+import com.google.inject.TypeLiteral;
+import com.google.inject.name.Names;
+
+/**
+ * @author Sven Efftinge - Initial contribution and API
+ * @since 2.30
+ */
+@Beta
+public abstract class AbstractXtextTests extends Assert implements ResourceLoadHelper {
```
cdietrich, line 72: am not sure if we want this as api. @szarnekow what do you think? copy to the tests instead?

cdietrich: can you also check if i copy&pasted the correct variant

cdietrich:
see also
https://github.com/eclipse/xtext/issues/2124
https://github.com/eclipse/xtext-core/issues/2029

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractXtextTests.java
szarnekow: I don't think we've used this base class in recent test cases. I think it's convoluted and was superseded by the `@InjectWith` / `@RegisterExtension`

szarnekow: Yes, I think so. C&P& strip down in the test projects.

cdietrich: any hints to strip

cdietrich:
@szarnekow ok to merge?

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/AbstractXtextTests.java
szarnekow: Subsequent work

Issue #1954, opened 2022-12-05
passing --local-repository to 1-maven-build.sh does not seem to work


cdietrich:
```
14:50:51  + ./1-maven-build.sh -s /home/jenkins/.m2/settings.xml --tp=r202203 --local-repository=/home/jenkins/.m2/repository
14:50:51  Calling Maven with ARGS: --batch-mode --update-snapshots --fae -Declipse.p2.mirrors=false -Dtycho.disableP2Mirrors=true -DJENKINS_URL=https://ci.eclipse.org/xtext/ -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=info -s /home/jenkins/.m2/settings.xml -Dmaven.repo.local=$VALUE -Pr202203
```

please also check other scripts

PR #1955, opened 2022-12-05
[#1954] fix local repo passing

cdietrich:
[#1954] fix local repo passing

### DIFF
--- a/1-maven-build.sh
+++ b/1-maven-build.sh
@@ -41,7 +41,7 @@ while [ "$1" != "" ]; do
       MVN_ARGS+=(-DskipTests=true)
       ;;
     --local-repository)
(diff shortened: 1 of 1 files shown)


cdietrich:
am not sure if `-Dmaven.repo.local=/home/jenkins/.m2/repository` is a good idea.
@kthoms do you know if /home/jenkins/.m2/repository will be a fresh dir?

LorenzoBettini:
@cdietrich as far as I know, Jenkins JIRO is completely ephemeral... you might want to ask in the cross dev mailing list, but I'm pretty sure. In general, that setting is useless if I'm right.

cdietrich:
ephemeral Is what i would want in this case.

Issue #1961, opened 2022-12-08
Adapt to JDT changes regarding classpath attribute changes


cdietrich:
Adapt to JDT changes regarding classpath attribute changes

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/492
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/486

cdietrich:
see also https://github.com/eclipse/xtext-eclipse/issues/1916

cdietrich:
Fixed in 2.30

Issue #1916, opened 2022-10-19
Changing Java Doc location for classpath jars leads to full builds


cdietrich:
Changing Java Doc location for classpath jars leads to full builds

the guard in  org.eclipse.xtext.ui.util.JavaProjectClasspathChangeAnalyzer.isAttachmentChangeOnly(IJavaElementDelta)
does not consider it as attachment change only, as there is no child

so we will return false here
https://github.com/eclipse/xtext-eclipse/blob/a8094ef531583336bee69037af8c26e9121675f9/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java#L76

To be investigated: 
* can there be empty children for non javadoc changes
* is there a problem in JDT (maybe there also something got broken somewhen between 2016 and now)

cdietrich:
Fixed in 2.30

PR #1964, opened 2022-12-09
Fix treatment for javadoc change only

cdietrich:
[#1916] added testcase for javadoc change only
[#1961] experimentally fix it

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2017 itemis AG (http://www.itemis.eu) and others.
+ * Copyright (c) 2016, 2022 itemis AG (http://www.itemis.eu) and others.
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License 2.0 which is available at
  * http://www.eclipse.org/legal/epl-2.0.
@@ -12,12 +12,18 @@ import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
+import org.apache.log4j.Logger;
+import org.eclipse.jdt.core.IClasspathAttribute;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaElementDelta;
 import org.eclipse.jdt.core.IJavaModel;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragmentRoot;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.xtext.xbase.lib.util.ReflectExtensions;
+import org.osgi.framework.Version;
 
+import com.google.common.annotations.Beta;
 import com.google.common.collect.Sets;
 
 /**
@@ -28,6 +34,12 @@ import com.google.common.collect.Sets;
  */
 public class JavaProjectClasspathChangeAnalyzer {
 	
+	private static final Logger logger = Logger.getLogger(JavaProjectClasspathChangeAnalyzer.class);
+	
+	private static final Version VERSION_3_33_0 = new Version(3,33,0);
+	
+	private ReflectExtensions reflectExtensions = new ReflectExtensions();
+
 	/**
 	 * retrieves all Java Projects whose classpath was effected by the given delta
 	 * 
@@ -71,9 +83,15 @@ public class JavaProjectClasspathChangeAnalyzer {
 	 * Determines if a given change is a attachment change only
 	 */
 	public boolean isAttachmentChangeOnly(IJavaElementDelta delta) {
-		// there must be at least one child delta
-		if (delta.getAffectedChildren().length == 0) {
-			return false;
+		if (!isJdtCoreGreaterOrEqual(VERSION_3_33_0)) {			
+			if (delta.getAffectedChildren().length == 0) {
+				return true;
+			}
+		} else {
+			// there must be at least one child delta
+			if (delta.getAffectedChildren().length == 0) {
+				return false;
+			}
 		}
 		// all child deltas have to be attachment changes only
 		for (IJavaElementDelta child : delta.getAffectedChildren()) {
@@ -82,6 +100,55 @@ public class JavaProjectClasspathChangeAnalyzer {
 			}
 		}
 		return true;
+	}
+	
+	private static Version installedJdtCoreVersion;
+	
+	private static boolean isJdtCoreGreaterOrEqual(Version version) {
+		if (installedJdtCoreVersion == null) {
+			installedJdtCoreVersion = JavaCore.getPlugin().getBundle().getVersion();
+		}
+		return installedJdtCoreVersion.compareTo(version) >= 0;
+	}
+
+	/**
+	 * this method checks if the change is a source or javadoc (de)attachment change only.
+	 * @since 2.30
+	 */
+	@Beta
+	protected boolean isAttachmentChangeFlagOnly(IJavaElementDelta child) {
+		int flags = child.getFlags();
+		if (flags == 0) {
+			return false;
+		}
+		int F_CLASSPATH_ATTRIBUTES = 0x800000; // TODO use IJavaElementDelta.F_CLASSPATH_ATTRIBUTES
+		int allOtherFlags = flags & ~(IJavaElementDelta.F_SOURCEATTACHED | IJavaElementDelta.F_SOURCEDETACHED | F_CLASSPATH_ATTRIBUTES); 
+		if (allOtherFlags != 0) {
+			// there are other flags present
+			return false;
+		}
+		if ((flags & F_CLASSPATH_ATTRIBUTES) != 0) {
+			// this is also an attribute change
+			// analyze child.getAttributeDeltas()
+			try {
+				// TODO dont use reflection when minimal target is 2023-03
+				Object[] attributeDeltas = (Object[]) reflectExtensions.invoke(child, "getClasspathAttributeDeltas");
+				for (Object attributeDelta : attributeDeltas) {
+					String name = (String) reflectExtensions.invoke(attributeDelta, "getAttributeName");
+					if (!IClasspathAttribute.JAVADOC_LOCATION_ATTRIBUTE_NAME.equals(name)) {
+						return false;
+					}
+				}
+			} catch (ReflectiveOperationException e) {
+				logger.error("Something went wrong with the reflection code to read classpath attribute deltas", e);
+				return false;
+			}
+			// the only attribute change was a javadoc change
+			return true;
+		} else {
+			// source attachment change only
+			return true;
+		}
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


cdietrich:
@iloveeclipse @trancexpress  would be nice if we can test this in our project context too

cdietrich:
@szarnekow would be nice if you could not only review this, but also do some manual tests and try it in your projects' contexts too

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
```diff
@@ -110,13 +173,6 @@ public boolean isRelevantPackageFragmentRootChange(IJavaElementDelta delta) {
 				);
 	}
 
-	private boolean isAttachmentChangeFlagOnly(IJavaElementDelta child) {
```
szarnekow, line 113: I'm aware of client code that uses reflection to access this method. Can we keep the name, add a comment and make it protected instead?

szarnekow: And adjust the implementation accordingly of course

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
```diff
@@ -71,19 +82,71 @@ public boolean isClasspathChangeOnProject(IJavaElementDelta delta) {
...
+		}
+		return installedJdtCoreVersion.compareTo(version) >= 0;
+	}
+
+	protected boolean isSourceAttachOrDetachOrJavadocAttributeChangeOnly(IJavaElementDelta child) {
+		int flags = child.getFlags();
+		if (flags == 0) {
+			return false;
+		}
+		if ((flags & IJavaElementDelta.F_SOURCEATTACHED) != 0) {
```
szarnekow, line 118: I think you can work unconditionally with `int flagsWithoutSourceAttachments = flags & ~(F_SOURCE_ATTCHED | F_SOURCE_DETACHED)`

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
```diff
@@ -71,19 +82,71 @@ public boolean isClasspathChangeOnProject(IJavaElementDelta delta) {
...
+			flags -= IJavaElementDelta.F_SOURCEATTACHED;
+		}
+		if ((flags & IJavaElementDelta.F_SOURCEATTACHED) != 0) {
```
szarnekow, line 121: The if condition has a c&p problem. Should be SOURCEDETACHED

szarnekow (approved):

szarnekow (changes requested):
Accidental approval reverted

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
```diff
@@ -84,6 +102,52 @@ public boolean isAttachmentChangeOnly(IJavaElementDelta delta) {
...
+		int flags = child.getFlags();
+		if (flags == 0) {
+			return false;
+		}
+		int flagsWithoutSourceAttachments = flags & ~(IJavaElementDelta.F_SOURCEATTACHED | IJavaElementDelta.F_SOURCEDETACHED);
+		if (flagsWithoutSourceAttachments == 0) {
+			// flags were attached/detached only
+			return true;
+		}
+		if (flagsWithoutSourceAttachments == 0x800000) { // TODO use IJavaElementDelta.F_CLASSPATH_ATTRIBUTES
```
szarnekow, line 129: This mask should be checked against the original flags:
```suggestion
		if ((flags & 0x800000) != 0) { // TODO use IJavaElementDelta.F_CLASSPATH_ATTRIBUTES
```

cdietrich: but what if the source attachment AND the jar changes.
then i dont want to go into the if and directly report "classpath change"

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
```diff
@@ -110,13 +174,6 @@ public boolean isRelevantPackageFragmentRootChange(IJavaElementDelta delta) {
 				);
 	}
 
-	private boolean isAttachmentChangeFlagOnly(IJavaElementDelta child) {
-		// change is sourceattached or sourcedetached or sourceattached and sourcedetached at the same time
-		return child.getFlags() == IJavaElementDelta.F_SOURCEATTACHED
-				 || child.getFlags() == IJavaElementDelta.F_SOURCEDETACHED
-				|| child.getFlags() == IJavaElementDelta.F_SOURCEDETACHED + IJavaElementDelta.F_SOURCEATTACHED;
```
cdietrich, line 117: @szarnekow here we also do exact check

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
cdietrich: (see also my review comment on the "old" code

cdietrich: and of course my jdt knowledge is too small to create a test for that

szarnekow: Ah, yes. Of course.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
```diff
@@ -84,6 +102,52 @@ public boolean isAttachmentChangeOnly(IJavaElementDelta delta) {
...
+	 * this method checks if the change is a source or javadoc (de)attachment change only.
+	 * @since 2.30
+	 */
+	@Beta
+	protected boolean isAttachmentChangeFlagOnly(IJavaElementDelta child) {
+		int flags = child.getFlags();
+		if (flags == 0) {
+			return false;
+		}
+		int flagsWithoutSourceAttachments = flags & ~(IJavaElementDelta.F_SOURCEATTACHED | IJavaElementDelta.F_SOURCEDETACHED);
```
szarnekow, line 124:
```suggestion
		int F_CLASSPATH_ATTRIBUTES = 0x800000; // TODO use IJavaElementDelta.F_CLASSPATH_ATTRIBUTES
		int allOtherFlags = flags & ~(IJavaElementDelta.F_SOURCEATTACHED | IJavaElementDelta.F_SOURCEDETACHED | F_CLASSPATH_ATTRIBUTES); // TODO use IJavaElementDelta.F_CLASSPATH_ATTRIBUTES
		if (allOtherFlags != 0) {
			return false;
		}
		if (flags == F_CLASSPATH_ATTRIBUTES) {
```

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
szarnekow: Yes, it was an oversight of mine.

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JavaProjectClasspathChangeAnalyzer.java
cdietrich: this would also not work as you can have source and javadoc attachment at the same time

cdietrich:
```
	protected boolean isAttachmentChangeFlagOnly(IJavaElementDelta child) {
		int flags = child.getFlags();
		if (flags == 0) {
			return false;
		}
		int F_CLASSPATH_ATTRIBUTES = 0x800000; // TODO use IJavaElementDelta.F_CLASSPATH_ATTRIBUTES
		int allOtherFlags = flags & ~(IJavaElementDelta.F_SOURCEATTACHED | IJavaElementDelta.F_SOURCEDETACHED | F_CLASSPATH_ATTRIBUTES); 
		if (allOtherFlags != 0) {
			// there are other flags present
			return false;
		}
		if ((flags & F_CLASSPATH_ATTRIBUTES) != 0) {
			// this is an attribute only change
			// TODO analyze child.getAttributeDeltas()
			try {
				// TODO dont use reflection when minimal target is 2023-03
				Object[] attributeDeltas = (Object[]) reflectExtensions.invoke(child, "getClasspathAttributeDeltas");
				for (Object attributeDelta : attributeDeltas) {
					String name = (String) reflectExtensions.invoke(attributeDelta, "getAttributeName");
					if (!IClasspathAttribute.JAVADOC_LOCATION_ATTRIBUTE_NAME.equals(name)) {
						return false;
					}
				}
			} catch (ReflectiveOperationException e) {
				logger.error("Something went wrong with the reflection code to read classpath attribute deltas", e);
				return false;
			}
			return true;
		}
```
???

trancexpress: I don't think you need this here:
```
Object[] attributeDeltas = (Object[]) reflectExtensions.invoke(child, "getClasspathAttributeDeltas");
```
Since this check will not evaluate to `true` unless the API also exists?
```
if ((flags & F_CLASSPATH_ATTRIBUTES) != 0) {
```

cdietrich: @trancexpress  i need to compile with old code, thus i need to read attributes and names with reflection

trancexpress: > @trancexpress i need to compile with old code, thus i need to read attributes and names with reflection
Ah right, never mind then. I thought you were using reflection to check if the method exist.

szarnekow: > ???

Yes, I think so. The last `return true` probably one line below.

cdietrich: i wanted to keep two returns intentionally. should i add an else for the 2nd?

szarnekow (approved):
:+1:

PR #1969, opened 2022-12-12
remove more outdated code switches for old targets

cdietrich:
remove more outdated code switches for old targets


cdietrich:
@szarnekow do you know other ways of feature switches we have besides bundle version check and Class.forName?

szarnekow (approved):

PR #1973, opened 2022-12-14
Fix JREContainerProvider.getJREContainerEntry(IJavaProject) to also work with Java 11 and Java 17

cdietrich:
Fix JREContainerProvider.getJREContainerEntry(IJavaProject) to also work with Java 11 and Java 17

### DIFF
--- a/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JREContainerProvider.java
+++ b/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JREContainerProvider.java
@@ -43,7 +43,7 @@ public class JREContainerProvider {
 	/**
 	 * @since 2.7
 	 */
-	public static final String PREFERRED_BREE = "JavaSE-1.8";
+	public static final String PREFERRED_BREE = "JavaSE-11";
 	private static IVMInstall defaultVMInstall = null;
 	private static boolean defaultVMinitialized = false;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JREContainerProvider.java
```diff
@@ -105,12 +105,11 @@ protected static IPath newPreferredContainerPath() {
 	}
 
 	public static IClasspathEntry getJREContainerEntry(IJavaProject javaProject) throws JavaModelException {
-		IClasspathEntry defaultJREContainerEntry = getDefaultJREContainerEntry();
-
+		IPath defaultJREContainerPathPrefix = JavaRuntime.newDefaultJREContainerPath().append(StandardVMType.ID_STANDARD_VM_TYPE);
```
szarnekow, line 108: It looks as if this should make sense, yes.

szarnekow (approved):
Shall we deprecate / remove the JRE15 (as in Java 1.5) stuff from this class?

cdietrich:
there is a lot more jre 5/6 stuff. also all over xbase i guess

org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/util/JREContainerProvider.java
```diff
@@ -83,6 +83,7 @@ private static IVMInstall defaultVMInstall() {
 	/**
 	 * @return JRE container path {@link IPath} for standard VM "J2SE-1.5"
 	 */
+	@Deprecated(since = "2.30", forRemoval = true)
```
cdietrich, line 86: @szarnekow ok this way?

szarnekow: Absolutely

PR #1980, opened 2022-12-16
Make resource save deterministic by forcing the resolution of cross-refs

rubenporras:
Otherwise it can be that between the builder process the resource and the builder participant is called, the resource set is cleared, and then by loading it on demand we might persist a resource with proxies.

### DIFF
--- a/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java
+++ b/org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java
@@ -34,6 +34,7 @@ import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.ResourceSet;
+import org.eclipse.xtext.EcoreUtil2;
 import org.eclipse.xtext.builder.DerivedResourceMarkers.GeneratorIdProvider;
 import org.eclipse.xtext.builder.EclipseResourceFileSystemAccess2.IFileCallback;
 import org.eclipse.xtext.builder.preferences.BuilderPreferenceAccess;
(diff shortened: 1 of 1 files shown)


org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java
```diff
@@ -679,6 +680,7 @@ protected void saveResourceStorage(Resource resource, IFileSystemAccess access)
 		if (resource instanceof StorageAwareResource && access instanceof IFileSystemAccessExtension3) {
 			IResourceStorageFacade storageFacade = ((StorageAwareResource) resource).getResourceStorageFacade();
 			if (storageFacade != null) {
+				EcoreUtil2.resolveLazyCrossReferences(resource, CancelIndicator.NullImpl);
```
cdietrich, line 683: will this have any consequences for the normal usecase? with the normal disabled case the rs never will be cleared
should we find out which clustering policy is used?

cdietrich:
@rubenporras in general can you create an issue and then a pr linking to the issue?

rubenporras:
for https://github.com/eclipse/xtext/issues/2484

org.eclipse.xtext.builder/src/org/eclipse/xtext/builder/BuilderParticipant.java
rubenporras: Yes, that is true, but if the resource has not been cleared, EcoreUtil2.resolveLazyCrossReferences will not do anything more than what the builder already did. So it should not have consequences.

As a side note, I think the real fix would be to save the resource in the builder, and not in the builder participant, because then the problem does not exists to begin with, but I do not know why it is not like that already, and thus propose the smallest change possible.

cdietrich: was this measured?

cdietrich: maybe we can guard this with an injected boolean.

szarnekow:
A more holistic approach would be to run the generator on the resource that have already been processed before the resource set is cleared. Might that be an option?

rubenporras:
Yes, I think that is the right option, if you do not mind the right refactor. That is what I meant with my post before "the real fix would be to save the resource in the builder, and not in the builder participant, because then the problem does not exists to begin with, but I do not know why it is not like that already, and thus propose the smallest change possible.".

cdietrich:

PR #1988, opened 2023-01-26
Topic/quickfix testing

### LINKED ISSUE #1987: Unstable ordering of markers in AnnotatedTextToString
Markers are currently sorted by region start only. So when multiple markers have the same region start, they are ordered undeterministically among each other. That makes it difficult to use in some test scenarios as we can't compare to a fixed reference.

I'll try to make a submission to fix this.

eliericha:
This fixes #1987 . I was hoping to run tests locally and update test references impacted by this change, but I couldn't get a fully working development setup with Oomph. I'm still working on fixing my setup but if someone else manages to do it in the mean time, that would be helpful. Thanks!

### DIFF
--- a/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/AnnotatedTextToString.java
+++ b/org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/AnnotatedTextToString.java
@@ -128,7 +128,15 @@ public class AnnotatedTextToString {
 	@Override
 	public String toString() {
 		String contents = getContents();
-		regions.sort(Comparator.comparing((region) -> region.start));
+		/*
+		 * This is used in testing to produce a string comparable with a static
+		 * test reference. Therefore it is important to produce a deterministic
+		 * result. First sort by region start, then end (so we get shorter
+		 * regions first), then by message. The latter is important so that
+		 * regions with the same start and end get ordered deterministically.
+		 */
+		regions.sort(Comparator.comparing((CommentedRegion region) -> region.start).thenComparing((region) -> region.end)
+				.thenComparing((region) -> region.text));
 
 		List<SimpleEntry<Integer, CommentedRegion>> sorted = new ArrayList<>();
 		for (int index = 0; index < regions.size(); index++) {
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/editor/quickfix/AbstractDeclarativeQuickfixProvider.java
```diff
@@ -37,7 +37,7 @@ protected Predicate<Method> getFixMethodPredicate(final String issueCode) {
 			@Override
 			public boolean apply(Method input) {
 				for (Fix annotation : input.getAnnotationsByType(Fix.class)) {
-					boolean result = annotation != null && issueCode.equals(annotation.value())
+					boolean result = annotation != null && annotation.value().equals(issueCode)
```
szarnekow, line 40: Maybe use `Objects.equal(annotation.value(), issueCode)` to be really safe against NPEs?

eliericha: Ah good point. Done.

szarnekow (approved):

org.eclipse.xtext.ui.testing/src/org/eclipse/xtext/ui/testing/util/AnnotatedTextToString.java
```diff
@@ -128,7 +128,15 @@ public AnnotatedTextToString withMarkersFromFile() {
...
 		String contents = getContents();
-		regions.sort(Comparator.comparing((region) -> region.start));
+		/*
+		 * This is used in testing to produce a string comparable with a static
+		 * test reference. Therefore it is important to produce a deterministic
+		 * result. First sort by region start, then end (so we get shorter
+		 * regions first), then by message. The latter is important so that
+		 * regions with the same start and end get ordered deterministically.
+		 */
+		regions.sort(Comparator.comparing((CommentedRegion region) -> region.start).thenComparing((region) -> region.end)
```
szarnekow, line 138: Nit: You could use `comparingInt` and `thenComparingInt`

PR #1995, opened 2023-02-08
add args4j to oomph

cdietrich:
add args4j to oomph


LorenzoBettini (approved):
I had also added that in my branch https://github.com/eclipse/xtext-eclipse/commit/9070754f40b733248acd0ac3dffd02784052269e and I will merge with this one later.

PR #1998, opened 2023-02-21
EclipseSourceFolder fixed getPath

### LINKED ISSUE #1997: EclipseSourceFolder.getPath should blindly add a trailing slash
The current implementation is

```java
	@Override
	public URI getPath() {
		return URI.createPlatformResourceURI("/" + project.getName() + "/" + name + "/", true);
	}
```

So, a trailing slash is always added. This usually is harmless, but when `org.eclipse.xtext.workspace.ISourceFolder.contains(URI)` kicks in:

```java
	default boolean contains(URI uri) {
		URI path = getPath();
		return UriUtil.isPrefixOf(path, uri);
	}
```

This is heavily used by trace file generation to create the source file relative path. In particular, from `org.eclipse.xtext.ui.workspace.EclipseProjectConfig.findSourceFolderContaining(URI)`:

```java
	@Override
	public ISourceFolder findSourceFolderContaining(URI member) {
		Optional<? extends ISourceFolder> srcFolder = getSourceFolders().stream().filter(folder -> folder.contains(member)).findFirst();
		return srcFolder.isPresent() ? srcFolder.get() : null;

	}
```

By the way, the source code should be split into several lines to make debugging easier, and the test for isPresent is an anti-pattern (one should do `orElse(null)`).

When the `.classpath` has source folders with a trailing slash, the `contains` method returns false, and the relative path in the trace file is wrong, breaking debugging of an Xbase DSL and the jump to the original DSL member from a Java file. (It was hard to track down the problem).

I seem to understand that trailing slashes in `.classpath` are admissible, and they are alos automatically added when updating a Maven/Tycho project because, typically, in `build.properties`, source folders have trailing slashes.

Maybe, `EclipseSourceFolder.getPath` should be implemented similar to `FileSourceFolder` **concerning** the trailing slash:

```java
	@Override
	public URI getPath() {
		final URI result = URI.createFileURI(name).resolve(parent.getPath());
		if (result.hasTrailingPathSeparator()) {
			return result;
		} else {
			return result.appendSegment("");
		}
	}
```

@szarnekow what do you think?

The problem is that I don't understand where this method is tested.

LorenzoBettini:
closes #1997

since for tests I need a project, I created a single test to speed up things

### DIFF
--- a/org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/workspace/EclipseSourceFolderTest.java
+++ b/org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/workspace/EclipseSourceFolderTest.java
@@ -23,20 +23,20 @@ public class EclipseSourceFolderTest extends AbstractWorkbenchTest {
 	@Test
 	public void testEclipseSourceFolderTrailingPathSeparator() throws CoreException {
 		IProject project = JavaProjectSetupUtil.createJavaProject("aproject").getProject();
-		EclipseSourceFolder sourceFolderWithoutTrailingSpace = new EclipseSourceFolder(
+		EclipseSourceFolder sourceFolderWithoutTrailingSlash = new EclipseSourceFolder(
 			project, "afolder");
-		EclipseSourceFolder sourceFolderWithTrailingSpace = new EclipseSourceFolder(
+		EclipseSourceFolder sourceFolderWithTrailingSlash = new EclipseSourceFolder(
 			project, "afolder/");
 		assertEquals("project P/aproject name afolder (platform:/resource/aproject/afolder/)",
-			sourceFolderWithoutTrailingSpace.toString());
+			sourceFolderWithoutTrailingSlash.toString());
 		// no double trailing path separator
 		// see https://github.com/eclipse/xtext-eclipse/issues/1997
 		assertEquals("project P/aproject name afolder/ (platform:/resource/aproject/afolder/)",
-			sourceFolderWithTrailingSpace.toString());
+			sourceFolderWithTrailingSlash.toString());
 
 		// equals must be based on path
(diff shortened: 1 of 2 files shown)


org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/workspace/EclipseSourceFolderTest.java
```diff
@@ -0,0 +1,42 @@
...
+ * @author Lorenzo Bettini - Initial contribution and API
+ */
+public class EclipseSourceFolderTest extends AbstractWorkbenchTest {
+
+	@Test
+	public void testEclipseSourceFolderTrailingPathSeparator() throws CoreException {
+		IProject project = JavaProjectSetupUtil.createJavaProject("aproject").getProject();
+		EclipseSourceFolder sourceFolderWithoutTrailingSpace = new EclipseSourceFolder(
+			project, "afolder");
+		EclipseSourceFolder sourceFolderWithTrailingSpace = new EclipseSourceFolder(
```
szarnekow, line 28: Nit: TrailingSpace should probably be TrailingSlash

szarnekow (approved):

org.eclipse.xtext.ui.tests/tests/org/eclipse/xtext/ui/tests/workspace/EclipseSourceFolderTest.java
LorenzoBettini: You're right! I fixed that in the next commit

LorenzoBettini:
@szarnekow @cdietrich just to be sure, I'm also building `xtext-xtend` on my branch before merging `lb_2052_maven_tycho`

LorenzoBettini:
`xtext-xtend` reacts well to this change, so I'm merging

PR #2003, opened 2023-03-17
update to 4.28 i-builds in latest target platform

cdietrich:
update to 4.28 i-builds in latest target platform


LorenzoBettini:
@cdietrich if you change things in the split repos we should remember to port them in the monorepo as well.. changes like this one cannot be simply cherry picked

cdietrich:
yes. i will pr this against monorepo as soon sebastian and you are through with the move
(tracked via https://github.com/eclipse/xtext/issues/2141)
i am getting 20 build failed slack messages a day cause of this which makes it hard to spot real problems.
thus i wanted to get this fixed.

